Issue #20, opened 2023-06-19
NSFW Blur thumbnails


jordan1776:
**Describe the solution you'd like**
There should be an option to blur NSFW thumbnails. I l like the way Sync for Reddit handles this with an option to blur all previews and another for just on the main feeds.

guigs4:
Working on it!

hjiangsu:
This should be released in the latest version v0.2.1+5! Closing this issue.

PR #30, opened 2023-06-20
Added the toggle for hiding nsfw post thumbnails

guigs4:
Addresses issue #20 
Currently behaving like Apollo, where it will blur both the post card and post view.

image

image

### DIFF
--- a/lib/community/widgets/post_card.dart
+++ b/lib/community/widgets/post_card.dart
@@ -29,7 +29,7 @@ class PostCard extends StatelessWidget {
     final bool showFullHeightImages = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_full_height_images') ?? false;
     final bool showVoteActions = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_vote_actions') ?? true;
     final bool showSaveAction = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_save_action') ?? true;
-    final bool hideNsfwPreviews = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_save_action') ?? true;
+    final bool hideNsfwPreviews = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_hide_nsfw_previews') ?? true;
 
     final bool isUserLoggedIn = context.read<AuthBloc>().state.isLoggedIn;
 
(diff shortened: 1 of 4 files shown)


lib/community/widgets/post_card.dart
```diff
@@ -29,6 +29,7 @@ class PostCard extends StatelessWidget {
     final bool showFullHeightImages = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_full_height_images') ?? false;
     final bool showVoteActions = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_vote_actions') ?? true;
     final bool showSaveAction = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_save_action') ?? true;
+    final bool hideNsfwPreviews = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_save_action') ?? true;
```
hjiangsu, line 32: noticed this is using a different setting than the one set in the settings - it should be `setting_general_hide_nsfw_previews`

lib/post/widgets/post_view.dart
```diff
@@ -24,6 +24,8 @@ class PostSubview extends StatelessWidget {
     final theme = Theme.of(context);
 
     final bool isUserLoggedIn = context.read<AuthBloc>().state.isLoggedIn;
+    final bool hideNsfwPreviews = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_show_save_action') ?? true;
+
```
hjiangsu, line 28: same here - the name of the setting should be `setting_general_hide_nsfw_previews`

lib/community/widgets/post_card.dart
guigs4: My bad!
Will fix.

guigs4:
Fixed!

lib/shared/media_view.dart
```diff
@@ -37,79 +40,101 @@ class MediaView extends StatelessWidget {
...
-                          ],
+                            onTap: () => _launchURL(post?.url!),
+                          ),
                         ),
                       ),
-                      onTap: () => _launchURL(post?.url!),
                     ),
                   ),
-                ),
+                  Container(alignment: Alignment.center,
```
hjiangsu, line 119: I think for this one (feel free to let me know your own thoughts on it), but maybe it have the conditional first rather than within the `child` parameter

```
if (hideNsfw) Container(alignment: Aligntment.center ...)
```

This way, we don't have to render an empty Container, which might be better performance wise

hjiangsu:
Just left a few comments, otherwise, everything else looks good! Thanks

lib/shared/media_view.dart
guigs4: You're right, definitely an oversight on my part 😅

Fixed!

PR #33, opened 2023-06-21
Used trim function to trim unwanted symbols :)

MrAntonS:
Pretty much 3 lines of code changed. I tried to play around with databases, but if i try to use sqflite_common_ffi, then it won't build for android. Do you know the way to make this app work with DB on linux?


MrAntonS:
Yep, no probs

PR #38, opened 2023-06-22
Linux/Windows sqflite support, Login var trimming

MrAntonS:
Simple change to the database code to support Windows and Linux build. It's much easier to develop without compiling APK. Also added trim function on login to prevent error when pasting thing like instance with new line byte at the front or back.


hjiangsu:
It looks good! Just one question - was this tested on an Android build with the change of  `sqflite` to `sqflite_common_ffi`?

Also, if `sqflite` is no longer used, then I would suggest removing it from `pubspec.yaml`!

MrAntonS:
Yes I did build the android app and it ran fine. You should consider building one yourself to make sure it euns same on different devices

hjiangsu:
Awesome, thanks for double checking - I'll approve this and perform a few tests on my end too for iOS devices

PR #44, opened 2023-06-22
Gif support

MrAntonS:
So apparently, when image full height option was selected, the program didn't have a handle for gif. So I implemented it. Take a look, i might've missed something. Works fine on linux, building android apk now


MrAntonS:
I think i forgot to change headers back, not sure what range header does anyway but feel free to change it back

hjiangsu:
> I think i forgot to change headers back, not sure what range header does anyway but feel free to change it back

The range header limits the amount of bytes to receive from the HTTP request. This is mainly so that I can have some logic to determine the width/height of the image without downloading the whole image first and then getting its dimensions. This _hopefully_ makes it so that its much faster to load up the feed if the "Show full image height" option is selected in the settings. Otherwise, you would see the loading icon for much longer which is something that I'm trying to reduce!

Generally, the range should be the upper limit of bytes required in order to read the image metadata (in this case, the upper limit of png/jpeg). I'm not sure how gifs store that width/height metadata, so the range numbers might have to be adjusted to fit that case

MrAntonS:
For GIF its 10 bytes i think? so we should be good

MrAntonS:
Other than that, now GIF show properly

MrAntonS:
Yeah, works just fine, much faster. I also played around with NSFW blur and i think we should increase the blur. I also would like to stop GIF and MP4(in the future) animation when the content is blurred. What do you think?

hjiangsu:
Just took a look at it, and it seems to be good! Thanks for the PR, I'll merge it in

> I also played around with NSFW blur and i think we should increase the blur.

Personally, I'm okay with keeping it or increasing the value. There's not enough feedback yet from the latest release to know if the blur amount should be adjusted. Maybe the best thing to do for now is to let the community decide if this is needed?

> I also would like to stop GIF and MP4(in the future) animation when the content is blurred. What do you think?

I agree! Currently, GIFs are displayed through `cached_network_image` if I recall correctly. I'm not sure if theres actually a way to stop a GIF so we might have to take a look at that. As for videos, that should be simple enough to do once we add in the ability to view videos within the app

Issue #52, opened 2023-06-24
Logging in with long passwords does not work


ShizuKoto:
**Description**  
When trying to login using a long password >13 chars the app just shows "wrong password".

**How to Reproduce**  
1. Set your password on any instance to a64 or  128 char long password (e.g. g?~zNvX`W-[=T8.4gb+XsCQcIK+jDj}-h4]P?${nYT[1K6X.ANNs9/1{O"W5%2AP) 
2. No don't worry that's not my password
3. try to login thru the app 
4. App shows "password wrong" 
5. Exact same password works on the lemmy instance web ui and jerboa directly
6. Set your password to 13 chars and try to login
7. works

**Expected Behavior**  
The login should be successful

**Device & App Version:**  
 - Device: Samsung Galaxy s21 Ultra
 - OS: Android 13

MrAntonS:
Hmm, i have a 28 symbol long password and don't seem to have an issue, let me try this password to check

MrAntonS:
> Hmm, i have a 28 symbol long password and don't seem to have an issue, let me try this password to check
Had no issue signing in on check account with this exact password

ShizuKoto:
Weird.. might have something to do with my instance? lemmy.world

ShizuKoto:
I just tried deleting the app and reinstalling it. Tried 3 different instances with my 128 char long passwords. Same issue.
Maybe try 128 chars?

MrAntonS:
Do they have spaces in the front by any chance?

ShizuKoto:
nah my password manager does not generate spaces in my passwords.

MrAntonS:
Even weirder

ShizuKoto:
Please don't tell me 128chars work for you? D:

MrAntonS:
Let me check that generated 128 with bitdefender just now

MrAntonS:
Works on lemmy.fmhy.ml. I'll check lemmy.world XD

MrAntonS:
Reinstalled app completely to version 0.2.1+5 just in case I accidentally fixed the issue, but it works fine

ShizuKoto:
So the only common link here is my password manager. I use KeePass and never had such issues.. 
Thank you for checking it out tho!

hjiangsu:
Hmm, interesting - I have not yet tested this but I dont see a reason why 128 char password shouldn't work.

[...]

PR #53, opened 2023-06-24
Fixed password issue with spaces in front/back

MrAntonS:
Might be related #52 to instances accept passwords with leading spaces, so there can be an issue at one point. Other than that no real changes .


hjiangsu:
Hmm, this brings up an interesting question - do signups on instances allow for white space at the beginning/end of usernames/passwords?

Either way, I'll merge this in just in case instances do allow for that edge case. Thanks!

MrAntonS:
> Hmm, this brings up an interesting question - do signups on instances allow for white space at the beginning/end of usernames/passwords?
> 
> Either way, I'll merge this in just in case instances do allow for that edge case. Thanks!

Yes, tested myself :D

PR #65, opened 2023-06-25
Changed default sort of search page

Benjamint22:
Related issue: https://github.com/hjiangsu/thunder/issues/63

The default sort for the communities in the "Search" page looks to be "Hot", which didn't seem to give very relevant issue and often required me to scroll quite a bit to find the community I was looking for.

I chose `SortType.Active` as the new default sort over others like `SortType.TopAll` to avoid suggesting inactive communities, but that's pretty subjective 🤷.

Here's a comparison of the different sorts when searching "technologies":
<table>
<tr>
 <th width="33%">Default (current)
 <th width="33%">`SortType.Active`
 <th width="33%">`SortType.TopAll`
<tr>
 <td>

image
 <td>

image

 <td>

image
</table>

### DIFF
--- a/lib/search/bloc/search_bloc.dart
+++ b/lib/search/bloc/search_bloc.dart
@@ -58,6 +58,7 @@ class SearchBloc extends Bloc<SearchEvent, SearchState> {
           q: event.query,
           page: 1,
           limit: 15,
+          sort: SortType.Active,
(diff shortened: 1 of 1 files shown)


hjiangsu:
Looks good, thanks!

PR #73, opened 2023-06-25
Add adaptive icons for Android

coslu:
Adds support for [adaptive launcher icons](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive) for Android, making it possible for the icon to take different shapes. Also supports API 33 themed-icons. Example:

RSScreenshot_20230625_163142


hjiangsu:
That looks good! Thanks for the contribution

PR #80, opened 2023-06-25
Added default feed type option

### LINKED ISSUE #16: Define Thunder's default feed window
Yes, I simply want to access my subscriptions feed by default because my local feed is not quite... good?

A function to choose the default feed in Thunder.

It may also be that the "default feed" is the last feed that the user left open, although it may not be intuitive at all.

**Additional context**

JulianPaulus:
Hi!

I added the option to select the default feed type for the main screen. This should resolve #16. I added the option to the general settings. I'm happy to take in feedback about positioning and naming!


hjiangsu:
Hey, could you share a screenshot of how it looks like? Thanks so much for helping out!

hjiangsu:
Just a heads up, I also adjusted a bit of the settings page so that theres different categories as well
- I added in a feed section to the general settings and moved theming to a different page

It would be good if the default sort type option is placed under the Feed section!

simulator_screenshot_8C125325-9D54-44C3-8AAB-E2328A43D627

JulianPaulus:
Moved it around, here's the screenshot!
Screenshot_1687712349

hjiangsu:
It looks good! Thanks!

hjiangsu:
Just a quick update - i made very slight changes to the look! Just removed the underline and added some rounding to the options

simulator_screenshot_1E166B66-CBD1-4427-AB76-A315C232296F

JulianPaulus:
Looks nice! I'm currently just also adding the same functionality for the default sorting since it'd basically the same mechanism.

I also added a constants.dart file under utils for the default definitions, since they are needed in various places.

Issue #34, opened 2023-06-21
Refreshing the feed resets sort mode to 'hot'.


jordan1776:
**Describe the bug**
Refreshing the feed resets sort mode to 'hot'.

**To Reproduce**
1. Go feed view
2. Change sort to anything that is not 'hot'
3. Refresh feed

**Expected behavior**
Sort mode should stay the same upon refresh.

**Smartphone (please complete the following information):**
 - Device: Xperia 1 II
 - OS: Android 11

hjiangsu:
Yup, noted - I also saw this when I was doing some testing prior to v0.2.0-alpha, but decided to leave it be for the time being.

In an upcoming release, I'm planning to add the option to choose the default sort/feed types within the settings, so I'll revisit this issue when I work on that!

hjiangsu:
This has been fixed by @JulianPaulus and will be released in the next version

PR #82, opened 2023-06-25
added default sorting option

JulianPaulus:
and here's the default sorting option :)
Screenshot_1687714324


JulianPaulus:
Ah, I just saw that issue #34 exists, I'll also take a look at that.

JulianPaulus:
That should do it!

The dropdown currently looks like this:
image

We could also change it so that it's using the same mechanism as the selector on the communities:
image

What do you think about that?

shortwavesurfer2009:
Doing it with that sort by dialog woild keep consistancy and looks good to me at least

hjiangsu:
Hey! Just took a quick look over it and it seems pretty good, thanks! I think for now, we can leave it as the dropdown so that we can bring out the feature a bit sooner.

However, I do agree that it makes more sense to have it be consistent with the current way we select the sort

Issue #90, opened 2023-06-25
Comments should collapse on tap anywhere


CTalvio:
**Is your feature request related to a problem? Please describe.**
Currently comments can only be collapsed by tapping the tiny sliver of space next to their username.

**Describe the solution you'd like**
Tapping anywhere on a comment, should collapse it.

Benjamint22:
Alternatively, I really like the behavior of Sync for Reddit where you can collapse by holding and you can get the upvote/downvote buttons by tapping.

machinaeZER0:
If it were combined with being able to do so anywhere on the comment, I would LOVE the option to collapse threads with a long press. For some reason long pressing to collapse feels more natural to me, not sure why. I think a single tap to collapse anywhere on a comment feels like it might get accidentally triggered more often than the current behavior, but long press + anywhere on the comment is a great combo (speaking personally)

machinaeZER0:
That makes sense to me!

CTalvio:
ABSOLUTELY DO NOT MAKE COLLAPSING A LONG TAP!

As an option, fine. But losing tap to collapse would be a deal-breaker for me. It would too slow. In Relay, I would often collapse over a dozen comments in quick succession, skimming their contents, browsing for threads that interest me.

A long tap would MASSIVELY slow down browsing for me. We already have gestures for votes, single tap does not need to be made into something that displays redundant UI for this.

machinaeZER0:
Lol. Sounds like you have nothing to worry about!

hjiangsu:
Closing issue - fixed in release v0.2.1+8!

PR #95, opened 2023-06-26
Display parent comment when collapsed and collapse comment on tap of anywhere

### LINKED ISSUE #21: Option to not collapse parent comment in comments
Taping comments currently hides the comment tapped as well as the comments. This isn't wrong, but some other apps only collapse the replies.

There should be an option in setting to keep the parent comment visible.

vbh:
Closes #21 
Also addresses #90 Closes child comments by tapping anywhere in the parent comment

### DIFF
--- a/lib/post/widgets/comment_card.dart
+++ b/lib/post/widgets/comment_card.dart
@@ -210,67 +210,73 @@ class _CommentCardState extends State<CommentCard> {
                   GestureDetector(
                     behavior: HitTestBehavior.opaque,
                     onTap: () => setState(() => isHidden = !isHidden),
-                    child: Padding(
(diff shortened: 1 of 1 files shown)


hjiangsu:
It looks good!

One thing though - how does this work with a single comment without any replies? Do you know what the behaviour is when you tap on a comment without replies?

vbh:
When there are no replies, single comment will still be displayed and there is no difference on tap. Only when there are replies we collapse and open comments.

thomasgirgenti:
Hey can we make a setting to toggle this behavior? I liked being able to fully collapse any comment.

hjiangsu:
Definitely! Feel free to open up a new issue on GitHub for this

Issue #101, opened 2023-06-26
Sort by Top


TheSaneWriter:
**Is your feature request related to a problem? Please describe.**
Right now, the Thunder app has all of the native Lemmy sorts except for the ability to sort by Top. That's my favorite sort method, so I would really like for that to be added.

**Describe the solution you'd like**
When I click the select sort button, one of the options is sort by Top, and then when tap it it has time-frames available (usually day, month, and all time).

**Describe alternatives you've considered**
Same as above, except with 3 separate buttons, one for each timeframe.

JulianPaulus:
I tried designing this:
![Screenshot_1687802522](https://github.com/hjiangsu/thunder/assets/6230679/675ecf51-6983-4fa0-aba5-ab0e14d4299e)

![Screenshot_1687802526](https://github.com/hjiangsu/thunder/assets/6230679/2530f76f-6a64-442d-a06c-dd4237bf0e06)

I'm mostly having problems finding the right Icons :)
@hjiangsu I also think this could be combined with the redesign of the sort picker in the settings.

hjiangsu:
@JulianPaulus oh man, that looks pretty good to me! Same here, I have a hard time trying to find the proper icons 😅

CTalvio:
Maybe a clock with the hour hand in different spots for the hourly top times?

hjiangsu:
This has been added thanks to @JulianPaulus and will be available in the next release!

hjiangsu:
Closing as this should be fixed in v0.2.1+10!

PR #143, opened 2023-06-29
Sort picker redesign

JulianPaulus:
Hey, this is a fix for #101.

This changes the ListOption in settings to be a bottom sheet.
I'm not 100% satisfied with the implementation, but I don't have a lot of time this week and don't want to keep you waiting.

Screenshots:
Screenshot_1688042761
Screenshot_1688042766
Screenshot_1688042768
Screenshot_1688042774
Screenshot_1688042778


JulianPaulus:
Oh, also:
This was added, I presume for compatibility with the old lemmy api package, but I think it now breaks stuff.
```dart
defaultPostListingType = PostListingType.values.byName(prefs.getString("setting_general_default_listing_type")?.toLowerCase() ?? DEFAULT_LISTING_TYPE.name);
      defaultSortType = SortType.values.byName(prefs.getString("setting_general_default_sort_type")?.toLowerCase() ?? DEFAULT_SORT_TYPE.name);
``` 
For top-sorts, it now tries to look up enums like "topday" that don't exist. I guess a simple fix like removing the toLowerCase() should fix this again.

hjiangsu:
Ahh yeah - that was added in when I was migrating over to the other lemmy api package. I can take a closer look at it!

I'll merge it in for now and apply any fixes that are needed! Thanks so much for the help :D

PR #147, opened 2023-06-30
build android via docker

Fmstrat:
A build method using Docker. I'll be using this for dev to contribute going forward. If you would like, this could also be used for a GitHub Action for post-push compilation, but I'm not sure if that's a workflow you're looking for since you're also supporting iOS.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+scripts/docker*
+docker
+build
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@
 </p>
 
 <p align="center">
-   <a href="#contributing">Contributing</a> •
+  <a href="#contributing">Contributing</a> •
   <a href="#building-from-source">Building From Source</a> •
   <a href="#conventions">Conventions</a> •
   <a href="#related-packages">Related Packages</a>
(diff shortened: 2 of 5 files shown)


hjiangsu:
It looks good! Thanks so much!

> GitHub Action for post-push compilation

Hmm, I'm not too familiar with this, maybe you can elaborate?

Fmstrat:
> It looks good! Thanks so much!
> 
> > GitHub Action for post-push compilation
> 
> Hmm, I'm not too familiar with this, maybe you can elaborate?

It can be used to generate an APK and push to the releases automatically. That being said I see you have release management of some type already ;)

hjiangsu:
Oh I see! That could be useful in the longer term. Right now, releases are somewhat manual:
- I increase the pubspec version to the correct one
- I run the `dart scripts/build.dart` file to generate the IPA/APK files
- I upload those to GitHub

I also have to generate an archive on XCode to submit to TestFlight users

If theres a way to better do that, it might help!

Fmstrat:
Honestly the best way probably isn't to use what I'm doing locally, it's to use these popular actions from the marketplace:

The official Flutter action: https://github.com/marketplace/actions/flutter-action
Followed by create-release: https://github.com/marketplace/actions/create-release

And basically have them fire when you tag the main branch.

hjiangsu:
Hmm yeah, I've seen them but I haven't explored them too much. I would need to figure out how they work first 😅

PR #148, opened 2023-06-30
Totp

MrAntonS:
Ok, I've looking at code, but because of lack of Dart knowledge, I can't figure out, how can I connect Login Page to AuthStatus and change login widget if only TOTP is required? Can you give me some advise, current implementation working, but i would much prefer for the second page to appear if TOTP is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fixed issue where an error would be thrown if trying to access http pages - contribution from @vbh
 - Fixed issue with haptic feedback on comment actions
 - Fixed issue where reply button would cover actions and comments for short posts
+- Removed mark as read for mentions and replies that have already been read
 
 ## 0.2.1+8 - 2023-06-28
 ### Added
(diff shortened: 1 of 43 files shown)


hjiangsu:
Let me see if I have the question correct - TOTP implementation is done, but you want it to only show TOTP if the user has set up TOTP?

I haven't looked too much into how TOTP works on lemmy, but I'm guessing its a code you enter along with your login info in order to authenticate you?

Or rather, do you have to log in first, and then provide the code in order to allow log in successfully?

If its the first option, then I think its okay to leave the TOTP text field there. If its the second option, then theres a few things you can do.

Pressing Login will trigger the corresponding AuthBloc function to run. You can add some code into the LoginAttempt function which checks for a specific message from the server after login is attempted. (would have to see what the error is that the server sends back)

From that response, you can store some variable `requiresTotp` in AuthState so that you can then access it from the Login page. In the Login page, you can add in a BlocListener to listen for changes in AuthState for that `requiresTotp`, and then update the state of the widget to then show the text field!

I hope that clears up your questions! Let me know if you need a deeper explanation :D

MrAntonS:
Basically, it doesn't matter if you have totp. You can sent blank totp field in api request and everything would work, so you can merge it. I will tinker with it a bit, it's just my personal preference for totp to appear after you try to log in. Thank you for the explanation, I'll try it out tomorrow. 😄

hjiangsu:
I'll merge this into the develop branch for now to allow for some testing!

Issue #117, opened 2023-06-27
[Feature Request] Open the keyboard when pressing on the search button in the bottom bar


willbennett-kaltura:
**Is your feature request related to a problem? Please describe.**
Reaching all the way to the top is a little awkward on larger phones.

**Describe the solution you'd like**
It would be nice to have the keyboard already open when you click on the search button in the bottom bar.

micahmo:
I've seen multiple apps where a second tab on the search tab opens the keyboard (Spotify and IMDb come to mind). What do you think?

https://github.com/thunder-app/thunder/assets/7417301/58069da5-b4dd-4b18-979e-05fb45ba25f2

willbennett-kaltura:
> I've seen multiple apps where a second tab on the search tab opens the keyboard (Spotify and IMDb come to mind). What do you think?
> 
> https://github.com/thunder-app/thunder/assets/7417301/58069da5-b4dd-4b18-979e-05fb45ba25f2
> 
> 

I think that's a decent option, but in general, if you're already opening the search, and the page does nothing else, why not just open the keyboard immediately? It's one less click

hjiangsu:
This should be implemented in the latest release - to trigger the keyboard to open, tap on the search button again after navigating to it (second tap opens the keyboard)

For any changes to this functionality, feel free to open a new issue!

PR #152, opened 2023-06-30
#117 request focus search page

ggichure:
#117 request focus search page using `FocusManager`


hjiangsu:
It looks good! Just one question - why was pubspec.lock removed?

ggichure:
> why was pubspec.lock removed?

Force of habit.

micahmo:
@ggichure Do we still need this now that #494 is done?

hjiangsu:
I'll close this as there has been no activity and this should be added in with #494!

Issue #122, opened 2023-06-27
change app opens sidebar


Tealk:
**Description**  
The swipe gesture to switch the app opens the sidebar This is a little annoying when returning to the app It would be nice if you could exclude the lower area which is for the app control

**How to Reproduce**  
1) Open app
2) Switch the app with the swipe gesture
3) return to the app

**Device & App Version:**  
 - Device: Pixel6
 - OS: GraphenOS13
 - Version 0.2.1+8

CTalvio:
I have recorded a video demonstrating this issue. The side menu gesture is recognized, even when interacting with the gesture bar/navbar in android.

https://github.com/hjiangsu/thunder/assets/4365015/94b1d0cc-07d4-4490-9fa6-1e084bb58f43

hjiangsu:
Oh, hmm. I'll have to think about how to fix this since theres no easy fix I can immediately think of.

I could add in a setting which enables/disables this gesture that'll default to off unless you toggle it on.

It'll be hard to get only the swipe gesture to work on the top half of the navigation bar since the whole gesture is detected inside of that bottom nav bar

machinaeZER0:
What if the side panel gesture only triggered when you swiped right on the Home icon? Right now that consistently works for me without triggering any OS shenanigans, and it seems like the side panel can only trigger when Thunder's nav bar is displayed anyway?

CTalvio:
> What if the side panel gesture only triggered when you swiped right on the Home icon/the nav bar icons in general? Right now that consistently works for me without triggering any OS shenanigans, and it seems like the side panel can only trigger when Thunder's nav bar is displayed anyway?

How exactly would this resolve the gesture conflict for people using the gesture bar?

machinaeZER0:
Right now the sidebar gesture detection extends to the bottom of my screen, which is the problem - it overlaps with the area where OS gestures get triggered. If the Thunder gesture changes so that the area below the actual icons in its nav bar is a "dead zone" for app interaction, then I think that would make a huge difference. Testing on Android right now, and swiping across the nav bar icons opens the side panel without activating the OS app switcher.

Are we saying the same thing?

CTalvio:
[...]

PR #165, opened 2023-07-02
Fix/Workaround + Feature

bactaholic:
Provides workaround for #122 allowing you to toggle the swipe gestures in the general settings page, also adds double tap on nav bar feature to replace it for people like me who still want quick access. In my testing both work seamlessly separate or together.

Additionally, I'm not sure if it was removed on purpose but I really liked the original total post/comments counter from the original alpha. I added that back in to the profile header and added a toggle in the settings to swap post/comments counts to post/comments score. I can remove this if it was removed intentionally.

### DIFF
--- a/lib/account/widgets/account_header.dart
+++ b/lib/account/widgets/account_header.dart
@@ -1,4 +1,7 @@
+import 'dart:ffi';
+
 import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
 
 import 'package:intl/intl.dart';
 import 'package:cached_network_image/cached_network_image.dart';
@@ -8,7 +11,10 @@ import 'package:lemmy_api_client/v3.dart';
 import 'package:thunder/shared/icon_text.dart';
 import 'package:thunder/utils/instance.dart';
 
+import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+
 class AccountHeader extends StatelessWidget {
+
   final PersonViewSafe? accountInfo;
 
   const AccountHeader({super.key, this.accountInfo});
@@ -19,6 +25,9 @@ class AccountHeader extends StatelessWidget {
 
     final DateFormat formatter = DateFormat('yyyy-MM-dd');
     final String formatted = formatter.format(accountInfo!.person.published);
+
+    final bool profileSettings = context.read<ThunderBloc>().state.preferences?.getBool('setting_general_hide_profile_score') ?? true;
+    final totalScore = accountInfo!.counts.commentScore + accountInfo!.counts.postScore;
 
     return Padding(
       padding: const EdgeInsets.only(top: 8.0, left: 24.0, right: 24.0, bottom: 24.0),
@@ -57,6 +66,26 @@ class AccountHeader extends StatelessWidget {
                         text: (accountInfo?.person.published != null) ? '${formatted}' : '-',
                       ),
                     ],
+                  ),
+                  Row(
+                    children: [
+                    if (profileSettings == false)(
+                      IconText(
+                          icon: const Icon(
+                            Icons.arrow_upward_rounded,
+                            size: 16,
+                          ),
+                          text: '${(accountInfo?.counts.postScore).toString()} · ${(accountInfo?.counts.commentScore).toString()} · ${(totalScore).toString()}')
+                  )else if (profileSettings == true)(
+                        IconText(
+                          icon: const Icon(
+                            Icons.note_rounded,
+                            size:16,
+                          ),
+                          text: '${(accountInfo?.counts.postCount).toString()} · ${(accountInfo?.counts.commentCount)}'
+                        )
+                      )
+                    ]
                   ),
                 ],
               ),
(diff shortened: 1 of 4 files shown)


hjiangsu:
Thanks! I took a quick look over it and it seems to look good - just a question:
- How does the double-tap interact with the options in the bottom nav bar? would you have to double tap outside of the icons in the bottom nav bar to trigger the drawer to open/close?

> Additionally, I'm not sure if it was removed on purpose but I really liked the original total post/comments counter from the original alpha. I added that back in to the profile header and added a toggle in the settings to swap post/comments counts to post/comments score. I can remove this if it was removed intentionally.

I removed it because I was working on introducing a better user profile page! This is all happening in the `develop` branch, and will be merged into `main` when I release the upcoming version (which would be sometime tonight for me)

What I would suggest is to rebase this to the develop branch so that its up to date with all the changes with user profiles that I've added. The second reason why I removed it was because it didn't really seem too clear to me at least what the numbers were representing. I was trying to figure out a better way to represent those numbers (e.g., which one is post count and which one is comment count)

Aside from that, great work and thank you for the contributions! I'll most likely merge this in for the next-next release if thats okay (that would be v0.2.1+10)

bactaholic:
Howdy!

Thanks for the feedback, I will rebase this first thing in the morning as it's late and my laptop is dead.

As far as the double tap functionality, when turned on with the home page as the active window you can tap anywhere along the taskbar, including on the icons, and it will open the sidebar. I havent had any issues with it accidentally selecting a different page instead of activing the side bar, the tap delay seems to be long enough.
It also doesn't interfere with the existing swipe command. In fact, if you have both enabled, you can swipe to open and double tap to close, or vice versa. They appear to work interchangeably with no issues but I've only tested on one device, on Android.

I'll remove my profile score/counters from the rebase since you're working on the user profile. I've also thought about how to differentiate the different scores. I thought about simply adding 
"P: [postScore] C: [commentScore] T: [total]"
But I wasn't totally sure if that was clear enough. Will give it more thought. Hopefully I'll have some more fixes and ideas before the next-next release.

Thanks!

hjiangsu:
> Thanks for the feedback, I will rebase this first thing in the morning as it's late and my laptop is dead.

Actually - I think your changes might just be fine with the `develop` branch, so don't worry about rebasing it! I just merged in develop into main to prep the the upcoming release, and there's no conflicts that I see :D

> As far as the double tap functionality, when turned on with the home page as the active window you can tap anywhere along the taskbar, including on the icons, and it will open the sidebar. 

Ahh okay okay, that seems to make sense! Thanks!

> I'll remove my profile score/counters from the rebase since you're working on the user profile. 

I think for now, it might be best to remove it, but I do want to re-introduce the idea of showing the scores in a future version since I think it's useful to have that information. But yeah, as you mentioned, I do want to make it a bit more obvious what the numbers represent (either through icons, or some sort of text/visual indicator)

bactaholic:
@hjiangsu howdy! Removed the counters and associated switch in settings, also noticed I missed a line and it was causing the swipe gestures button to not retain its position after you switch pages, fixed that. Good to go.

hjiangsu:
Thanks! I'll check it out later when I have the time 😁

hjiangsu:
Looks good! If I were to nitpick, I would just suggest using more descriptive names for the settings. 

`swipeGestures` and `doubleTapGestures` is a bit difficult to know what exactly it affects! For example, maybe `bottomNavBarSwipeGestures` and `bottomNavBarDoubleTapGestures` would be more appropriate!

bactaholic:
Sure thing, easy fix. Good to go.

hjiangsu:
It looks good! Thanks :D

hjiangsu (approved):

hjiangsu:
Just a quick update on something - I've temporarily disabled the double tap to open sidebar. There was a issue where I noticed a longer delay between navigating the bottom nav bar when the double-tap gesture was implemented.

This is most likely because `GestureDetector` increased the delay of a single tap to be longer in order to detect cases of double-tapping. I'll take a look how we can maybe still have the same responsiveness feel while also including the double-tap functionality!

bactaholic:
> Just a quick update on something - I've temporarily disabled the double tap to open sidebar. There was a issue where I noticed a longer delay between navigating the bottom nav bar when the double-tap gesture was implemented.
> 
> This is most likely because `GestureDetector` increased the delay of a single tap to be longer in order to detect cases of double-tapping. I'll take a look how we can maybe still have the same responsiveness feel while also including the double-tap functionality!

Yeah I noticed that slight delay as well, I got used to it pretty fast but I can see how it would be annoying. I've been thinking about ways to get around it, maybe there is a custom delay timer we can set. I'll be looking into this more as well.

hjiangsu:
Yeah - I was thinking maybe a compromise to this issue is to completely remove the double-tap gesture when the option is off, as opposed to keeping the double-tap there but checking if the setting is set in the function.

This would simply just swap around the logic: If the double-tap setting is false, then set the GestureDetector double-tap function to null, otherwise have the function as it at the moment

I think this might fix the issue for those who have it disabled, but those who have the double-tap enabled will have to readjust themselves to have a slightly longer delay

hjiangsu:
Just an update - I have made the changes I mentioned previously, and they seem to be working as intended!
- When double tap is disabled, the delay between tapping an icon and the action is what it was previously
- When double tap is enabled, there is a slightly longer delay between tapping to allow the double tap gesture to register

PR #185, opened 2023-07-02
Add setting to show text content on cards

### LINKED ISSUE #182: Show content of text posts on card
Currently, text posts only show the title on their cards and not the content. I would like to have the option to see the content on the card as well. For longer text posts this should act as a preview, showing only the first few lines on the card. This would help get a better idea on the contents of a post while scrolling and would also make text posts stand out more between "bigger" posts with images or links. 

Here's a screenshot from Infinity for Reddit to show what I mean. I would like to go for something like this:

Screenshot_20230626_112234_Infinity

coslu:
Resolves #182.


hjiangsu:
Could you add in a screenshot to show how it looks like if possible? Otherwise, it looks good!

hjiangsu (approved):

Issue #171, opened 2023-07-02
Cannot download media using download media button


KhazAkar:
**Description**  
When trying to download media from post, nothing happens. By checking gallery app, also nothing new appears. Additionally, no indication is there that permissions are needed (I did that manually)

**How to Reproduce**  
1. Go to any post with media attached
2. Click on media to show media download button
3. Tap on this button
4. Nothing happens

**Expected Behavior**  
If permission is not granted, ask for it
If permission is granted, either allow for selecting saving folder using storage access framework (see Open Camera) or allow customization of folder name in DCIM folder.

**Screenshots**  
![Screenshot_20230702-095108_Permission controller](https://github.com/hjiangsu/thunder/assets/12693890/fb26aff2-e47b-4e4a-8269-c13b3660dcdc)

**Device & App Version:**  
 - Device: LG v30
 - OS: LOS 19 (Android 12)

**Additional context**
Fresh install, apk from github. Giving permission manually does not help.

Poolitzer:
Worked fine for me on android 13. Fresh install, clicked on the download button, asked for permission and the button turned into a checkmark, the photo appeared in Pictures, top level folder.

KhazAkar:
@Poolitzer interesting, worked for you on A13, not for me on A12 - not. What device BTW?

Poolitzer:
Pixel 4a 5G

hjiangsu:
I wonder if there was a change to how Android handles permissions between the versions. I'm not too sure off the top of my head what could be causing this

hjiangsu:
A fix for this from @minicit may have resolved this issue - I have released a pre-release version v0.2.1-alpha.1+11 which contains these fixes. Let me know if its working for you or if there are still issues with this!

KhazAkar:
@hjiangsu tested it and seems to work!
1. Access request work
2. Download works
3. Button changes from download button to tick
4. Only thing missing for me is where it saves. I found it in Pictures folder, but it's not indicated when downloading, for example as toast notification (on the bottom)

Thanks a lot! :)

hjiangsu:
Closing as its been released in v0.2.1+11!

PR #203, opened 2023-07-03
Fixed Issue #171 Download not working

minicit:
Fixes Issue https://github.com/hjiangsu/thunder/issues/171

Issue is in permission handler and explained here: https://github.com/Baseflow/flutter-permission-handler/issues/944

Not sure if I have to add anything in the Android Manifest but I don't think so.

Tested on Android Emulator for Android 11 & Android 13 (API Level 30 and 34)

If I made any mistakes with the code or pull request please let me know, I am new to contributing :)


hjiangsu:
It looks good! I'll do a physical test with an iOS device later to make sure it still works, but thanks for contributing!

PR #208, opened 2023-07-03
Fix image download button appearance on light mode

coslu:
Fixes the image download button being invisible on light mode. Its color was taken from the theme, which made it a dark color on light themes even though the background was set to black. I put the IconButton inside an AppBar. This also makes the status bar visible on Android, fixes the download button sometimes overlapping with the image, and adds a back button. Here is the before and after:

Screenshot_20230703_234033  Screenshot_20230703_234525

### DIFF
--- a/lib/shared/image_viewer.dart
+++ b/lib/shared/image_viewer.dart
@@ -41,16 +41,12 @@ class _ImageViewerState extends State<ImageViewer> {
 
   @override
   Widget build(BuildContext context) {
-    final theme = Theme.of(context);
     return Scaffold(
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):

hjiangsu:
Looks good! Thanks 😄

hjiangsu:
Wait, just double checking - this still works with the swipe gestures to dismiss the image correct?

coslu:
Yes, that still works.

hjiangsu:
Alright, thanks for confirming!

PR #210, opened 2023-07-03
use external browser setting in comments

Fmstrat:
Fixes https://github.com/hjiangsu/thunder/issues/144


hjiangsu:
Man, what a silly mistake I did there - thanks for fixing it!

PR #212, opened 2023-07-04
Added filtering of posts that already exist in view to prevent duplicates

ajsosa:
This is accomplished via a `Set<int> postIds` we now maintain along side the `postView` list in `community_state.dart` that we can check against before adding new posts to `postView`.

I'm still learning my way around Dart, Flutter, and this project. So feel free to let me know if I did something dumb :)

### DIFF
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -229,11 +229,17 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
             // Parse the posts and add in media information which is used elsewhere in the app
             List<PostViewMedia> formattedPosts = await parsePostViews(posts);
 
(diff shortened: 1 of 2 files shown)


hjiangsu (commented):
I looks pretty good! One thing I would suggest is to declare the types of the variables so that it makes it cleaner and easier to work with. For example:

```
var id = postMedia.postView.post.id;
```

In this case I would change it to `int id = postMedia.postView.post.id;` since I know the `id` is type int.

This line `for (var postMedia in postMedias)` could be changed to `for (PostViewMedia postMedia in postMedias)` for more clarity!

I'm also fairly new with Dart/Flutter, so its a learning experience for me too

ajsosa:
Done!

hjiangsu:
Looking good, thanks :D

Issue #207, opened 2023-07-03
Add refresh option other than scroll up


CTalvio:
The only way to trigger a feed refresh is currently to scroll all the way up.

My vote is for a refresh button next to the sort button in the top right, same as in inbox. it should also appear on community pages and for comments.

Also add pull down to refresh to inbox.

bactaholic:
Been working on this issue, I successfully added a manual refresh button to the home screen, just working on getting a pull to refresh for the inbox, but I'm running into a lot of problems getting the refresh to activate. Tried using `RefreshIndicator()` when I tell it to `onRefresh: context.read<InboxBloc>().add(const GetInboxEvent());` it throws an error.

Removing my work for the inbox until I can get something working, but I'll commit my other changes so hopefully we can get the refresh button on the home screen in the next release.

bactaholic:
pull request relating to this issue, #213

hjiangsu:
@bactaholic I'll assign this to you for now if thats okay!

bactaholic:
@hjiangsu sure thing!

CTalvio:
I spotted the refresh button in some new places. Nice.

I also noticed, that it is not possible to refresh on the comments and saved tabs in the profile view.

bactaholic:
@CTalvio howdy, yeah I noticed it's missing from a few places as well. I'll be going through all the pages and adding refresh options where needed, just finished implementing pull-to-refresh in the inbox, but I gotta clean up my code and add a quick fix for an error when refreshing within communities, then I'll be moving onto adding refresh to all pages.

krestenlaust:
Shouldn't this have an "fixed-in-upcoming-release"-tag?

CTalvio:
@bactaholic Bug report for you, when in "All" I noticed the button goes back to "Local posts" when used, while pull down to refresh does not.

bactaholic:
> @bactaholic Bug report for you, when in "All" I noticed the button goes back to "Local posts" when used, while pull down to refresh does not.

@CTalvio Does the behavior match that of #216 by chance? Sounds similar, if so, I have that fixed in a pull request that was successfully merged a few days ago, I imagine it will be in the release version soon.

CTalvio:
I'm on the latest alpha. That particular issue is already fixed.

Essentially, refreshing does not just refresh the current feed, it also resets to your default, be that local/subbed/all.

bactaholic:
> I'm on the latest alpha. That particular issue is already fixed.
[...]

PR #213, opened 2023-07-04
Added manual refresh button to home screen

bactaholic:
Pertaining to issue #207, added refresh button to home screen. Still working on pull to refresh in Inbox but its giving me problems, also tried adding a "Long Press on Nav Bar" to refresh function but ran into issues with that as well. 

Will keep working on those, but I wanted to commit this change just in case it could get accepted before the next release to solve that enhancement request.

Please ignore my string of commits, was running tests earlier and then accidentally stashed my changes.

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -1,202 +1,47 @@
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:lemmy_api_client/v3.dart';
 
-import 'package:thunder/account/bloc/account_bloc.dart' as account_bloc;
-import 'package:thunder/community/bloc/community_bloc.dart';
-import 'package:thunder/community/pages/create_post_page.dart';
-import 'package:thunder/community/widgets/community_drawer.dart';
-import 'package:thunder/community/widgets/post_card_list.dart';
-import 'package:thunder/core/auth/bloc/auth_bloc.dart';
-import 'package:thunder/shared/error_message.dart';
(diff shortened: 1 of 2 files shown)


hjiangsu:
Looks good! Thanks

I'm guessing the issues that you're running into is probably something along the lines of "Could not find Provider InboxBloc".

This is because the call to `context.read<InboxBloc>()` is outside of the scope of where `InboxBloc` was provided. This is the line that provides the InboxBloc to all its child widgets:

```
 BlocProvider(create: (context) => InboxBloc()),
```

The child context that calls this has to be within the scope of this `BlocProvider` for it to find `InboxBloc`. Not too sure if these are the errors you're encountering, but feel free to let me know otherwise and maybe I can provide some insight on it!

bactaholic:
> Looks good! Thanks
> 
> I'm guessing the issues that you're running into is probably something along the lines of "Could not find Provider InboxBloc".
> 
> This is because the call to `context.read<InboxBloc>()` is outside of the scope of where `InboxBloc` was provided. This is the line that provides the InboxBloc to all its child widgets:
> 
> ```
>  BlocProvider(create: (context) => InboxBloc()),
> ```
> 
> The child context that calls this has to be within the scope of this `BlocProvider` for it to find `InboxBloc`. Not too sure if these are the errors you're encountering, but feel free to let me know otherwise and maybe I can provide some insight on it!

I meant to reply to this as well, the issue i'm running into with the inbox is "The body might complete normally, causing 'null' to be returned, but the return type, 'Future<void>', is a potentially non-nullable type.". Too tired to keep looking for a solution, will test more in the morning with a fresh brain.
Can't remember what the error was for the long-tap to refresh feature. Will look more into that tomorrow.

PR #222, opened 2023-07-04
Title first and edge-to-edge images

Fmstrat:
Adds two options under Post, one for putting the title above the image/content, and another for edge-to-edge images. I've adjusted padding to look like it was before if those options are turned off, but altered a bit to fit with the changes when they are on.

Updates for: https://github.com/hjiangsu/thunder/issues/136


hjiangsu:
It would be good to see some screenshots of the different views! Aside from that, it looks good!

Fmstrat:
Before:
Screenshot_1688560596

After (both settings toggled):
Screenshot_1688560634

hjiangsu:
Looks good!

PR #224, opened 2023-07-04
gray body text in post view

Fmstrat:
A first-step for https://github.com/hjiangsu/thunder/issues/223 is to blend out the body content in the post view. This seems to be a good balance for light/dark themes.


hjiangsu:
Could you add in screenshots if possible? That would be helpful!

Fmstrat:
Before:
Screenshot_1688560397

Screenshot_1688560508


After:
Screenshot_1688560444

Screenshot_1688560484

hjiangsu:
Thanks!

PR #234, opened 2023-07-05
Improve Material 3 Theme

TomBursch:
This PR improves the material 3 theming of the app. It includes two main changes:
1. Generate the Theme using a `ColorScheme` → I think this is simpler than using a package for it and I prefer the theme generated by it (especially for chips since it only uses one color and not a primary and secondary)
2. Migrate `BottomNavigationBar` to `NavigationBar`

I attached some screenshots. Feel free to just close this PR if you're not happy with it, I know this is purely subjective. 
Thanks for all the work you did!

image
image

### DIFF
--- a/lib/account/pages/login_page.dart
+++ b/lib/account/pages/login_page.dart
@@ -158,13 +158,9 @@ class _LoginPageState extends State<LoginPage> {
                 enableSuggestions: false,
               ),
               const SizedBox(height: 32.0),
-              ElevatedButton(
+              FilledButton(
                 style: ElevatedButton.styleFrom(
                   minimumSize: const Size.fromHeight(60),
-                  backgroundColor: theme.colorScheme.primary,
-                  textStyle: theme.textTheme.titleMedium?.copyWith(
-                    color: theme.colorScheme.onPrimary,
-                  ),
                 ),
                 onPressed: (_usernameTextEditingController.text.isNotEmpty && _passwordTextEditingController.text.isNotEmpty && _instanceTextEditingController.text.isNotEmpty)
                     ? () {
--- a/lib/community/widgets/post_card.dart
+++ b/lib/community/widgets/post_card.dart
@@ -192,7 +192,11 @@ class _PostCardState extends State<PostCard> {
             Divider(
               height: 1.0,
               thickness: 4.0,
-              color: useDarkTheme ? theme.colorScheme.background.lighten(7) : theme.colorScheme.background.darken(7),
+              color: ElevationOverlay.applySurfaceTint(
+                Theme.of(context).colorScheme.surface,
+                Theme.of(context).colorScheme.surfaceTint,
+                1,
+              ),
             ),
             InkWell(
               child: useCompactView
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:dynamic_color/dynamic_color.dart';
+import 'package:flex_color_scheme/flex_color_scheme.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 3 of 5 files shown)


hjiangsu:
First of all, thanks for creating a PR! I really do appreciate it!

I'm open to switching back to theming without an external package, but I'm just wondering about a couple things:

- How does it look when its in pure black mode? From what I briefly read in the `flex_color_scheme` package, they seem to do it by darkening all the colours within the themes by a couple points. This allows everything to become a bit more muted in terms of colours which might be a preference that people like about the black theme
- How easily customizable are the themes with the use of the `flex_color_scheme` package?

I do like how much more customizable `flex_color_scheme` seems to be with regards to making new themes (and already having preset themes). This would be especially useful for those who may not have Material You compatibility (older Android devices and iOS users)

I'm thinking of a future feature where theres a whole bunch of pre-generated themes that the user can pick from if they don't have Material You support (especially iOS) or just simply want to try on a new look

> Migrate BottomNavigationBar to NavigationBar

I do like this change - I personally had it initially set to `NavigationBar` but switched over because of some issues before (I can't remember what the issue was)

Fmstrat:
> I'm thinking of a future feature where theres a whole bunch of pre-generated themes that the user can pick from if they don't have Material You support (especially iOS) or just simply want to try on a new look
> 

Seems like this PR may be exactly what you're looking for, then. I.E. you could use the playground (https://rydmike.com/flexcolorscheme/themesplayground-v7-1) to build out whatever themes you would like.

hjiangsu:
> Seems like this PR may be exactly what you're looking for, then. I.E. you could use the playground (https://rydmike.com/flexcolorscheme/themesplayground-v7-1) to build out whatever themes you would like.

Yeah, that's what I was looking at too - this way, we can have the community also contribute theme ideas without requiring them to actually know how to code

> I've added it back, but changed the design slightly to match what they call the default material 3 theme. This fixes the main thing I didn't like about it (the different colored chips).

I think this works! We can build off of this in the future as well, and have a higher level of customization like I mentioned previously

hjiangsu:
Hmm, I'm running this and came across a few visual issues. I might revert back this PR for now until some of these issues are resolved:

General:
- Inputs have a different background colour than its surrounding container (e.g., search bar)
- When switching across multiple tabs at the `NavigationBar`, there is a weird visual artifact
  - This is most likely due to the animation that occurs when switching between the sections. For this one, maybe we can just disable that animation and it might fix that

In dark mode:
- The divider colour between posts has been greatly reduced which makes it much harder to distinguish between posts

simulator_screenshot_C44CEECC-9E7E-4738-8407-2787076EDBF9

simulator_screenshot_E70A0A18-55FE-4A08-BED7-AB951406D07E

comment_voting

TomBursch:
> Inputs have a different background colour than its surrounding container (e.g., search bar)
Hmm strange I don't have that issue. Could be that `flex_color_scheme` does some adaptive theming for ios?

> When switching across multiple tabs at the NavigationBar, there is a weird visual artifact
This is due to `onPageChanged: (index) => setState(() => selectedPageIndex = index)` in the PageView which sets the selectedPageIndex to all intermediate values

hjiangsu:
> Hmm strange I don't have that issue. Could be that flex_color_scheme does some adaptive theming for ios?

Possibly - I have encountered cases before where the theming changes depending on iOS/Android, and even on Linux/Mac/Windows. I think we can still continue to do some more refinements here to check for inconsistencies across platforms.

Alternatively, we can add in a toggle specifically for Android to push the theme changes you've applied using `Platform.isAndroid` flag!

> This is due to onPageChanged: (index) => setState(() => selectedPageIndex = index) in the PageView which sets the selectedPageIndex to all intermediate values

Good catch there - seems like that might be be the case!

PR #241, opened 2023-07-06
Use titles on community pages, add text wrapping on overflow

CTalvio:
Changed community header to using community titles, and added wrapping for long ones

Screenshot_20230706_035007


hjiangsu:
Targeting develop instead of main for the branch

hjiangsu:
It looks good otherwise, thanks so much!

Issue #181, opened 2023-07-02
[Feature Request] Share images through Android Sharesheet


jdevowe:
Having used RIF is Fun until recently, I would prefer to share images in The directly to another app via the Android Sharesheet instead of a link to the lemmy post. Current option is downloading from Thunder and sharing from an external app (such as sharing from Gallery to a contact in Messages).

Possible implementation: Select share on a post, tap and hold on a post title, or tap and hold on the image itself to have options appear in a menu to share a link to the post or the image itself. "Share link,"" "Share image".

jdevowe:
Looks promising! Can't wait to try it out in the release that features it.

micahmo:
Hey @hjiangsu, thanks for adding this in the latest alpha! It works great and it's really nice to have.

My only comment would be that I was expecting it to share the media itself, not a link to the media. I suppose that may require downloading it first, but I will say some Reddit apps have that function.

Thanks again for the update!

hjiangsu:
> My only comment would be that I was expecting it to share the media itself, not a link to the media. I suppose that may require downloading it first, but I will say some Reddit apps have that function.

Oops, hmm - I can take a look to see if theres a way to do that! Do you happen to know if the other Reddit apps had to download the media first before sharing it? That'll help me figure this out a bit better

micahmo:
> Oops

What you added is great, btw! Didn't mean to take anything away from that!

> Do you happen to know if the other Reddit apps had to download the media first before sharing it?

Unfortunately I don't know offhand. The one I used wasn't open-source but I could try to find one that is. I suppose the media is in memory somewhere, but maybe you're just assigning a URL to a control and letting it handle that?

hjiangsu:
Alright - I'll do some tests when I have the time and see if its possible to do what you described!

jdevowe:
> Alright - I'll do some tests when I have the time and see if its possible to do what you described!

I find some info regarding using three Android Sharesheet for sending binary data such as images. I'm in the early stages of learning to code and I think this is beyond me at the moment.

https://developer.android.com/training/sharing/send

Under heading "Send binary data."

hjiangsu:
Thanks - that'll come in a bit useful!

hjiangsu:
Closing this issue since it has been released in v0.2.1+11!

PR #244, opened 2023-07-06
Share media

micahmo:
This PR attempts to address #181 more fully by allowing us to share the media (e.g., picture) directly as a file rather than as a link to the media source.

### Notes
* We still have the "Share Link" option for media, which now shares the link to the media.
* I did end up having to download the media before sharing; however, I used a temporary directory to do so, and I cleaned up the temp files. This seems to be a common practice ([here](https://stackoverflow.com/questions/71175980/how-to-share-image-file-from-flutter-app-to-other-apps/71176265#71176265) and [here](https://stackoverflow.com/questions/61385937/how-to-share-image-from-apiurl-in-flutter/71916658#71916658)).
* I only tested on Android.
* Do we need any error checking (try/catch) around the media downloading?
* Since we download the media between the time the user clicks on "Share Media" and the time the share sheet appears, I'm concerned that this might introduce a delay when the media is large and/or the network connection is slow. Do we need a loading indicator during this time, and if so, what would that look like?

### Demo

* Share Post
* Share Link
* Share Media

https://github.com/hjiangsu/thunder/assets/7417301/e6a4560e-267b-4b36-ba69-570ab36c54b3

### DIFF
--- a/lib/community/utils/post_card_action_helpers.dart
+++ b/lib/community/utils/post_card_action_helpers.dart
@@ -1,11 +1,9 @@
 import 'dart:io';
 
 import 'package:flutter/material.dart';
+import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 
-import 'package:http/http.dart' as http;
 import 'package:flutter_bloc/flutter_bloc.dart';
-import 'package:path/path.dart';
-import 'package:path_provider/path_provider.dart';
 import 'package:share_plus/share_plus.dart';
 
 import 'package:thunder/account/bloc/account_bloc.dart';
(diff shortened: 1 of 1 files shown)


krestenlaust:
> We still have the "Share Link" option for media, which now shares the link to the media.

I'm unsure as to what this means. Does the 'Share Link' still do the same on a text post as a media post?

> ... by allowing us to share the media (e.g., picture) directly as a file rather than as a link to the media source.

I think it would be nice to add a miniscule footer to image, similar to how reddit did it. Though I'm not sure if I'm the only one who thinks that. Since there's already a way to just share the media without attributing the post (downloading and sharing).

CTalvio:
I would definitely not want a footer. Just the actual file, as-is.

hjiangsu:
Hey! Thanks for the PR - I have a couple of suggestions that you can apply if you think they're good!

One way to potentially reduce the time it takes to download the media, and also skip having to save it to a temporary directory, is to use the cached image (if it already exists)

For example, in `image_viewer`, we fetch the image like this before we download it:

```dart
File file = await DefaultCacheManager().getSingleFile(url);
```

This will allow the media download to be faster in some cases, where we already have the cached image.

Alternatively, since we're using `extended_image` as the way to handle all the image thumbnails and image previews, you could do something like this: taken from https://pub.dev/packages/extended_image#save-network

```dart
///save network image to photo
Future<bool> saveNetworkImageToPhoto(String url, {bool useCache: true}) async {
  var data = await getNetworkImageData(url, useCache: useCache);
  var filePath = await ImagePickerSaver.saveFile(fileData: data);
  return filePath != null && filePath != "";
}
```

Note: I'm not too sure if `extended_image` stores its cached files in the same directory as `DefaultCacheManager`, so that might have to be double checked. Otherwise, using `DefaultCacheManager` may not actually be faster if they don't share the same directory

krestenlaust:
I'll also suggest renaming "Share link" to "Share", and leaving "Share Media" as-is, to leave sharing the post as the neutral option.

hjiangsu:
> Do we need any error checking (try/catch) around the media downloading?


> Since we download the media between the time the user clicks on "Share Media" and the time the share sheet appears, I'm concerned that this might introduce a delay when the media is large and/or the network connection is slow. Do we need a loading indicator during this time, and if so, what would that look like?

I think we can tackle both of these at once. When we share the media, we can show a toast notification (SnackBar) to let the user know that we are currently fetching the image. We can use a try catch for the image fetching, so that if it fails, we show a secondary toast notifying its failure

micahmo:
> I'm unsure as to what this means. Does the 'Share Link' still do the same on a text post as a media post?

@krestenlaust

* On text posts, there is no "Share Link" option.
* On link posts, "Share Link" shares the external link (e.g., news article).
* On image posts, "Share Link" shares a link to the media source (i.e., the hosting provider), whether that's on Lemmy itself or something else like Imgur.

You can watch the demo video I posted to see how "Share Post", "Share Media", and "Share Link" function on an image post.

---

> I would definitely not want a footer. Just the actual file, as-is.

@CTalvio Agreed!

---

> One way to potentially reduce the time it takes to download the media, and also skip having to save it to a temporary directory, is to use the cached image (if it already exists)

@hjiangsu Thanks for this suggestion! I like the idea of using the `DefaultCacheManager()`. We can share the same mechanism with the `image_viewer` so whichever one happens first can speed up the second; it significantly shortens the code; and it removes the cleanup responsibility from out of the image sharing code. I have made this change in 0b82acd77fc621e4d91d4fe897d9ecdf57aa84a4.

---

> I'll also suggest renaming "Share link" to "Share", and leaving "Share Media" as-is, to leave sharing the actual post as the neutral option.

@krestenlaust I think this would introduce some confusion, as the options would be "Share Post", "Share Media", and an ambigious "Share". How would the user know what the last option will do?

---

> When we share the media, we can show a toast notification (SnackBar) to let the user know that we are currently fetching the image. We can use a try catch for the image fetching, so that if it fails, we show a secondary toast notifying its failure

@hjiangsu Great idea! I have implemented this in 0b82acd77fc621e4d91d4fe897d9ecdf57aa84a4. Question: Is it ok to print the exception message directly in the toast?

---

### Demos

#### Uncached Image

https://github.com/hjiangsu/thunder/assets/7417301/479a5f37-6b67-4052-95f1-b95c6b3fae2f

#### Cached Image
https://github.com/hjiangsu/thunder/assets/7417301/17760847-9370-4163-9a62-8fb9af00dee3

#### Error
https://github.com/hjiangsu/thunder/assets/7417301/96febffd-6987-4cfc-994e-97f78b035813

---

P.S. It looks like there is no localization yet (strings are hard-coded). Is that right?

hjiangsu:
That looks to be pretty good! Nice job on it

> Question: Is it ok to print the exception message directly in the toast?

I think its okay for now because this is still in what I consider the alpha stages. As we progress, we should probably abstract the exception messages into a more human readable form, but for now, having the exact exception messages allows us to fix issues faster for those who describe the exception message in full

> P.S. It looks like there is no localization yet (strings are hard-coded). Is that right?

Yeah, localization is not yet here (would definitely want to get on that in the future, but it seems like such a large task at the moment and there are some more pressing issues to resolve)

I'll do a quick look over the code and apply any suggestions I see fit!

hjiangsu:
Looks good - I'll go ahead and merge this in. If there are further discussions on this, feel free to open that up in the Discussions page!

Issue #243, opened 2023-07-06
All screens should go back to home feed before exiting app


jordan1776:
**Is your feature request related to a problem? Please describe.**
This enhancement expands on #180. When pressing back on the main screens of search, profile, inbox, and settings the app is exited.

**Describe the solution you'd like**
When on those screens, the back button should go to the home feed. Then pressing back again should exit the app.

**Describe alternatives you've considered**
A double tap to exit or popup would be welcome on the home feed.

tunglith:
Strongly agree with that one. Many times I go back from the settings or profile and am thrown out of the app.

bactaholic:
If accepted this should be partially fixed by #245, app now asks to confirm before exit and if a back button press/swipe is made again it dismisses the exit warning. Still working on getting it to navigate back to home if using the back button/swipe from any navbar page but the first.

Issue #198, opened 2023-07-03
Swipe goes back when actions are disabled


micahmo:
**Is your feature request related to a problem? Please describe.**
When swiping on posts/comments are disabled, swiping does nothing. We can take advantage of this additional operation.

**Describe the solution you'd like**
It would be nice if swiping from left-to-right in the comments view would go back to the main posts view. It would be nice if swiping from left-to-right in the main posts view would open the hamburger menu. Both of these things are implemented by Relay for Reddit and Jerboa, if you want examples.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

jordan1776:
>It would be nice if swiping from left-to-right in the main posts view would open the hamburger menu. 

This is already implemented as a partial swipe on the left edge.

micahmo:
> This is already implemented as a partial swipe on the left edge.

I can't get this to work, can you demonstrate? On Android, if I swipe from the very left, it goes back. I would prefer a swipe somewhere in the middle.

jordan1776:
That is what I mean, the swipe on the very left edge. Maybe it should be more forgiving by extending it a few pixels right so it should be possible to swipe it with your right thumb, but I actually prefer the something like the current behavior over going half screen or Fullscreen with gestures on.

micahmo:
This has been working much more consistently as of #1413. Any objections to closing?

hjiangsu:
No objections!

Issue #149, opened 2023-06-30
Tap Home button to scroll to top of feed/return to main feed?


machinaeZER0:
In most reddit/lemmy apps I've tried, tapping the home button serves a few additional purposes beyond returning to the feed view from a different nav bar section - most of the time it can be:

- pressed while in feed view to auto-scroll back to the top
- pressed while scrolling through a specific community to return to the "main feed"

If possible it would be great to see these functions implemented in Thunder at some point (if you also feel they would be useful, of course!)

kaibagley:
Since this _seems_ like a simple thing to implement, I'm keen to try to do this. However I'm **very** new to dart/flutter and have only been learning it to hopefully contribute a bit to this amazing app.

That being said, if anyone has any tips off the top of their heads to do this please let me know, who knows how long it'll take me to figure it out

hjiangsu:
This one is a little bit more tricky to figure out than it initially seems - the bottom nav bar is not directly connected to the community/feed screen so you would have to tie in some logic to be able to control the scroll of the community/feed screen from the main bottom nav bar screen

The bottom app bar is in the main view, and there's a subview that shows the feed. In order to implement this, you would have to 
- pass a `ScrollController` from the feed/community page up to that main view
- detect when the user taps on the bottom nav bar option when they are currently in the feed view
- use that `ScrollController` to move the scroll position back up

However, do feel free to play around with it and see if you can get it to work!

kaibagley:
By the way if anyone wants to just get this done that's fine, I don't want to hold features up just for the sake of my own learning

kaibagley:
Is there a way to use Bloc for this? Rather than passing a `ScrollController` around? I'm having trouble understanding the best way to do this...

hjiangsu:
You could use Bloc for this as well. Here's an initial and very rough idea:

The BottomNavBar is connected to the ThunderBloc
- When you perform a secondary tap on the bottomNavBar icon while you're in the feed view, you can create a new event `context.read<ThunderBloc>().add(OnScrollToTopEvent());
- Within the Bloc event, you can set the add in a new state variable `scrollToTopId` which is a randomly generated string to treat as a token to check for changes to that variable
- Within the main view, add in a BlocListener which listens to that specific state variable (you can use `listenWhen`) This will give you access to the previous and current state of the bloc
- If the state (token) changes for the `scrollToTopId`, then trigger the scrollController to go back up to the top

Issue #75, opened 2023-06-25
Up/Down arrows for navigating comment threads


machinaeZER0:
One feature I found very useful in Boost for reddit (that recently got added to Jerboa) is the presence of navigation buttons in a post's comment view - using the Up and Down arrows jumps you between parent threads, which makes comments very easy to navigate.

In Boost this feature is even more useful - long-pressing the Up arrow brings you immediately to the top of the post you're viewing, which I've always really appreciated!

Screenshot attached showing this feature as implemented in Boost, if that's helpful - in their last update they even added a handy search function in addition to the navigation arrows!
![Screenshot_20230625-112504](https://github.com/hjiangsu/thunder/assets/3624023/a33d3243-1943-49ae-9ce3-64fc038665dd)

CTalvio:
Would also be cool if the volume buttons could be used for this.

PR #245, opened 2023-07-06
added swipe to return on post body + scroll to top buttons

bactaholic:
Added function to return to home screen when you swipe left to right on post body, doesn't effect existing gesture controls for post voting on home screen or comment voting.

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -95,7 +95,7 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
                         icon: const Icon(Icons.refresh_rounded, semanticLabel: 'Refresh'),
                         onPressed: () {
                           HapticFeedback.mediumImpact();
-                          return context.read<CommunityBloc>().add(GetCommunityPostsEvent(reset: true, sortType: sortType, communityId: state.communityId));
+                          return context.read<CommunityBloc>().add(GetCommunityPostsEvent(reset: true, sortType: sortType, communityId: state.communityId, listingType: state.listingType));
                         }),
                     IconButton(
                         icon: Icon(sortTypeIcon, semanticLabel: 'Sort By'),
--- a/lib/community/widgets/post_card.dart
+++ b/lib/community/widgets/post_card.dart
@@ -64,6 +64,8 @@ class _PostCardState extends State<PostCard> {
     isUserLoggedIn = context.read<AuthBloc>().state.isLoggedIn;
   }
 
+  final GlobalKey<ScaffoldState> _feedScaffoldKey = GlobalKey<ScaffoldState>();
+
   @override
   Widget build(BuildContext context) {
     final ThemeData theme = Theme.of(context);
@@ -72,7 +74,13 @@ class _PostCardState extends State<PostCard> {
     VoteType? myVote = widget.postViewMedia.postView.myVote;
     bool saved = widget.postViewMedia.postView.saved;
 
-    return Listener(
+    return GestureDetector(
+      onHorizontalDragEnd: (details) {
+        if (details.primaryVelocity! > 0) {
+          _feedScaffoldKey.currentState?.openDrawer();
+        }
+      },
+    child: Listener(
       behavior: HitTestBehavior.opaque,
       onPointerDown: (event) => {},
       onPointerUp: (event) => {
@@ -211,6 +219,7 @@ class _PostCardState extends State<PostCard> {
           ],
         ),
       ),
+    ),
     );
   }
 }
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -137,8 +137,8 @@ class _PostCardListState extends State<PostCardList> {
             ),
             if (_showReturnToTopButton)
             Positioned(
-              bottom: 16.0,
-              right: 16.0,
+              bottom: 16,
+              left: 20,
               child: FloatingActionButton(
                 onPressed: () {
                   _scrollController.animateTo(
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -44,6 +44,7 @@ class PostCardList extends StatefulWidget {
 
 class _PostCardListState extends State<PostCardList> {
   final _scrollController = ScrollController(initialScrollOffset: 0);
+  bool _showReturnToTopButton = false;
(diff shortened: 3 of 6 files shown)


micahmo:
I believe this is related to #198.

bactaholic:
@micahmo Yes and no, I haven't been able to get it to open the side bar when swiping left to right on the home page, just from post view to exit to the home page. Been working on it but for some reason the drawer won't open when activated, no errors thrown, will take some more thinking before I can say that problem is solved.

bactaholic:
ALSO ADDED:

-Scroll to top button for home page
-Scroll to top for post view
(haven't figured out how to add it to the community view yet)

These two can be used in conjunction to quickly return home. Tap return button, swipe to return to homepage.

bactaholic:
ANOTHER ADD:

-Confirmation to close app when OS "back" button or swipe gesture is used.

This should prevent accidental exits as requested in #243

micahmo:
> from post view to exit to the home page

That's actually the main thing I wanted from #198. Since then, I discovered that I can open the hamburger menu with a swipe on the bottom nav bar, so the other request is much less important.

hjiangsu:
Nice job on all the features so far! I haven't had the chance to look at this fully, but I do have a couple of questions:

- For the return to top, which views does it currently affect? It looks like its on the `post_card_list` and `post_page` so I'm guessing that its within a given post, and also anywhere that PostCardList shows up (viewing a specific, user, viewing a community, the account profile page, viewing home feeds)
- Could you take some screenshots to help me visualize the placement of the buttons?

> Confirmation to close app when OS "back" button or swipe gesture is used.

I'm wondering if we should do a double tap to close rather than an alert dialog to prompt the user
- What we can do instead of an alert dialog is to show a snackbar/toast to let the user know to tap the back button again to exit
- Having a snackbar/toast seems less disruptive to the experience than a full alert dialog

I would like to get others to input their thoughts/opinions into this!

bactaholic:
Made some adjustments to make sure everything lines up with existing buttons on all pages

https://imgur.com/a/h5Zrh9e

Page Functionality:
-Works on Feed
-Works in Post View
-Works in User Page
-Works in Community
-Does not work in Account Page comments when scrolling, I think the account header pushes the button down, can't figure out how to adjust button position based on current page

Then the swipe back on post option works with it to quickly scroll up then swipe out, also warning for app exit prevents accidental exits when doing this.

bactaholic:
> I'm wondering if we should do a double tap to close rather than an alert dialog to prompt the user
> 
>     * What we can do instead of an alert dialog is to show a snackbar/toast to let the user know to tap the back button again to exit
> 
>     * Having a snackbar/toast seems less disruptive to the experience than a full alert dialog
> 
> 
> I would like to get others to input their thoughts/opinions into this!

Forgot to reply to this, I saw Sync for Reddit has that functionality, to swipe back twice to exit. Was looking around at other apps this morning to see how each one handles certain things. Backing a swipe-twice-to-exit would allow us to preserve the double tap to open side bar gesture, and I think its more clean as a solution. I can work on getting that added instead.

bactaholic:
Alright, @hjiangsu  I swapped the dialog box with a snackbar which pops up for 3.5 seconds, and after 3.5 seconds resets the counter for how many times you've swiped back after 3.5 seconds so if the snackbar expires, it will confirm again later next time you swipe back instead of just exiting. Pushing one more commit to up the time from 3 seconds to 3.5...

https://imgur.com/a/mJNdgDD

bactaholic:
-added fix for error mentioned in  #207

refreshing from All page now retains All page instead of reverting to Local page

machinaeZER0:
This is really cool! I will say, it feels like the FAB is doing some work that (in some cases) could be done by existing buttons (or buttons planned for the future):

- In other apps, hitting the home button on the nav bar usually brings me to the top of the feed - that said, having the FAB pop up when the nav bar is hidden makes perfect sense to me
- In comment view I'm still hoping we'll eventually have up/down arrows that will allow us to quickly navigate between reply threads; I don't know if other apps do this, but in Boost for Reddit long-pressing the up arrow in this view would scroll you back to the top of the page

I still think this is great, and the other functionality in this PR is a game changer. Just weighing in in case it provides any food for thought :)

bactaholic:
> This is really cool! I will say, it feels like the FAB is doing some work that (in some cases) could be done by existing buttons (or buttons planned for the future):
> 
>     * In other apps, hitting the home button on the nav bar usually brings me to the top of the feed - that said, having the FAB pop up when the nav bar is hidden makes perfect sense to me
> 
>     * In comment view I'm still hoping we'll eventually have up/down arrows that will allow us to quickly navigate between reply threads (see issue [Up/Down arrows for navigating comment threads #75](https://github.com/hjiangsu/thunder/issues/75)); I don't know if other apps do this, but in Boost for Reddit long-pressing the up arrow in this view would scroll you back to the top of the page
> 
> 
> I still think this is great, and the other functionality in this PR is a game changer. Just weighing in in case it provides any food for thought :)

I was thinking about this first point during working on it today, I think the one on the main feed could probably be replaced with a tap on the home icon of the Nav bar, will look into that as a replacement. The problem is as its currently built, the return-to-top button is built straight into the post_card_list.dart file so it appears anywhere that class is called which includes the main feed.
I looked into separating the button per page but was having issues getting the scrollController to react, will look into it more as it seems that's probably the best way forward.

As far as the long press on the upvote, I am a boost user of like...6 years and I had no idea that was a feature haha. I can look into it for sure, if possible it probably won't be in the main release for a while.

machinaeZER0:
> As far as the long press on the upvote, I am a boost user of like...6 years and I had no idea that was a feature haha. I can look into it for sure, if possible it probably won't be in the main release for a while.

I have a screenshot of this! Just to be clear, it was the up arrow on this little nav popup:
Screenshot 2023-07-09 083301

And yeah, a long-press brought you right back to the top of the page :) very handy little overlay!

machinaeZER0:
Also to nod back to Home button functionality (if anything in here there feels relevant), linking issue #149

bactaholic:
@hjiangsu there was a merge conflict, but the code in develop wouldn't compile as it was missing a return statement for the builder in line 164. Added back the return to show page view, good to go

hjiangsu:
Just double checking - this is good to go? I'll merge it in if it is!

bactaholic:
> Just double checking - this is good to go? I'll merge it in if it is!

Yep, all good to go.

PR #254, opened 2023-07-06
Shorten code for edge-to-edge images and add link padding

Fmstrat:
For some reason I had not considered a variable for this before, so it just shortens the code a bit. Also, edge-to-edge was getting applied to the links pre-preview which looked good at the time, but the more I look at it the more it looks weird, so I solved for that.


hjiangsu:
Looks good! I'll merge it in

Issue #74, opened 2023-06-25
Hide menus on downward scroll (/reveal on up-scroll)


machinaeZER0:
Suggestion: an option (perhaps toggleable) to hide the top bar (with the hamburger menu) and bottom bar (with the home button) while scrolling through your feeds. Starting to scroll back up the feed would cause the menus to slide back into view/become accessible again.

I see a fair number of lemmy/reddit apps with this feature, and it feels like a useful thing to implement - the main benefit is that you get a not-insubstantial amount of screen real estate back as you're scrolling, which always feels like a win to me :)

CTalvio:
Example of the feature in Relay for Reddit.
https://github.com/hjiangsu/thunder/assets/4365015/d65f7536-3d2b-4cae-870e-50c3f35d93ce

everdred:
I understand it may be popular but I'd greatly appreciate it if this could be implemented as a toggle-able option, as one of the most refreshing things I noticed about Thunder is that it _doesn't_ have this behavior. I'm really not a fan of having basic UI elements I might want to glance at or use disappear, and there's just something fundamentally weird to me about how the unconnected act of scrolling up is how you bring back the title, filter and hamburger buttons. Maybe we can both be happy!

Fmstrat:
For more context, and I think the OP means this, both the top and bottom bars should hide on swipe up and restore on swipe down/reload. When returning to the feed from reading a post, the bars should remain in the state they were when the item was tapped on. I also agree with making this a toggle option (default true).

hjiangsu:
Yeah, I'll chime in on this. I think it's best suited to be a toggle!

hjiangsu:
This is partially added in with this PR: #1029 for the top app bar. For the bottom navigation bar, there is a lot more work that would be required to get it to work well so that is on-hold for the future.

machinaeZER0:
Heck yes!!!

vinoff:
Oh! This is the one feature I am waiting for! Please do a release!!

PR #257, opened 2023-07-06
DRAFT: Hide menus on scroll

Fmstrat:
@hjiangsu This is a very first-draft of menu hiding on scroll. There are some commented items that I'm not sure of the full impact: 
- https://github.com/hjiangsu/thunder/compare/develop...Fmstrat:thunder:feature/74/menu-hide-on-scroll?expand=1#diff-f1fe73cc104d20a21495aeae446221985476f457f51ed3ad51f3d8336dc0be12R69
- https://github.com/hjiangsu/thunder/compare/develop...Fmstrat:thunder:feature/74/menu-hide-on-scroll?expand=1#diff-7d653dd249e9c08128aa24d99a847ba6fe3b8273d6201b4a179253ab1d2a6e90R88

These were required to make the hiding work, but I'm unfamiliar with Dart and just doing this in spare time ;)

I also can't test horizontal swipes in the Android Emulator for some reason. Lastly, I didn't focus on the bottom bar yet as I'd like to know if it all works for you with horizontal swipes, etc.

Addresses: https://github.com/hjiangsu/thunder/issues/74

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -66,48 +66,60 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
         },
         builder: (context, state) {
           return Scaffold(
-            key: widget.scaffoldKey,
-            appBar: AppBar(
-              title: Text(getCommunityName(state)),
-              centerTitle: false,
-              toolbarHeight: 70.0,
-              actions: [
-                Row(
-                  mainAxisAlignment: MainAxisAlignment.center,
-                  children: [
-                    if ((state.communityId != null || state.communityName != null) && isUserLoggedIn)
-                      IconButton(
-                        icon: Icon(
-                          (state.subscribedType == SubscribedType.notSubscribed || state.subscribedType == null) ? Icons.library_add_check_outlined : Icons.library_add_check_rounded,
-                          semanticLabel: (state.subscribedType == SubscribedType.notSubscribed || state.subscribedType == null) ? 'Subscribe' : 'Unsubscribe',
-                        ),
-                        onPressed: () {
-                          HapticFeedback.mediumImpact();
-                          context.read<CommunityBloc>().add(
-                                ChangeCommunitySubsciptionStatusEvent(
-                                  communityId: state.communityId!,
-                                  follow: (state.subscribedType == null) ? true : (state.subscribedType == SubscribedType.notSubscribed ? true : false),
-                                ),
-                              );
-                        },
-                      ),
-                    IconButton(
-                        icon: const Icon(Icons.refresh_rounded, semanticLabel: 'Refresh'),
-                        onPressed: () {
-                          HapticFeedback.mediumImpact();
-                          return context.read<CommunityBloc>().add(GetCommunityPostsEvent(reset: true, sortType: sortType, communityId: state.communityId));
-                        }),
-                    IconButton(
-                        icon: Icon(sortTypeIcon, semanticLabel: 'Sort By'),
-                        tooltip: sortTypeLabel,
-                        onPressed: () {
-                          HapticFeedback.mediumImpact();
-                          showSortBottomSheet(context, state);
-                        }),
-                    const SizedBox(width: 8.0),
-                  ],
-                )
-              ],
+            // key: widget.scaffoldKey,
+            body: NestedScrollView(
+              floatHeaderSlivers: true,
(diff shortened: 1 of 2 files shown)


hjiangsu:
I just took a quick look at the code, haven't ran it myself yet, but it seems like it's okay for now. Just a couple of points:

- I'm a bit wary of removing the `scrollController` from the `ListView` because that is what controls when more posts are fetched based on the scroll distance
- I would have to do some profiling on this change, to see if its still cleaning up the ListView as expected. I'm not too familiar myself with how the nested scroll view works, but from what I've experienced with `SingleChildScrollView` and `ListView.builder`, there can be cases where the ListView fails to clean up after itself and causes memory issues (which then causes lag and performance issues)

Fmstrat:
The commit I just added prints out `endScroll` when it gets to the end. As I only had a few minutes I didn't take the time to figure out how to get the `GetCommunityPostsEvent` to call from there, though.

Fmstrat:
I also think we could consider switching away from the `ListView` altogether and using a `MasonryGridView` (which is a Sliver style widget for good performance and includes the same `builder` feature as `ListView` for caching since it seems to be a `GridView`. This way we can use a `crossAxisCount` of 1 on phones, and 2 on tablets. I'm testing it out now and it seems to work quite well. (I'll make another PR when it's ready to show off)

Fmstrat:
As example, converting from `ListView` to `MasonryGridView` and setting the columns to 2 immediately worked with dynamic hights:
Screenshot_1688680782

Fmstrat:
Honestly this is pretty amazing, it's performing fantastically and was only a few lines of code change. I will make a new PR now, and revise this one once you take a look at that (as this should still apply).

hjiangsu:
Sorry, I've been really busy with work and other stuff today so I couldn't reply earlier but wow, that looks really good for a test!

I'll reply with some more details when I can but first glance, I think switching to MasonryGridView seems to be the way to go, especially since it supports columns natively

I can do some profiling when I have the chance to double check for memory issues or performance regressions but good work!

Fmstrat:
No need to apologize, you should see how slow I am on my own repos 😉

hjiangsu:
@Fmstrat Do you still consider this to be a draft, or should I take another look at the code?

lib/community/pages/community_page.dart
```diff
@@ -28,6 +29,22 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
...
+
+  @override
+  void initState() {
+    super.initState();
+    WidgetsBinding.instance!.addPostFrameCallback((timeStamp) {
+      globalKey.currentState!.innerController.addListener(() {
+        if (globalKey.currentState!.innerController.position.pixels >= globalKey.currentState!.innerController.position.maxScrollExtent * 0.7) {
+          print('endScroll');
+          // How do we call this from here?
+          // context.read<CommunityBloc>().add(GetCommunityPostsEvent(communityId: widget.communityId));
```
hjiangsu, line 42: Unfortunately, we won't be able to call this function in the initState function since it has no access to the context.

One "hacky" way to pass in the context is to allow CommunityPage to accept a BuildContext as one of its parameters and then use that context here. However, I'm not sure if that's the best way to go about it

hjiangsu (commented):
Just added a quick review of the code!

Fmstrat:
I will take a look when I can, but yes, this is definitely still draft. It will need to be merged into the new 2-column code, and of course the review items above are still outstanding issues. I also think we need to find a way to hide the bottom bar as well as the top bar.

hjiangsu:
Sounds good! Let me know if you want me to help with some of this

Fmstrat:
> Sounds good! Let me know if you want me to help with some of this

If you feel like it, go for it! I'm pretty busy right now and will be traveling next week.

tom-james-watson:
Personally I dislike menus auto-hiding on scroll, though I feel like I'm in the minority. It would be good if this behaviour could be optional.

everdred:
> It would be good if this behaviour could be optional.

There's two of us! The developer had [previously suggested](https://github.com/thunder-app/thunder/issues/74#issuecomment-1621805328) this could be implemented as a toggle.

hjiangsu:
I'll close this PR as it's changes are no longer relevant because of the refactor of the feed page. With the refactor, it should be much easier now to implement a toggle which automatically hides the top bar on scroll!

PR #261, opened 2023-07-06
Support for marking posts as read after viewing media

ajsosa:
Enhancement: https://github.com/hjiangsu/thunder/issues/205

#### Fix Details
Added the option to mark posts as read after only viewing media. I also made this work when viewing the list of posts for a particular user. The previous behavior of marking posts as read only after opening the full post was not working in the user posts view. These changes fix that as well.

Default behavior is still the same as before, you'll have to turn on the option in general settings.

#### Known Issues
Marking posts as read from a user posts view won't update the previous existing community post view. Decided to leave that alone for now. It can always be addressed later if people want that behavior.

#### Implementation Details & Concerns
I added a new `enum` to be able to tell what viewing context a post is in since it can either be in a community post view or user post view and to get this to work I had to mark as read functionality in the appropriate bloc. The thought was if there are future views where posts may be displayed, this could be added to. 

I'm not 100% sold on my implementation, I feel like there is probably a much better way of doing this so please let me know if you know of a better/correct approach.

### DIFF
--- a/lib/settings/pages/general_settings_page.dart
+++ b/lib/settings/pages/general_settings_page.dart
@@ -134,8 +134,8 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> {
         await prefs.setString('setting_post_default_comment_sort_type', value);
         setState(() => defaultCommentSortType = CommentSortType.values.byName(value ?? DEFAULT_COMMENT_SORT_TYPE.name));
         break;
-      case 'setting_general_mark_post_read_on_meda_view':
-        await prefs.setBool('setting_general_mark_post_read_on_meda_view', value);
+      case 'setting_general_mark_post_read_on_media_view':
+        await prefs.setBool('setting_general_mark_post_read_on_media_view', value);
         setState(() => markPostReadOnMediaView = value);
         break;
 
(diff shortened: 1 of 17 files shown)


hjiangsu:
Thanks! I'll take a deeper look at this when I have the chance (I've been pretty busy today so didn't get to spend as much time as I'd like)

ajsosa:
Just realized I completely botched my original commit message after fixing merge conflicts. Should be fixed now 😅

hjiangsu:
Just a quick update on this so that you're also updated - I took a look at the code already, but I'm just trying to see if there's maybe a better way to do this.

There _might_ be a way to tell which context a post might be by doing a try-catch on fetching a given bloc, but I'm not sure if thats the best way to go about it either.

e.g.,
```dart
try {
  // try to get a given context
  UserBloc = context.read<UserBloc>();
  // ... rest of the logic to mark post as read
} catch (e) {
  // User bloc is not present, attempt to get community bloc
  CommunityBloc = context.read<CommunityBloc>();
   // ... rest of the logic to mark post as read
}
```

Either way, I think it still looks okay, but maybe I'll take another day to think of any other alternatives before merging this in!

ajsosa:
Yep looks like this works.

```dart
try {
  UserBloc userBloc = BlocProvider.of<UserBloc>(context);
  userBloc.add(MarkUserPostAsReadEvent(postId: postId, read: true));
} catch(e){
  CommunityBloc communityBloc = BlocProvider.of<CommunityBloc>(context);
  communityBloc.add(MarkPostAsReadEvent(postId: postId, read: true));
}
```

This would get rid of the new `enum` I added and would eliminate passing that around. Would you prefer I make this change?

hjiangsu:
Yeah, I think this might be the best approach - it'll reduce the overall amount of changes required to many of the files, and reduces the amount of passing around required as well!

Sorry for the changes, I hope its not too much of a hassle  😅

ajsosa:
Should be good to go!

lib/settings/pages/general_settings_page.dart
```diff
@@ -133,6 +134,10 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> {
         await prefs.setString('setting_post_default_comment_sort_type', value);
         setState(() => defaultCommentSortType = CommentSortType.values.byName(value ?? DEFAULT_COMMENT_SORT_TYPE.name));
         break;
+      case 'setting_general_mark_post_read_on_meda_view':
```
hjiangsu, line 137: Theres a small typo - `meda` instead of `media`!

hjiangsu (commented):
Just one small typo - otherwise I think it looks good!

ajsosa:
Fixed it.

PR #263, opened 2023-07-06
Improve vote indicator apparance in post feed

CTalvio:
image

This change makes the up/down vote color affect the text of the indicator in post view, as well as the arrow.
The arrow now also flips for posts you've downvoted.

FYI, I removed some code on line 31 in `lib/shared/icon_text.dart` in order to make my changes, I am unsure what it did, but nothing seems to have broken?


hjiangsu:
This looks good! Thanks :D

Issue #258, opened 2023-07-06
Display banners


CTalvio:
Both users and communities can have banners. I propose something like this:

![Screenshot_20230706_032048](https://github.com/hjiangsu/thunder/assets/4365015/02afa004-3446-45c0-9806-d4cefeb7111a)

This is already in-app, though I have not found a good way to fallback to no banner for users and communities that do not have one. It seems BoxDecoration is unable to fall back to a color the way CircleAvatar can.

@hjiangsu You can assign this one to me.

hjiangsu:
I like the look of it, it keeps it clean. I would say that if a community has no banner, then default it to what it is currently (match the background colour)

If you provide the code for the BoxDecoration, I can suggest ways to fallback to no banner if you'd like

CTalvio:
Fallback to solid color is the plan.

As far as I could tell, BoxDecoration with an image can only fallback to another image, by using a conditional similar to the one currently in CircleAvatar.

The simplest solution would be to just give it a blank placeholder image.

But the perfectionist in me wonders if there is a better solution. I tried to look up ways to conditionally disable the whole BoxDecoration, but didn't find anything.

I can get you the code later.

hjiangsu:
I can give you a quick alternative - you can probably set a conditional on checking if the image exists first and then setting the box decoration

It would be something like this:

```dart
return Container(
  decoration: communityInfo?.communityView.community.banner != null ? BoxDecoration(...) : null,
  child: Container(...)
```

I _think_ that `decoration` can be set to null to trigger the default behaviour but maybe double check that!

CTalvio:
@hjiangsu That works perfectly. Exactly what I wanted to do. PR incoming after some fiddling.

CTalvio:
@hjiangsu you can mark this ready in the next release?

hjiangsu:
Done!

machinaeZER0:
I think this looks super clean! My only thought (and I know this was coded using the current layout as a base, with the icon next to the title/community info) is that this layout hides a lot of the community's banner, which seems less than ideal depending on what's being displayed.

I unfortunately don't have a screenshot of it handy, but I really liked the way Boost did this on reddit - I believe they had the banner mostly visible, with the community icon centered and slightly offset below? I've tried to create a rough mockup of what I mean:

![icon + banner mockup](https://github.com/hjiangsu/thunder/assets/3624023/7ff30e9f-e598-40cc-9fbb-dd3b0a5568c4)

[...]

PR #264, opened 2023-07-07
Community and User banners

CTalvio:
Users:

image

Communities:

image

Both fall back to solid background color if no banner is present. Both also follow light/dark/black theme in the gradient.

image


CTalvio:
#258

PR #267, opened 2023-07-07
Fix uppercase image dimensions

Fmstrat:
This fixes a bug where the URI was not parsed to lowercase before checking extensions to parse image dimensions.

Also, the static offset of 24 that allows for the curves on image corners was causing issues with cropping on edge-to-edge images, so I fixed that. For optimization reasons I did not have it clear the cache if someone changes the setting, so if they swap back/forth previously loaded images will continue to have the crop (or no curved corners if they swap back) until an image reloads.


hjiangsu:
Looks good to me, thanks!

Issue #204, opened 2023-07-03
Remember the collapsed state of comments when collapsing/expanding parents


micahmo:
**Is your feature request related to a problem? Please describe.**
The collapsed state of comments is not remembered.

**Describe the solution you'd like**
The collapsed state of comment should be remembered within a thread. For example...
Collapse a second-level comment. Collapse a top-level comment. Expand the top-level comment. The second-level comment is now expanded again.

Another possible enhancement would be to remember even when leaving the comment feed and coming back, but that may be a separate issue and/or undesirable.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

hjiangsu:
> Another possible enhancement would be to remember even when leaving the comment feed and coming back, but that may be a separate issue and/or undesirable.

Did we still want to implement this feature here? I kept this open previously because we don't currently have this! Although I'm not sure if this should be the default behaviour or not

micahmo:
Ahh, the title of this issue, "Remember the collapsed state of comments when collapsing/expanding parents" is implemented, but as I re-read my request, the second part, "remember even when leaving the comment feed and coming back" is not. Should I open a separate issue for that?

hjiangsu:
Yeah, we can open up a new issue for this!

micahmo:
Done in #742.

Issue #190, opened 2023-07-02
Make compressed comments visible with a count


Fmstrat:
**Is your feature request related to a problem? Please describe.**
When compressing comments and having the parent comment still displayed, you can't tell the comment is compressed.

**Describe the solution you'd like**
When a comment is compressed but still displayed, put a numerical indicator for how many comments were compressed underneath.

hjiangsu:
This should be possible to do fairly easily - do you think its better to check for the length of the direct replies, or the total number of replies (including replies to replies)

Fmstrat:
Personally I'd do total vs direct, but I have no real reason for that beyond showing volume. It would probably be useful on the fully compressed view, too.

hjiangsu:
Closing this issue since it has been released in v0.2.1+11!

PR #268, opened 2023-07-07
Show collapsed comment count

micahmo:
This PR introduces the feature suggested in #190. It accomplishes two goals: (1) to indicate when parents whose contents are hidden have collapsed children, and (2) to show how many child comments there are.

### Notes

- The UI is heavily inspired by Relay.
- I chose to show the count of top-level children only because (a) there was easy access to this value in code without a recursive count and (b) it matches the behavior of Relay.
- Small downside: the width of the publication time may vary, causing the count to have different y-axis placement on different comments. However, this is the same behavior as the star, so maybe not a big deal.
- When we allow the parent comment to be hidden, I show the number (even if it's 0). I'm not sure whether we need this, since there is a clear UI indication that the parent comment is collapsed, but it still may be nice to know that there are no children. In the mode where the parent does not collapsed, I don't show anything on press because nothing else changes.

#### Light mode / do not hide parent commend on collapse (video)

https://github.com/hjiangsu/thunder/assets/7417301/2250427c-b21c-4694-8969-0c0be8ddeb5d

#### Dark mode (screenshot)

| image |
| - |

#### Dark mode / hide parent comment on collapse (screenshot)

| image |
| - |

---

P.S. There is still a small "issue" where the collapsed state of comments is not remembered. This is not new with my change, and has been logged in #204.

### DIFF
--- a/lib/post/widgets/comment_card.dart
+++ b/lib/post/widgets/comment_card.dart
@@ -233,7 +233,7 @@ class _CommentCardState extends State<CommentCard> with SingleTickerProviderStat
                       crossAxisAlignment: CrossAxisAlignment.start,
                       mainAxisAlignment: MainAxisAlignment.start,
                       children: [
-                        CommentHeader(commentViewTree: widget.commentViewTree, isOwnComment: isOwnComment),
+                        CommentHeader(commentViewTree: widget.commentViewTree, isOwnComment: isOwnComment, isHidden: isHidden),
                         AnimatedSwitcher(
                           duration: const Duration(milliseconds: 130),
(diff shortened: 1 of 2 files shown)


CTalvio:
What determines the color of the badge? Theme primary?

micahmo:
@CTalvio

It's the same color as the FAB.

https://github.com/hjiangsu/thunder/blob/1bd6a931e9582d613b9e0ef374323d507557c314/lib/post/widgets/comment_header.dart#L70

hjiangsu:
Just took a look and everything looks good! Nice job!

Issue #280, opened 2023-07-08
Add additonal profile info to user header


CTalvio:
As we will be using display names by default, some users may have a different handle. Therefore I'd like to display the full handle of a user below their display name, on their profile.

I also co-opted the icons commented out into post and comment counts.

![image](https://github.com/hjiangsu/thunder/assets/4365015/976b040c-55a0-4c1c-b14a-0e7e3459ec03)

I'll roll this into the same PR as for display names, but thought I'd give anyone who wants to a chance to comment.

CTalvio:
Same goes for communities. The display names are not always the same as the handles, and as such:

![image](https://github.com/hjiangsu/thunder/assets/4365015/79d0b993-0598-4098-8f70-cc14d180de5b)

Poolitzer:
Great change!

hjiangsu:
I think those are good changes overall - I would maybe like to see if there are better icons for the post/comment points (unless those are actually the number of posts/comments that you've made) as they might be slightly confusing

CTalvio:
That is the **actual** number of posts and comments I've made. 😅

hjiangsu:
In that case, I think that works!

CTalvio:
I have also changed the online count, which doesn't seem to work anymore, with monthly active.

Issue #272, opened 2023-07-07
New comments indicator


CTalvio:
Looking into adding some kind of indicator for when a post you've already read has new comments.

The Lemmy web UI tells you when a post you've already read has new comments, and also highlights new comments in the comment view.

I'd like to implement both of these. @hjiangsu feel free to assign this one to me as well.

En example of what this could look like:

![image](https://github.com/hjiangsu/thunder/assets/4365015/acf532d4-f39d-4f85-aa76-7ce4f2b07325)

CTalvio:
I have settled on something like this:

![image](https://github.com/hjiangsu/thunder/assets/4365015/919b19bd-55a6-4bb6-945b-b1cc92b43de7)

hjiangsu:
I like the new look, it makes it more minimalistic. I wonder if we should still display the number of new comments? e.g., it could be in brackets 112 (4) would indicate 4 new comments

CTalvio:
I changed it so the number is replaced by a +number to indicate the amount of new comments. The total is not shown, but can be viewed when opening the post.

This is in the screenshot in the PR.

I figure if there are additional comments, and you are seeing the post, having the total number visible at a glance is not necessary to show that it is a post with a lot of activity.

Issue #92, opened 2023-06-26
Support User Display Names


CTalvio:
**Is your feature request related to a problem? Please describe.**
Currently, users are identified using their handles, not their self-chosen display names.

**Describe the solution you'd like**
Respect user display names.

In the future, user avatars and home instances would also be nice.

hjiangsu:
Do you think this is better as a toggle in the settings to switch between them, or should the current user handles be directly replaced with the display names?

wildebeest3:
> Do you think this is better as a toggle in the settings to switch between them, or should the current user handles be directly replaced with the display names?

I think it would be better as a toggle

CTalvio:
A toggle. Display names by default.

CTalvio:
I am looking into this.

Assign to me @hjiangsu ?

hjiangsu:
Sounds good! Thanks!

PR #281, opened 2023-07-08
ABLE TO MERGE: Display names and more

CTalvio:
I ended up including a lot of things here.

- #92 
    - There is now a setting for using display names for users that have them, which is on by default 
- #272
    - Read posts that have new comments, show a theme color highlighted unread indicator instead of the comment count 
    - In post view, numbers of both new and total comments are shown
    - Comments made in the last 15 minutes are highlighted, similar to lemmy webUI (maybe consider highlighting just timestamp text?)
- #280
    - Profile and Community pages show full handle below display name/title 
    - User profiles now display post and comment stats 
- Edited comments have a pencil icon next to their timestamps
- The clock icon on edited posts changes to a refresh icon (maybe pen instead as well?)
- Tapping usernames in comments now opens their profile

There are a couple issues that will need to be resolved:
Thunder/lemmy_api_client is not marking comments read on a post, hence the unread comments indicator only works for posts that have been read in the webUI/Jerboa, and it also does not go away after reading said comments. I tried to look into writing the logic for this, however it seems there may be an undocumented API change involved...

Both Jerboa and the WebUI are able to mark comments read on a post when opening it. Somehow.

image
Screenshot_20230708_175550
Screenshot_20230708_152406


hjiangsu:
> Thunder/lemmy_api_client is not marking comments read on a post, hence the unread comments indicator only works for posts that have been read in the webUI/Jerboa, and it also does not go away after reading said comments. I tried to look into writing the logic for this, however it seems there may be an undocumented API change involved...

I know theres a specific API call that must be done to set a comment as read. This has not yet been implemented in yet into Thunder. I would need to see what the general consensus is for marking a comment as read:
- The easiest way would be to mark any comments that have been loaded as read
  - Thunder loads 50 comments at a time, so when you first open up a post, it'll mark the first 50 as read.
  - As you scroll down, it'll load 50 more comments, and at which point it can mark those as read, etc..
- Marking comments as read as you scroll pass them. This would require some more logic to check if the comment is currently within view of the user's screen, and then mark those as read. This would be more accurate, but would be more difficult to implement

I'm not too sure what the behaviour is for the lemmy web-ui, or Jerboa, so maybe some feedback on this would be good!

CTalvio:
Both Jerboa and the webUI simply mark all comments read upon opening a post.

hjiangsu:
Ahh okay - I think this is the relevant API function https://join-lemmy.org/api/classes/LemmyHttp.html#markCommentReplyAsRead

I have to double check to see if it is, and I can add it in if needed

CTalvio:
Note the name markCommen**Reply**AsRead

I think this is the one used for marking inbox comment replies as read. Thunder is already using it via lemmy_api_client from what I could tell.

CTalvio:
image

CTalvio:
> Unless it somehow gets used in both cases.

You know what. This might be exactly how it works.

I'll get back after doing some testing.

hjiangsu:
Hmm, I wonder if marking a post as read causes it to also mark all the comments as read? I can look into this more after I finalize the upcoming release, and I'll see if I can get back to you on this.

These are my initial thoughts (not sure if this is actually how it works):
- Marking a post as read marks all the corresponding comments as read
- Posts have a count for how many unread comments there are, so this unread comment count is tied to the post logic (at least it seems like from first glance)
- Maybe what we have to do then, if this is the case, is to mark the post as read every time a user opens up a post, which then refreshes the unread comments count

Placing this for reference: 
- https://join-lemmy.org/api/interfaces/PostView.html#unread_comments
- https://join-lemmy.org/api/interfaces/PostView.html#unread_comments

CTalvio:
This was my assumption as well. On a new post, the unread count of a post is the same as the number of comments, before the first time a user opens it. I'm unsure how this is able to handle new comments on a post, if you open it for the first time when it has zero comments.

After I open a post in the webUI the unread count is set to zero, and goes up for each new comment. I added a check for a mismatch in this number, and if they are the same, to not display the unread counter on new posts. Otherwise ALL new posts were be showing "new" comments.

However, opening an unread post in Thunder, does NOT set the count to zero, causing the unread and total comment count to continue increasing in lockstep. Thereby failing to show the number of new comments since last having opened the post.

...

A thought occurs to me. In the webUI, and Jerboa, the unread count only resets to zero when LEAVING a post. While viewing it, the unread count remains. Does Thunder only set a post to read, the one first time?

If setting a post as read, simply sets it as read, would sending another read event reset the count?

hjiangsu:
> Does Thunder only set a post to read, the one first time?

Thunder sets a post as read any time the post is opened from the feed, so that shouldn't be the case. However, it does not mark the post as read again when the user leaves the post page. Maybe we have to mark the post as read again when the user leaves the page?

https://github.com/hjiangsu/thunder/blob/40c7dd9dfd5af8eb854e931b8c011d37c6b1782b/lib/community/widgets/post_card.dart#L180-L188

If you try to enter a post, leave the post, and enter the post again, does the unread comment count change? (this would technically trigger the read post function twice)
- You can also try to refresh the feed and see if the unread comments change - I can't recall immediately if the post gets updated whenever we mark it as read (which would then keep the same unread comment count)

CTalvio:
I have not been able to get the unread count to change using any interaction within Thunder. I was hoping thunder only marked posts read, once.

CTalvio:
image

Changed new comment highlight to a badge. The reason for the highlight may not be clear to users. The literal word "new" is.

CTalvio:
@hjiangsu I have commented out the new comment indicator features. They are ready to go, provided you find a way to set comments as read on post open. Once someone does, all thats needed is to uncomment what was commented in 44ea5fe2b4135bbbd9b01a3c1adaa526707da797.

All other features are good to go.

hjiangsu:
Looking good - I'll merge it in! Thanks :D

Issue #249, opened 2023-07-06
All comments don't show up in profile tab


subhashay:
**Description**  
So far I have commented on 2 posts using thunder. But under my profile section of the app, I can only see one. The latest one is not visible.

**How to Reproduce**  
NA

**Expected Behavior**  
NA

**Device & App Version:**  
 - Device: Pixel 6
 - OS: Android 13
 - Version: Latest (v0.2.1+10)
**Additional Context**  
User url: https://lemmy.world/u/conversationkiller7

ajsosa:
Looks like `comment.buildCommentViewTree()` is used by both `post_bloc` and `user_bloc`. Problem is that for child comments (replies) the function tries to insure there is a parent to attach to. For the user view we just want a flat list of comments made by that user.

hjiangsu:
That sounds about right. Maybe the easiest way to resolve this is to add in a parameter to determine if we need to build the comment tree or not?

If the parameter is false, we can simply return a flat list of comments. This would also enable us to create a toggle in the future for comments in a post to be flat as well (similar to what the web-ui has)

ajsosa:
As I was messing around with a fix. I encountered this oddity. I'm not sure what's going on. In the first screen shot you can see the latest comment the user made. But when you click to go to the post for context, the comment doesn't exist. However you can see one of his earlier comments is there which is also visible in the first screen shot. Any ideas? As an aside, perhaps the highlighted user name should have more contrast. Kind of easy to miss.

![userProfileComment](https://github.com/hjiangsu/thunder/assets/41923324/539e7331-b463-4051-a76d-539da0718c57)

![postViewUserComment](https://github.com/hjiangsu/thunder/assets/41923324/6648a7cb-951c-4e4e-9102-9e004e0f79aa)

hjiangsu:
That's interesting. Could it be a federation issue maybe? Either that or there's some logic that's causing it to not show all the comments

Try and update your branch to the latest changes in develop and double check?

ajsosa:
It looks like it's a Lemmy specific issue. It's really weird. So me and that user are actually on the same instance. If I go to that post via my instance, I see the same data I saw in Thunder. If I go to the host instance where the post originated. I see the missing comment and a whole bunch of other comments I wasn't seeing before.

[...]

PR #282, opened 2023-07-08
#249 Fixed comments not all showing on User page.

ajsosa:
Bug Fix: https://github.com/hjiangsu/thunder/issues/249

`comment.buildCommentViewTree()` can now optionally return comments flattened so that we get a simple list to show on User page.


hjiangsu:
Looks good! Thanks :D

Issue #251, opened 2023-07-06
Ability to copy comment text (long-press menu)


krestenlaust:
**Describe the solution you'd like**
Long-pressing a comment opens a menu with the option to copy its content.

**Describe alternatives you've considered**
None.

**Additional context**
#189

PR #287, opened 2023-07-09
Copy/share comment on long press 

### LINKED ISSUE #232: Ability to share comment (long-press menu / swipe gesture)
Long pressing a comment should open a menu letting you share a link to it.

A custom swipe gesture that also let's you share it (like staring and upvoting a comment)

**Additional context**
#189

vbh:
Closes #232, #251


hjiangsu:
Looks good! I think one thing we can do in the future is to allow customizable selection of the text - we can do this by using the Markdown's `selectable` attribute: https://pub.dev/documentation/flutter_markdown/latest/flutter_markdown/MarkdownWidget/selectable.html

This way, we can either let the user select the whole comment, or they can use the default system methods to select parts of the text!

Issue #223, opened 2023-07-04
More contrast in text for themes


Fmstrat:
**Is your feature request related to a problem? Please describe.**
With content displayed in the post feed, it's all starting to run together:

![text](https://github.com/hjiangsu/thunder/assets/971474/e755618f-16f1-4b6b-8634-76bb1c6bcdb8)

**Describe the solution you'd like**
Updates to theming to support content adjustments, including expansion of themes to use a drop down with "System", "Dark", "Light" to be the defaults and "OLED" + others to be additional options.

**Describe alternatives you've considered**
Making content displayed in the feed gray instead of white. Realistically, this could work, but a broader theme capability would be nice.

machinaeZER0:
Oh yeah - in compact view the non-title text all looks to be a slightly darker shade, which I agree is helpful for readability when scrolling!

CTalvio:
The color on the community name and instance should probably match between compact and normal.

acidrums4:
I think choosing a bold(er) and a bit bigger font for post titles, and a lighter and dimmer font for metadata (magazine and instance names) could work.

I prepared this sketch with Inkscape to show what I mean. Note that I used the "Manrope" typeface as I did not have Roboto installed on my computer.

![bitmap](https://github.com/hjiangsu/thunder/assets/513421/ac580081-1273-46ea-81ae-f838a420bbd1)

machinaeZER0:
Yeah, this is pretty much exactly how compact mode handles font colors :)

CTalvio:
Should I make this change or is someone else already on it?

machinaeZER0:
This feels like it would be a total lay-up for me to try and tackle, but I've never done a PR before and I'm not sure where to begin :(

CTalvio:
> This feels like it would be a total lay-up for me to try and tackle, but I've never done a PR before and I'm not sure where to begin :( presumably I'd find where the color values are set for normal and compact modes and then make them match?

Yes.

I'll get you started. The files are `lib/community/widgets/post_card_view_comfortable.dart` and `lib/community/widgets/post_card_view_compact.dart`.

machinaeZER0:
Thank you! As mentioned over there, I don't actually know how to set up a build environment on Windows (or at all, honestly) to compile and test in Thunder myself, but looking at those two files side by side I believe I have been able to get compact and "normal" view to match with this PR?

Would be interested in setting up a test environment in the future! Would be cool to be able to help out in small ways such as this :)

[...]

PR #292, opened 2023-07-09
Update post_card_view_comfortable.dart to match compact view styling

machinaeZER0:
Update post_card_view_comfortable.dart to have the font colors of unread posts more closely match compact mode (see issue #223). If it works, then this:
image

Should now match this:
InkedScreenshot_20230704-082754

The goal being to standardize font color across views, and to help the post titles "pop" more as you scroll.

**IMPORTANT: I am not a coder!** Just a fan :) I don't currently know where to begin in terms of compiling code in order to test this change, so apologies if this either doesn't work or has some unintended consequence.

Looking at post_card_view_compact.dart and post_card_view_comfortable.dart side by side, it looked like this was the value governing the color of at least the community and instance names, so I've updated the line in the latter to more closely match the former. Pretty elegant that this color change seems to be based on opacity levels, as it means it gives the intended effect across light and dark modes. Clever!

Please take a look at your convenience and let me know if this even works, lol. Thank you to @CTalvio for showing me where to start digging (and for everyone's help as I learn how GitHub actually works)!

### DIFF
--- a/lib/community/widgets/post_card_view_comfortable.dart
+++ b/lib/community/widgets/post_card_view_comfortable.dart
@@ -124,7 +124,7 @@ class PostCardViewComfortable extends StatelessWidget {
                           textScaleFactor: state.contentFontSizeScale.textScaleFactor,
                           style: theme.textTheme.titleSmall?.copyWith(
                             fontSize: theme.textTheme.titleSmall!.fontSize! * 1.05,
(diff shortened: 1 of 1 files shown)


CTalvio:
Yup, you did it right :D

machinaeZER0:
YesssSSSssSs. Thanks for your help!

hjiangsu:
Looking good! Nice job!

PR #305, opened 2023-07-09
Create post improvement

MrAntonS:
Upload image works like it should, I use one "Upload image button" for both uploading image to the URL and to the text body(It switches automatically). I did so because the markdown widget does not support custom actions, or i couldn't find any information on how to do so. Also, currently markdown preview doesn't work, this is not my bug, since it doesn't work on the latest release as well. Please review this code and make any necessary changes 😄

### DIFF
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -387,6 +387,7 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
         communityId: state.communityId!,
         name: event.name,
         body: event.body,
+        url: event.url,
       ));
 
       // Parse the posts, and append them to the existing list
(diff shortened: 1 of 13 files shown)


MrAntonS:
If anyone in here codes in VScode, please let me know how to fully disable auto formatting with flutter extension

MrAntonS:
Here how it looks depending on if error was caught or nsfw is selected, how image is included in the body. 

photo_4_2023-07-09_22-25-57
photo_3_2023-07-09_22-25-57
photo_2_2023-07-09_22-25-57
photo_1_2023-07-09_22-25-57

MrAntonS:
I tried other media types, but if i try to attach mp4 it's unsupported, gifs are support but it converts them to MP4 and now client cannot support it XD, so attaching other media types might be a problem, however if we implement media player it should be easy.

hjiangsu:
I have an idea for how to implement a media player, but the package used only supports iOS and Android, which might limit compatibility: https://pub.dev/packages/video_player

If we can find a more universal video player, then I might be able to tackle that

MrAntonS:
For now I think it's OK, we can add a checker and if the type is not image or gif then display link instead of preview

PR #307, opened 2023-07-10
Fix exit confirmation theming

micahmo:
This is just a tiny fix on top of #245. In light mode, the snackbar is hard to read.

### Before

| image |
| - |

By changing the background color, the foreground (hard-coded as white) looks fine on both modes.

### After

| image |
| - |

And dark still looks good.

| image |
| - |


bactaholic:
Nice, this looks great. Does it work with Material You when that option in enabled?

micahmo:
> Does it work with Material You when that option in enabled?

Yes, although the colors look funky in light mode. It probably hasn't been given a lot of love. 😆 

| image |
| - |

| image |
| - |

hjiangsu:
Looks good!

PR #308, opened 2023-07-10
Update db version to fix SQL error in the latest release

vbh:
When I updated the app to  latest version I'm getting SQL exception as `userId` column is not getting inserted. 
There was release where db version was 2 and there was no `onUpgrade` function written, so currentVersion is 2 and on the latest release `onupgrade` is included but the version is not updated so `userId` column is not added. Changed version to 3 so `onUpgrade` would be called.


hjiangsu:
Thanks for opening up a PR for this!

Just one thing - I would change `(oldVersion < newVersion)` to `(oldVersion < 3)`. This would allow us to continue to do upgrades to the database in the future and also limits this change to any version below 3 (if I recall correctly, the affected versions that are encountering this bug would be on version 2)

hjiangsu:
Looks good! I'll rebase it to the develop branch and merge it in

PR #310, opened 2023-07-10
Clean up subscription actions

micahmo:
This PR adds the following:

1. Unify the subscription icons between the search and community pages (and show all three possible states on the community page).
2. Show a tooltip describing the subscription action/state.
3. Allow unsubscribing while pending from the search page.
4. Some subscription requests return pending but go through very quickly. To handle this case, we refresh the subscription status briefly after subscribing. (This seems most common when subscribing cross-instance.)
5. Fix wording of add/remove message.

### Notes
 - For 1: I am totally open to feedback on the icons as long as they can be shared between both views and can indicate all three states.
 - For 2: I debated whether the tooltip should show the status (current subscription) or the action (what will happen when pressed). I went with the latter except in the case of the pending which shows both. Again I'm flexible here.
 - For 4: This is a hacky workaround and maybe we're just accounting for a Lemmy backend bug, in which case I'd be fine with removing this.

---

#### Demo 1 -- Search page subscribe (delayed) and unsubscribe

https://github.com/hjiangsu/thunder/assets/7417301/7cf1c482-113f-4deb-b40d-3dfea29218a0

#### Demo 2 -- Community page subscribe (delayed) and unsunscribe

https://github.com/hjiangsu/thunder/assets/7417301/b3bb789b-5fa4-4b17-acd0-2c5d1149ef2e

#### Demo 3 -- Search page subscribe (instant) and unsubscribe

https://github.com/hjiangsu/thunder/assets/7417301/69c05aa0-db4a-430e-89f2-1c941764857a

#### Demo 4 -- Community page subscribe (instant) and unsubscribe

https://github.com/hjiangsu/thunder/assets/7417301/c6010953-5ad8-4a04-8b61-1f919e8c788b

#### Demo 5 -- Community page subscribe (pending) and unsubscribe

https://github.com/hjiangsu/thunder/assets/7417301/e5180836-4ff0-4aaf-b136-41aed18b5275

#### Demo 6 -- Community page subscribe (pending) and unsubscribe

https://github.com/hjiangsu/thunder/assets/7417301/2c441aac-066c-4510-b05c-884683e6ab69

### DIFF
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -336,12 +336,29 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
         follow: event.follow,
       ));
 
-      return emit(state.copyWith(
+      emit(state.copyWith(
         status: CommunityStatus.success,
         communityId: state.communityId,
         listingType: state.listingType,
         communityName: state.communityName,
         subscribedType: communityView.subscribed,
+      ));
+
+      await Future.delayed(const Duration(seconds: 1));
+
+      // Update the community details again in case the subscribed type changed
+      final fullCommunityView = await lemmy.run(GetCommunity(
+        auth: account.jwt,
+        id: event.communityId,
+        name: state.communityName,
+      ));
+
+      return emit(state.copyWith(
+        status: CommunityStatus.success,
+        communityId: state.communityId,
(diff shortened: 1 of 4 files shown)


hjiangsu:
Looking good! I think the icons are probably more representative of the action which should reduce confusion on how to subscribe/unsubscribe to a given community

PR #311, opened 2023-07-10
Fix webp previews

Fmstrat:
Fix for https://github.com/hjiangsu/thunder/issues/276


Fmstrat:
I will say, while it was interesting to figure out the binary placement of data in a webp file, there's got to be an image library in Dart that does this more cleanly across a variety of file types.

hjiangsu:
I wasn't able to find any packages that did this. Usually, the API would return back the width/height of the image, so this whole step could be skipped

hjiangsu:
This is a related PR regarding width/heigh metadata on media: https://github.com/LemmyNet/lemmy/issues/3328

PR #315, opened 2023-07-10
Show instance icons

micahmo:
This PR adds the following items related to instance icons.

- Decreased size of icon above login fields.
- Show instance icon after entering instance URL.
- Add more suggestions to instance hint (so not everyone uses lemmy.ml).
- Moved instance field to top of login page (this may be controversial, but it makes sense to me; username/password is meaningless without the instance, plus you see the icon sooner).
- Show instance icons in profile switcher (we may prefer profile icons, not sure).

#### Notes
- We lose the yellow person indicator for the active account, but we already have an active text field to indicate that.
- I didn't put a solid color behind the instance icons because they generally include transparency and I'm assuming they're generally designed for white/black backgrounds.
- I'm not sure if `AccountExtended` is a good pattern and/or if it should go in a separate file.

#### Account Switcher

image

#### Add Account

https://github.com/hjiangsu/thunder/assets/7417301/f798ccb0-8279-47d0-9406-702cb98cd7b0

### DIFF
--- a/lib/account/pages/login_page.dart
+++ b/lib/account/pages/login_page.dart
@@ -107,9 +107,14 @@ class _LoginPageState extends State<LoginPage> {
             mainAxisAlignment: MainAxisAlignment.start,
             crossAxisAlignment: CrossAxisAlignment.center,
             children: [
-              Container(
-                child: instanceIcon == null 
-                  ? Image.asset('assets/logo.png', width: 80.0, height: 80.0)
+              AnimatedCrossFade(
+                duration: const Duration(milliseconds: 500),
+                crossFadeState: instanceIcon == null
+                  ? CrossFadeState.showFirst
+                  : CrossFadeState.showSecond,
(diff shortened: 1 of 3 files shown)


CTalvio:
I like this a lot. Only thing I'd change is add a fade animation to the icon changing...not that I know how to do that in flutter yet.

micahmo:
> Only thing I'd change is add a fade animation to the icon changing

Done!

https://github.com/hjiangsu/thunder/assets/7417301/14ee9a8e-2c11-4789-bd66-a7edab6cbf64

machinaeZER0:
Just wanted to say, this is lovely! The fade was a really nice touch :)

hjiangsu:
This looks really good, I like the little details you put into it! I think `AccountExtended` is okay for now - we could potentially class inheritance somehow to make it cleaner but I think its good :D

Issue #200, opened 2023-07-03
Indicate originating instance of comments


micahmo:
**Is your feature request related to a problem? Please describe.**
Since comments can be posted cross-instance, it's not always clear where a comment came from (whether it's my current instance, the instance of the post, or another one).

**Describe the solution you'd like**
Show the originating instance in the comment view. I'm not sure whether this should always be shown (which might clutter things) or be available from a menu.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

CTalvio:
I'd prefer the usernames become tappable to open the user profile, same as poster usernames right now.

Eventually comment should also be long tappable, like posts, to view additional info and a menu with actions outside the configured gestures.

It doesn't really matter what instance someone is on, so in the name of minimalism, I would not display it.

CTalvio:
#281 will make comment usernames tappable

micahmo:
I'm happy having these details available via tooltip (#319), although I believe @hjiangsu said he might consider showing this on the main UI (optionally), so I will leave this open for now.

PR #319, opened 2023-07-11
Additional info/icons

micahmo:
In the name of minimalism, I agree with the comments on my issue #200 that we don't have to show more than is necessary by default. However, I still like to have the information accessible quickly, which is where tooltips come in handy. This PR does the following via tooltips:

- Show the full user/instance name on posts and comments (this becomes even more important with display names).
- Show the full community/instance name on posts.

I also wanted to tackle special users. For me, (a) the colors are very subtle and (b) I can't remember what they mean. Therefore I added...
 - Icons for OP, self, and admin.
 - Tooltips to describe this.

Finally, I fixed a "bug" where accounts with identical usernames from different instances could both be identified as the current user.

---

### Demos

https://github.com/hjiangsu/thunder/assets/7417301/02e63e5d-d854-4eb1-9a07-0028092c947f

https://github.com/hjiangsu/thunder/assets/7417301/8e35fbf5-ce5f-42ca-b34c-fc0d8ba9fc24

https://github.com/hjiangsu/thunder/assets/7417301/8f355885-f8df-46f9-b0d8-9808bf110ea3

Can't show the "me" icon without leaking my username, but it's the person icon that we used to see in the account switcher.

### DIFF
--- a/lib/post/widgets/comment_card.dart
+++ b/lib/post/widgets/comment_card.dart
@@ -110,9 +110,8 @@ class _CommentCardState extends State<CommentCard> with SingleTickerProviderStat
 
     final theme = Theme.of(context);
 
-    // Checks for either the same creator id to user id, or the same username
-    final bool isOwnComment = widget.commentViewTree.comment?.creator.id == context.read<AuthBloc>().state.account?.userId ||
-        widget.commentViewTree.comment?.creator.name.toLowerCase() == context.read<AuthBloc>().state.account?.username?.toLowerCase();
+    // Checks for the same creator id to user id
+    final bool isOwnComment = widget.commentViewTree.comment?.creator.id == context.read<AuthBloc>().state.account?.userId;
 
     final bool isUserLoggedIn = context.read<AuthBloc>().state.isLoggedIn;
 
--- a/lib/post/widgets/comment_header.dart
+++ b/lib/post/widgets/comment_header.dart
@@ -7,6 +7,7 @@ import 'package:thunder/core/models/comment_view_tree.dart';
 import 'package:thunder/account/bloc/account_bloc.dart' as account_bloc;
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
 import 'package:thunder/utils/date_time.dart';
+import 'package:thunder/utils/instance.dart';
 import 'package:thunder/utils/numbers.dart';
 import 'package:thunder/user/pages/user_page.dart';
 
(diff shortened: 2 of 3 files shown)


hjiangsu:
I like the subtle changes, especially with the use of the Tooltips rather than attaching them to the UI (eventually, I think this should also be included in the UI but as a customizable feature). One thing I do want to maybe change up or get more feedback on are the icons to represent the different users

I mentioned it in this comment: https://github.com/hjiangsu/thunder/issues/175#issuecomment-1620958344, but I'll summarize it here too: Should we use icons or chips, or another method as a way to represent the user information? Which one is better from a minimalistic/usability perspective?

If we do agree with icons, then I would maybe switch up some of the icons:
- The shield icon seems to be better reserved for moderator status
- The poster icon (mic) seems to be okay, but is there maybe a better icon for it?
- I think the person icon to indicate yourself seems good!
- Admin icon could use a different icon if the shield icon ends up representing the moderator - potentially a variation of a shield icon with a different colour?
image

I'll pull in @CTalvio and @machinaeZER0 into this to see what their thoughts are!

machinaeZER0:
- I kinda like the idea of using chips for self, just to cut down on unnecessary icons, but the person icon is also good!
- I like the shield idea for sure, but not sure how to differentiate between admin and mod. Maybe a wrench icon for mods?
- I think a microphone icon makes sense for OP, though I think a different version of it would be better - the line work feels like it matches less with the style of the other icons in the app? Feel like the other icons have slightly thicker lines. Maybe something like this? [https://cdn4.vectorstock.com/i/1000x1000/43/18/microphone-icon-vector-21894318.jpg](url)
- I'd consider reducing the space between the username and the icons, but not by a ton!
- Only thing that feels particularly off layout wise is that (since up and down votes are shown), there's potentially a lot of information clustered around the top left if you go the icon route. Not sure if that's just me though. If others felt similarly then one might consider moving the vote scores to the top right, next to the timestamp? Also thinking about a future in which user icons might also be displayed, otherwise I'd suggest moving the icons to the left of the username

micahmo:
Thanks for the feedback!

1. If we used chips, would it be an icon + word, or just word? If the latter, would we do something like `Admin`, `Mod`, `Me`, and `OP`? It seems like chips would take up more space than icons, but they may be "cleaner" I suppose.
2. For icons, I think we're agreed on the mic (stolen from Reddit) and person. Shield for admin is good and wrench for mod is good. Are there any other options than the icons [here](https://api.flutter.dev/flutter/material/Icons-class.html)? For example I don't see `Shield Person` (which I quite like).
3. In terms of the layout, I'm totally open to shifting things around. I think the user identifiers should be the minority of comments, so I'm not too afraid of overloading that corner of the comment. But I'll take a swing at whatever we agree on.

@hjiangsu Any specific direction here?

machinaeZER0:
I actually hadn't thought of the chips being used as a separate badge, though for shorter words I don't dislike that idea. Rather, I think in some apps the username becomes a chip where relevant to help it stand out more, and given a certain assigned color the user will associate with the role in question. This is what Jerboa does for some of its roles (in the attached example I believe the chip is representing OP, though sunaurus has so many roles I can't say that with confidence!)

micahmo:
> I think in some apps the username becomes a chip where relevant to help it stand out more, and given a certain assigned color the user will associate with the role in question.

I see what you mean. Relay does this too. I have no objection to this design, but it has the same fundamental problem for me which is that I can't always remember what the colors mean. However, I suppose the tooltips help.

Let me see if I can implement chips for usernames, and we can discuss icons further.

micahmo:
Added chips.

| image |
| - |

| image |
| - |

| image |
| - |

CTalvio:
Tooltips are a great idea!

Only thing you should change, is to make them appear above, instead of below. Appearing below makes hirearchical sense, and on desktop its best practice. But on touch devices there's the additional consideration of things appearing below where you tap, being obscured by the users finger.

Not sure the wrench icon will be self-evident to a lot of users. I think using the shield icon for both would work best, but with a stack to carve out either an "M" or "A" inside the shield, that will make its meaning instantly recognisable.

For OP indicator, I might've used a simple text badge saying "OP" before the username. If we stick with a mic icon, we'll need to change it, that one is associated with sound settings and actual audio recording. Something more "stage mic"y should be the goal.

CTalvio:
Tooltips are a great idea!

Only thing you should change, is to make them appear above, instead of below. Appearing below makes hirearchical sense, and on desktop its best practice. But on touch devices there's the additional consideration of things appearing below where you tap, being obscured by the users finger.

Not sure the wrench icon will be self-evident to a lot of users. I think using the shield icon for both would work best, but with a stack to carve out either an "M" or "A" inside the shield, that will make its meaning instantly recognisable.

For OP indicator, I might've used a simple text badge saying "OP" before the username. If we stick with a mic icon, we'll need to change it, that one is associated with sound settings and actual audio recording. Something more "stage mic"y should be the goal.

Edit: as mentioned, consider moving the user icons closer together, almost as if part of the username, and having a larger gap for the votecounts.

machinaeZER0:
Oh yeah, an M or A chip might actually be a great way to go for those two roles! I like the wrench idea, but letter chips feel super easy to parse. Hadn't thought about truncating them in that way. 

That was a good call on tooltips appearing above/out of the way of your thumb/finger!

micahmo:
> Maybe even make the icon be inside the chip with the username?

Setting aside the icons themselves for now, how do we feel about this?

| image |
| - |

| image |
| - |

| image |
| - |

hjiangsu:
I like the overall changes that have been made! It feels much cleaner to me and more distinguishable for different special roles (admin/mod/op/me)

I do wonder about the M/A for the moderator/admin - we might have to take into consideration localization if this is the way we want to go about differentiating them like @micahmo mentioned

It kinda sucks that Flutter doesn't incorporate all the material icons, I think the shielded person would've been a nice one to use

> Another argument in favor of keeping icons is that it lets us denote multiple attributes of a user. For example, you could be an admin, the OP, and yourself.

Thats also a good point I never thought of!

machinaeZER0:
Yeah, I think consistency is the way to go here. If we're concerned about the letters not making sense or needing/not needing to localize then I wouldn't go down that route :)

CTalvio:
I wasn't suggesting we use just the letter alone.

We can add the letter to the shield icon, using a stack widgets which contains the letter, on top of the icon.

micahmo:
I think we're mostly in agreement except for the icons. I'm skeptical that we can construct anything legible in the space we have using a `Stack`. Here was a very brief attempt at adding "M".

image

I explored other options and found it's not too hard to generate new fonts/icons using [this guide](https://medium.com/codechai/how-to-use-custom-icons-in-flutter-834a079d977) and [this site](https://www.fluttericon.com/). I think [shield-home](https://pictogrammers.com/library/mdi/icon/shield-home/) and [shield-account](https://pictogrammers.com/library/mdi/icon/shield-account/) are good candidates for mod and admin, respectively, and [microphone-variant](https://pictogrammers.com/library/mdi/icon/microphone-variant/) for self.

How do we feel about this?

| image |
| - |

| image |
| - |

| image |
| - |

machinaeZER0:
Those look awesome!! Chips really help it pop, and I like the new spacing between name and icons :)

micahmo:
Pushed latest changes if you want to take a look @hjiangsu and @CTalvio.

CTalvio:
Very nice. 

You can fit the letters a lot bigger if you use the solid shield, and set the letter to use the same color as the chip. This way you can "cut out" the letter into the shield. This was my intent. The icons do end up quite small in the chip, though.

I think we should use just the [shield](https://pictogrammers.com/library/mdi/icon/shield/) for mods.

And shield crown, for [admins](https://pictogrammers.com/library/mdi/icon/shield-crown/). Or perhaps the house shield.

With the house and person, it can be difficult to intuit which means what, but I think if we make moderators "normal" and the admins "special", that makes it quite clear which is which.

micahmo:
My only objection to the solid shield is that it kinda becomes a blob at that size. See one my earlier attempts.

|  |
| - |

I also don't think the icons will ever be completely intuitive, which is why we have the full description in the tooltip.

@hjiangsu Want to arbitrate? 😆 How about shield house for mods and shield crown for admins?

CTalvio:
Nothing is very sharp in general on a desktop monitor, compared to a real phone screen, but I get your point.

I think one with nothing, and one with something, will be much more distinct than if both have something. Both somethings will be even tinier than the shield and hence even harder to tell apart in that way.

hjiangsu:
I think having one solid, and the other one with an icon in the center works. That _should_ be distinctive enough to tell them apart. For now, we can apply the suggestions @CTalvio made earlier.

If anything, I feel like making the overall chip slightly bigger would solve the issues with not being able to tell them apart

micahmo:
Thanks for the feedback again! Good point about bad resolution on desktop. Also the MDI solid shield has a little more distinction than the Flutter one, which I like. I like crown better than house for admin. I also didn't want to expand the chip because it currently matches the "New" and collapsed comment counts.

Thoughts?

| image |
| - |

CTalvio:
I think its perfect. The gap between the icons is just right, but I'd make the one between the username and the icons, just a pixel or two wider.

machinaeZER0:
Agree - personally I still like shield for mods and a wrench for admins (just to have two very distinct things), but if others feel strongly then not too concerned :) people will figure out what the icons mean through context I think.

micahmo:
Do we have our winner? 😊

| image |
| - |

| image |
| - |

| image |
| - |

CTalvio:
I think so.

@hjiangsu ?

hjiangsu:
I'd agree! I'll merge it in

Issue #284, opened 2023-07-09
Can’t paste into community search box


phareous:
**Description**  
You can’t paste text into the community search box

**How to Reproduce**  

**Expected Behavior**  

**Screenshots**  

**Device & App Version:**  
 - Device: [e.g. iPhone6] iPhone 12 mini
 - OS: [e.g. iOS8.1] 16.5.1

**Additional Context**  

![IMG_3540](https://github.com/hjiangsu/thunder/assets/1590943/d30a0b45-f1c5-4ea7-839b-56104edd5e8b)

PR #321, opened 2023-07-11
#284 Community Search Paste Fix

ajsosa:
Bug: https://github.com/hjiangsu/thunder/issues/284

SearchBar for some reason doesn't implement the copy/paste functionality. Converted the SearchBar to a TextField, retaining all the previous functionality and look. 

The only thing that I was unable to get working is the Inkwell effect but otherwise it should look and function more or less the same as before.


hjiangsu:
Looks good!

PR #334, opened 2023-07-12
Truncate long community names in drawer

### LINKED ISSUE #333: side panel text cuts off if too long
I think this would have been a thing before, but the addition of community icons in the sidebar has pushed longer community/instance names far right enough that they can get cut off:

InkedScreenshot_20230711-070155

It might make sense at some point in the future to truncate sidebar content with ellipses? Or line wrap, but I kinda like that everything takes up the same amount of space here.

(Noting that the screenshot isn't the newest version, but it is still happening in the latest apk)

micahmo:
Fixes #333. In addition to truncating long names, the full text is also now available in a tooltip.

### Before

| image |
| - |

### After

| image |
| - |

#### With Tooltip

| image |
| - |

### DIFF
--- a/lib/community/widgets/community_drawer.dart
+++ b/lib/community/widgets/community_drawer.dart
@@ -185,20 +185,27 @@ class _CommunityDrawerState extends State<CommunityDrawer> {
                                             ),
                                           ),
                                           const SizedBox(width: 16.0),
-                                          Column(
-                                            mainAxisAlignment: MainAxisAlignment.start,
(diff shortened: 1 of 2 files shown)


micahmo:
Made the same fix in the search page.

### Before

| image |
| - |

### After 

| image |
| - |

#### With Tooltip

| image |
| - |

hjiangsu:
Looks good!

machinaeZER0:
Beautiful work @micahmo!

PR #338, opened 2023-07-12
Back goes home

### LINKED ISSUE #243: All screens should go back to home feed before exiting app

micahmo:
Fixes #243: Going back from any page but the main one navigates to the main one. After that, it still takes two backs to exit.

### Demo

https://github.com/thunder-app/thunder/assets/7417301/3b84229e-220e-4169-bce7-cfe4dfbc91cb

### DIFF
--- a/lib/thunder/pages/thunder_page.dart
+++ b/lib/thunder/pages/thunder_page.dart
@@ -111,6 +111,14 @@ class _ThunderState extends State<Thunder> {
   }
 
   Future<bool> _handleBackButtonPress() async {
+    if (selectedPageIndex != 0) {
+      setState(() {
+        selectedPageIndex = 0;
+        pageController.animateToPage(0, duration: const Duration(milliseconds: 500), curve: Curves.ease);
+      });
+      return Future.value(false);
+    }
+    
     if (appExitCounter == 0) {
       appExitCounter++;
       _showExitWarning();


machinaeZER0:
I don't know if it's this PR specifically or some other interaction, but I'm finding that while the back action works as intended, it also fairly regularly (but not 100% of the time) triggers the "press back again to exit" popup immediately (where in theory it should take an additional swipe to trigger that action). Any idea why that might be?

micahmo:
@machinaeZER0 You're saying a back press from another page (like settings) goes home and immediately triggers the toast?

machinaeZER0:
@micahmo it seems to be happening when i swipe out of a post, back into the main feed or a community. I am still having trouble replicating 100% but it seems like mayyyybe it's more likely to happen on a longer/more aggressive swipe (rather than a flick)? And on occasion I've gotten the app to actually exit on me, which seems to indicate it's registering three swipes worth of swipes. Very odd!

If it's helpful, I generally swipe right to exit stuff, and it seems like going in and out of a post rapidly makes it more likely to happen. Let me know if I can provide any more info, or if I should file a bug? I'm using an android phone, as well, and I'm testing the latest pre-alpha.

micahmo:
> it seems to be happening when i swipe out of a post

Aha! I can repro now.

https://github.com/thunder-app/thunder/assets/7417301/17ce553f-9ba3-45fe-a807-9c4b5cf647bc

This is all speculation so far, but I think the reason for the initial toast is that we're already on the first tab (despite being within a post). Maybe reason for the exit is that we're still within the window of a previous toast (if that makes sense).

machinaeZER0:
I think I understand what you mean! Is there some way to test that? And would that explain the instances where sometimes we can get it to straight up exit the app? I've found at times it will trigger even if I'm not navigating the app all that quickly

machinaeZER0:
Okay, new info (I think) - so I'm not sure I can trigger this by swiping left, but I was testing swiping right but not lifting my thumb up at the end? In theory (I believe) nothing should actually happen until I let go of the action; however, sometimes the swipe action itself causes the Back behavior on its own, and in that case when I lift up my thumb the toast pops up 100% of the time

micahmo:
I just discovered (totally missed this in #245!) that swiping left-to-right on the post body goes back. I think that gesture is activating, going back, and then the actual back gesture from Android is activating. Does that seem right?

PR #342, opened 2023-07-12
Improve login flow

micahmo:
The goal of this PR is to improve the login flow. Currently, you are always kicked back to the main feed after a login attempt. After a successful attempt, this may be confusing at best. After a failed attempt, the experience is much worse. First you have to acknowledge the error and let it refresh the main page. Then you have to navigate back to the accounts page and start the login process all over again.

Now we have the following flow:
 * When logging in, the login page stays up. The login button is grayed out to indicate that the log in is in process.
 * If there is an error, it will be displayed directly on that page. It is easy to tweak the inputs and try again.
   * For this, I moved the responsibility of handling `AuthState.loading` and `AuthState.failure` to the `login_page` instead of the `thunder_page`. I'm not sure if it's typical to split bloc handling this way.
 * If the login is successful, the login page hides and the profile page is visible.
   * For this, I had to fix a small issue where the profile bloc handler was using the stale states.

I would highly recommend a few people test this before we merge/release!

Note: I'm not too sure I like the layout changing when error appears. And I don't think we can reserve the space because we don't know how many lines it will be. How do we feel about either...
* aligning the input fields at the top and the buttons at the bottom
* aligning everything at the top and putting the error at the bottom
* or is it fine as is

### Demo

https://github.com/thunder-app/thunder/assets/7417301/e838d151-5647-4e9e-a9f4-32cb5be8a13a

---

P.S. I'm not sure if this is your standard practice, but I highly recommend reviewing Dart/Flutter with whitespace changes hidden. Otherwise, even just by surrounding one widget with another, it will look like everything in between was changed.

image

### DIFF
--- a/lib/account/pages/account_page.dart
+++ b/lib/account/pages/account_page.dart
@@ -22,31 +22,40 @@ class _AccountPageState extends State<AccountPage> {
     AuthState authState = context.read<AuthBloc>().state;
     AccountState accountState = context.read<AccountBloc>().state;
 
-    return BlocConsumer<AuthBloc, AuthState>(
-      listener: (context, state) {},
-      builder: (context, state) {
-        if (authState.isLoggedIn && accountState.status == AccountStatus.success) return UserPage(userId: accountState.personView!.person.id, isAccountUser: true);
-        return Center(
-          child: Padding(
-            padding: const EdgeInsets.symmetric(vertical: 8.0, horizontal: 24.0),
-            child: Column(
-              crossAxisAlignment: CrossAxisAlignment.center,
-              mainAxisAlignment: MainAxisAlignment.center,
-              children: [
-                Icon(Icons.people_rounded, size: 100, color: theme.dividerColor),
-                const SizedBox(height: 16),
-                const Text('Add an account to see your profile', textAlign: TextAlign.center),
-                const SizedBox(height: 24.0),
-                ElevatedButton(
-                  style: ElevatedButton.styleFrom(minimumSize: const Size.fromHeight(60)),
-                  child: const Text('Manage Accounts'),
-                  onPressed: () => showProfileModalSheet(context),
-                )
-              ],
-            ),
+    return MultiBlocListener(listeners: [
+      BlocListener<AuthBloc, AuthState>(
+        listener: (context, state) { 
+          setState(() => authState = state); 
+        },
+      ),
+      BlocListener<AccountBloc, AccountState>(
+        listener: (context, state) { 
+          setState(() => accountState = state); 
+        },
+      ),
+    ],
+    child: (authState.isLoggedIn && accountState.status == AccountStatus.success && accountState.personView != null)
+      ? UserPage(userId: accountState.personView!.person.id, isAccountUser: true)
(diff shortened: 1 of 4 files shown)


machinaeZER0:
Would it help for the error to just be a toast? Feel like we've been gravitating towards that for a lot of informational messages, and then no space needs to be reserved

machinaeZER0:
Personally I like that logging in brings me to the feed (when successful), so taking me to my profile feels like an extra step - but if others like it that's fine! I think for me I'd consider this workflow:

Failed login: toast error popup, login page stats open

Successful login: toast SUCCESS popup ('username login successful' or something) and redirects to your feed

Thoughts on that?

micahmo:
> Would it help for the error to just be a toast?

Great idea!! I added toasts for failed and successful login attempts.

https://github.com/thunder-app/thunder/assets/7417301/7c39d8cc-6614-4f2c-95b5-835bede3a968

> Personally I like that logging in brings me to the feed (when successful), so taking me to my profile feels like an extra step - but if others like it that's fine!

I'm not set in stone here. The only thing I'd say is that, now that back brings you home, it's pretty easy to navigate from accounts to the main feed. But let's see what others say.

CTalvio:
One idea, since you are already checking if there is an image to load from the instance entered, why not make the "invalid instance" error appear before even attempting a log-in? Is such a check possible?

You'd use a similar error text as before. Many UIs do this, like when they detect an email as-typed to be invalid, or the like.

micahmo:
> since you are already checking if there is an image to load from the instance entered, why not make the "invalid instance" error appear before even attempting a log-in?

This is not a bad idea. I think we need a general way to detect whether a given URL is a valid Lemmy instance (it was also referenced in #343).

Maybe something like this would work.

``` dart
try {
  await LemmyApiV3(url!).run(const GetSite());
  return true;
} catch (e) {
  return false;
}
```

Combined with form validation, this could look good. Note that I would argue in favor of still allowing a login attempt (just in case our check is flawed in some way).

https://github.com/thunder-app/thunder/assets/7417301/8a895649-d8e8-4058-8222-4fe3063fcca6

Thoughts?

machinaeZER0:
That's pretty cool! Would it be possible to have it wait an extra second or so before throwing up an error, for slower typers?

micahmo:
> Would it be possible to have it wait an extra second or so

Sure. It's already debouncing at 500ms (same as the icon). It's a bit of a balancing act because if it's much slower, then it will also be slower to resolve once you've corrected it. How does one second look?

https://github.com/thunder-app/thunder/assets/7417301/dc1373d1-a04c-4eab-92e5-7534f903c8e6

machinaeZER0:
I think that feels like an improvement, personally! Defer to you and others if they feel strongly one way or another :)

micahmo:
@CTalvio and @hjiangsu How do you feel about the field "validation"? If you like it, I can push for review.

CTalvio:
I think you nailed it.

micahmo:
Ready for review. Note that there are some formatting changes due to the new workspace settings.

Issue #176, opened 2023-07-02
Use browser provided Custom Tabs on Android


vfosnar:
**Is your feature request related to a problem? Please describe.**
the webview does not respect user selected browser (no addons support, etc.)

**Describe the solution you'd like**
Use Android Custom Tabs: https://android.stackexchange.com/a/219950

**Describe alternatives you've considered**
Webview - no addons
Open in external browser automatically - creates a new tab for every link I click

**Additional context**
None

hjiangsu:
Interesting - I'll have to take a closer look at that when I have the time. Thanks for sharing it!

jordan1776:
If I can add a feature to this is to include support for chrome private custom tabs as well. Jerboa has this feature.

micahmo:
A potentially related issue is the fact that, if you navigate several levels deep in the embedded browser, pressing back will go back to the post. Android Custom Tabs might handle this better.

PR #344, opened 2023-07-12
Use custom tabs for embedded browser

micahmo:
This PR uses [flutter_custom_tabs](https://pub.dev/packages/flutter_custom_tabs) for in-app web navigation.

As much as I love the `WebView` that @hjiangsu designed, I think best practices are moving towards using native browser experiences even within apps. This was mentioned in #176. My personal motivation for this is to get back/forward navigation. But there are plenty of other features we'd get for free (e.g., find in page) and overall less maintanance. In addition, we won't lose any of the features you added, such as refresh, open in browser, and share.

Note: I believe the behavior should be similar (and native) for iOS, but it should be tested.

### Demo

https://github.com/thunder-app/thunder/assets/7417301/32318a82-04b4-4a9a-9725-15925c47ec5b

### DIFF
--- a/lib/settings/pages/about_settings_page.dart
+++ b/lib/settings/pages/about_settings_page.dart
@@ -1,8 +1,9 @@
 import 'package:flutter/material.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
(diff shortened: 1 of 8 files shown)


hjiangsu:
Looks good!

PR #346, opened 2023-07-12
Combine theme options into a list

coslu:
Combines the individual toggle options for system theme, dark theme and pure black theme into a `ListOption`. 

The toggles were always overriding each other, which could be confusing, and only one of system/light/dark/black would be active at a time, so a list made more sense.

Also changes the default theme to system instead of dark. 

Here is how it looks: 

Screenshot_20230712_232702

### DIFF
--- a/lib/core/enums/theme_type.dart
+++ b/lib/core/enums/theme_type.dart
@@ -1,1 +1,1 @@
-enum ThemeType { white, black }
+enum ThemeType { system, light, dark, pureBlack }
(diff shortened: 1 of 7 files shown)


micahmo:
Very nice, this was much needed!

machinaeZER0:
This looks great! I haven't looked at the code, just the screenshot - should the icons for dark and black mode be more differentiated? In the current version they are different (though they definitely could be differentiated further)

coslu:
Good idea. This looks better, right?

Screenshot_20230713_201905

hjiangsu:
Yup, I think that looks good!

PR #353, opened 2023-07-13
Various Comment Thread Modifications

hjiangsu:
This changes some logic regarding comment threads and how they are fetched.
- Fetching comments on a post now has a max depth
- For comments that have replies that have not been fetched yet, there is a new indicator to allow the user to load up more comments
- You can pass in a `commentParentId` to `_getPostCommentsEvent` to fetch comments from that given  `commentParentId`
- Renamed `comment` in `CommentViewTree` to `commentView` to better match its type
- Updated lemmy_api_client to the latest version

Here's a gif of how it looks like:

comment_thread

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Settings have been re-organized to be more consistent, and to show available options between different views - contribution from @CTalvio
 - Feed view will no longer show full screen error messages
 - Added comment button actions and added an option to toggle comment button actions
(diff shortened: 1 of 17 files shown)


micahmo:
Looks great! Two thoughts.

Should the comment be collapsed after loading? Seems like you always need two presses to keep reading.
It looked like one of the comments was already collapsed instead of saying "Load x more replies". Not sure why.

hjiangsu:
> Should the comment be collapsed after loading?

I think that was a bug - I'll check it out and see

PR #355, opened 2023-07-13
user logout dialog

### LINKED ISSUE #351: [Suggestion] Confirmation popup for logging out account
As the title suggests, a popup confirmation for logging out would be nice to prevent it from happening by accident.

ggichure:
Fixes #351

> Confirmation popup for logging out account

logout


micahmo:
Do you want to wait until #342 is done? Specifically, after logout, the new flow should leave you on the accounts page. Just want to make sure there are no conflicts.

Also, the wording seems a bit stilted to me (maybe just me). How about a question, like "Are you sure you want to log out?". Also are we sure we want to mention "instance"? We could still be logged into the same instance as another user.

ggichure:
> Do you want to wait until #342 is done? Specifically, after logout, the new flow should leave you on the accounts page. Just want to make sure there are no conflicts.
> 
> Also, the wording seems a bit stilted to me (maybe just me). How about a question, like "Are you sure you want to log out?". Also are we sure we want to mention "instance"? We could still be logged into the same instance as another user.


I'll wait. for #342 
I'll also chnage the wording to "Are you sure you want to log out?"

hjiangsu:
I have just merged in #342!

ggichure:
This is ready. @hjiangsu

PR #365, opened 2023-07-14
Home feed button scrolls to top

### LINKED ISSUE #149: Tap Home button to scroll to top of feed/return to main feed?

micahmo:
Fixes #149. I know this is somewhat redundant now with the FAB, but there's also been some talk about removing the the FABs or at least making them optional.

https://github.com/thunder-app/thunder/assets/7417301/c328409a-75a8-4ed2-933f-6a97bf6ffeb5

### DIFF
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -47,6 +47,7 @@ class PostCardList extends StatefulWidget {
 class _PostCardListState extends State<PostCardList> {
   final _scrollController = ScrollController(initialScrollOffset: 0);
   bool _showReturnToTopButton = false;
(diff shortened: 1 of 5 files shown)


machinaeZER0:
Yessssss! Thank you for working on this, that's huge!

hjiangsu:
Looks good!

Issue #63, opened 2023-06-25
Default search sort order doesn't seem relevant


Benjamint22:
**Description**  
The default sort order in the "Search" tab doesn't seem to give particularly relevant results for some reason.

I tested changing the sort order in `search_bloc.dart` from default to `SortType.Active`, and it looks like the results are much more relevant (subjectively).

I'm not sure what's the default sort order. It looks like it may be "Hot" since it gives me the most similar results. I'm not sure what the most relevant sort order is, but I think "Active" or "Top[somerange]" looks best.

I can open a PR if you want.

**How to Reproduce**  
1. Go to the search page.
2. Search "technology"

**Expected Behavior**  
I expected the most popular (or active) communities to be near the top results, but have to scroll far to find them.

**Screenshots**  

<table>
<tr>
 <th width="33%">Default (current)
 <th width="33%">`SortType.Active`
 <th width="33%">`SortType.TopAll`
<tr>
 <td>

![image](https://github.com/hjiangsu/thunder/assets/6548393/cebe2aa2-79a3-4ed8-a3a8-edb2a795fb91)
 <td>

![image](https://github.com/hjiangsu/thunder/assets/6548393/d8d6643d-323a-4d21-b7e5-3f2cc0fe1cea)

 <td>

![image](https://github.com/hjiangsu/thunder/assets/6548393/72261836-c3f0-434d-8049-356076a21c55)
</table>

**Device & App Version:**  
 - Device: ROG Zephyrus G14 GA401QM_GA401QM
 - OS: Microsoft Windows 11 Home 10.0.22621 Build 22621
 - Version: [v0.2.1-prerelease+5](https://github.com/hjiangsu/thunder/releases/tag/v0.2.1-prerelease%2B5)

**Additional Context**  
I am on the `lemmy.world` instance and I'm authenticated.

Benjamint22:
This is subjective, no hard feelings if you disagree :stuck_out_tongue_winking_eye:

hjiangsu:
I agree - we should probably make the default either active or top. I'm not too sure what the default is either, but I'll leave it to your discretion!

In the future, it would be best to add in a way to change between sort types within the search as well

Benjamint22:
> I agree - we should probably make the default either active or top. I'm not too sure what the default is either, but I'll leave it to your discretion on the default it should be!

Awesome! Will open a PR.

> In the future, it would be best to add in a way to change between sort types within the search as well

Fully agreed!

[...]

Issue #22, opened 2023-06-19
Changing sort modes or refreshing resets view to Local Posts


jordan1776:
**Describe the bug**
Changing sort modes or refreshing the feed resets view to Local Posts

**To Reproduce**
1. Go subscriptions or All Posts
2. Change sort to anything
2 alternative. Refresh feed
3. See error

**Expected behavior**
changing sort or refreshing shouldn't affect the feed loaded 

**Smartphone (please complete the following information):**
 - Device: Xperia 1 II
 - OS: Android 11

hjiangsu:
Hey! I also noticed that during development and have applied some fixes to it. The "refresh feed" bug should be fixed in an upcoming release

hjiangsu:
This should hopefully be fixed in v0.2.0-alpha! I'll close this issue for now, but do let me know if it is still occurring

jordan1776:
This refresh bug is fixed but it seems like refreshing the page changes the sort to hot, I'll put an issue in

PR #373, opened 2023-07-15
Feature/search sorting

micahmo:
1. Support for sorting communities in search (mentioned in [#63](https://github.com/thunder-app/thunder/issues/63)).
   - Note that I didn't make this a setting like Post or Comment sorting. Instead, I just remember the current selection for next time.
2. Support for paging search results (looks like this was only half implemented).

https://github.com/thunder-app/thunder/assets/7417301/a6d2b4d1-7ba1-4473-808a-46557049ae8b

### DIFF
--- a/lib/search/bloc/search_bloc.dart
+++ b/lib/search/bloc/search_bloc.dart
@@ -21,7 +21,7 @@ EventTransformer<E> throttleDroppable<E>(Duration duration) {
 }
 
 class SearchBloc extends Bloc<SearchEvent, SearchState> {
-  SearchBloc() : super(const SearchState()) {
+  SearchBloc() : super(SearchState()) {
     on<StartSearchEvent>(
       _startSearchEvent,
       transformer: throttleDroppable(throttleDuration),
@@ -59,7 +59,7 @@ class SearchBloc extends Bloc<SearchEvent, SearchState> {
         sort: event.sortType,
       ));
 
-      return emit(state.copyWith(status: SearchStatus.success, results: searchResponse, page: 2));
+      return emit(state.copyWith(status: SearchStatus.success, communities: searchResponse.communities, page: 2));
     } catch (e, s) {
       return emit(state.copyWith(status: SearchStatus.failure, errorMessage: e.toString()));
     }
@@ -73,7 +73,7 @@ class SearchBloc extends Bloc<SearchEvent, SearchState> {
 
       while (attemptCount < 2) {
         try {
-          emit(state.copyWith(status: SearchStatus.refreshing, results: state.results));
+          emit(state.copyWith(status: SearchStatus.refreshing, communities: state.communities));
 
           Account? account = await fetchActiveProfileAccount();
           LemmyApiV3 lemmy = LemmyClient.instance.lemmyApiV3;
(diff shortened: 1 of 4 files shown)


hjiangsu:
I think it looks good overall! Just one UI thing - should the sort by be after of before the clear button? Which one makes more sense from a UI perspective?

I took a look at the material guidelines but didn't see any examples of this, so I'd like to open up discussion about this: https://m3.material.io/components/search/guidelines

micahmo:
> should the sort by be after of before the clear button?

I put it after because we currently hide the clear button when there is no text. Therefore, if the sort button were before, there would either be a blank space when there's no clear button, or the sort button would move around. I didn't like the sound of either of those.

I guess the other option is to always show the clear button.

hjiangsu:
You made some good points - so I'll stick with the current way things are implemented!

machinaeZER0:
Apologies if this is outside the scope of this PR - should we consider updating the default search sorting? It looks like currently it sorts by Hot by out of the gate, but in my experience that means a lot of less relevant results (with far fewer subscribers) show up before the thing you may have been looking for. Would it make sense to set the default to something equivalent to Top?

micahmo:
The default sort used to be Top, but @Benjamint22 changed it to Active in #22. Personally, I agree that Top is very useful. While it's nice to surface smaller communities, I usually find I'm searching for the "main" instance of a given community, which is usually the one with the most subscribers. I think having the option to sort (and having it sticky) is good enough, but I'm certainly open to changing the default back to Top as well.

machinaeZER0:
Defer to the group (as always) but I like the idea of starting people with Top - feels sort of in line with the idea of changing the default feed view to All over Local, in my brain. In both cases people have the means to tailor their preferences, but feels like it gives people the optimal first-time experience, personally.

micahmo:
@machinaeZER0 Maybe this would be a good candidate for you to make another PR? 😉 Take a look at where the search page picks up the default sort here:

https://github.com/thunder-app/thunder/blob/45eb2b7e2f613fa5252baa13604affeaa7b7e71d/lib/search/pages/search_page.dart#L47

And where the default sort is defined here:

https://github.com/thunder-app/thunder/blob/c8110c69884259fd194388e35c07072de6f87e9b/lib/utils/constants.dart#L5

You could make a new constant like `DEFAULT_SEARCH_SORT_TYPE` with value `SortType.topAll` in the second file, then use that as the default in the first file.

machinaeZER0:
Oh jeez, cross-file shenanigans! I'm definitely interested in giving it a shot - I'll try and check this out this weekend!

Benjamint22:
if you guys decide to switch it to top, might be worth looking into switching it to top of a shorter time span than `all` (like year). I don't know for how many people that's the case, but generally, when I search for a community, it's because I want to know which one will result in the greatest number of good posts in my feed, and top all could return dead communities (although I'm not sure how concerning that is 🤷).

machinaeZER0:
would that just be SortType.topYear? which file houses all of these options? Either way your reasoning makes sense to me!

machinaeZER0:
image
@micahmo Is this all I would be changing in this file once I've made the new const in thunder/lib/utils/constants.dart?

Benjamint22:
> would that just be SortType.topYear? which file houses all of these options?

I think https://github.com/thunder-app/lemmy_api_client/blob/master/lib/src/enums.dart#L49

machinaeZER0:
PR #422 has been created 😄  please let me know if I did that right, if anyone has time at some point?

PR #380, opened 2023-07-16
Add option to show community icons in post card

sant0s12:
I thought it would be cool to have an option to show the community icons in the comfortable post cards so I implemented it.

I've never coded in Flutter before so I am very open to feedback in case I implemented it incorrectly or missed something.

This is a preview of what it looks like:
Preview

### DIFF
--- a/lib/community/widgets/community_drawer.dart
+++ b/lib/community/widgets/community_drawer.dart
@@ -8,6 +8,7 @@ import 'package:thunder/account/bloc/account_bloc.dart';
 import 'package:thunder/community/bloc/community_bloc.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
 import 'package:thunder/core/singletons/lemmy_client.dart';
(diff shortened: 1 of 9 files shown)


micahmo (commented):
Very nice, love the `CommunityIcon` class refactoring.

hjiangsu:
@sant0s12 Did you want me to merge this in? or would it be better to close this PR and focus on #381 since it also includes the changes you made here?

I'm not too sure if there was any logic that was changed for community icons, so please let me know! Otherwise, I think it looks good!

sant0s12:
@hjiangsu I don't mind either way. I was not sure if you'd rather have them in different pull requests or all together. Guess I messed it up a bit since now the other one has both...

In terms of the community icons logic, the main difference is that the icon with just the letter is used as a placeholder while the image is loaded and in case the loading fails (e.g., when the community has no image).

hjiangsu:
Cool, in that case, I'll close this and we can merge in the other one together with all the logic!

Issue #303, opened 2023-07-09
[Request] hide community name on posts when viewing community?


machinaeZER0:
Right now when you view posts in a community, the posts themselves still show the community handle:
![Screenshot_20230709-140157](https://github.com/hjiangsu/thunder/assets/3624023/a69e7925-dd53-4746-a2b2-138017404b76)

In this example, every post says mst3k, which feels somewhat redundant. While it's not causing any harm, removing this line from this specific view will save vertical space (in compact mode, at least) when a post's title contains a few lines of text already (making it taller than the thumbnail, I mean)

Would be interested to know other people's thoughts, as I may not be thinking of an edge case where this would be helpful. I'm also not sure whether Thunder differentiates between a post in a feed versus in a community, which would be the other blocker (but I feel like it must, since the community posts don't show the instance?)

If this does sound like a good thing to implement, I could try and attempt it? Feels like it might be semi-straightforward, unless we wanted it to be toggleable. Let me know!

hjiangsu:
What I was thinking was to display the user when looking at a particular community

tunglith:
I think the user should always be displayed, maybe we could inspiration from what Jerboa and liftoff for this one, they display OP's name alongside the community icon and name.

PR #381, opened 2023-07-16
Show post author

sant0s12:
This PR builds on #380 and attempts to show the post author in each post card. In #303, it was discussed to perhaps add this.

Let me know if you would like this to be a toggleable option.

It is marked as draft because I would have to implement this for the compact layout as well.

This is what it looks like with and without the icons:


CTalvio:
Would you mind matching this to the style used in the post view, displaying the user first, with the word "to", and then the community?

machinaeZER0:
To your point, I'd be into this being a toggleable option if possible! Love how compact the feed can look right now, but definitely see the desire to include this for those that want/need it. Thanks for making the PR!

sant0s12:
> Would you mind matching this to the style used in the post view, displaying the user first, with the word "to", and then the community? This would also follow the wording the webUI uses:
> 
> image

How about something like this? 

image

I honestly like having the community name first because for it's more important to know what community a post belongs to rather than the user who posted it. But it makes sense to stick to what LemmyUI does.

sant0s12:
> To your point, I'd be into this being a toggleable option if possible! Love how compact the feed can look right now, but definitely see the desire to include this for those that want/need it. Thanks for making the PR!

No problem! I'll add an option switch for it.

CTalvio:
> I honestly like having the community name first because for it's more important to know what community a post belongs to rather than the user who posted it. But it makes sense to stick to what LemmyUI does.

You could achieve a similar effect by downsizing the username text to the same font size as "to" which you already have smaller there, or matching the color of the name to the "to". This would also make the community name the more prominent text again.

sant0s12:
I included your feedback, let me know what you think :)

CTalvio:
Have any pictures of what it looks like now?

micahmo:
+1 I'd like to see some screenshots in compact and normal view. It also definitely should be toggleable (especially for those who have gotten used to the very minimal UI by default). Also, does tapping the username bring up their profile?

hjiangsu:
Hey, just took a quick look at the code, and it looks quite good from what I can see! If you could provide some screenshots like the others mentioned, that would be great!

> Also, does tapping the username bring up their profile?

From looking at the code, it does bring you to the user page

Thanks so much for contributing :D

CTalvio:
> Hey, just took a quick look at the code, and it looks quite good from what I can see! If you could provide some screenshots like the others mentioned, that would be great!
> 
> > Also, does tapping the username bring up their profile?
> 
> From looking at the code, it does bring you to the user page
> 
> Thanks so much for contributing :D

I've been thinking we might want to disable that tap zone, so that tapping anywhere on the card except the image, opens the post.

This would be fine I think, now that long tapping post cards to do these actions is possible.

machinaeZER0:
I would agree ^ long pressing is easy and makes the cards less fiddly to interact with

machinaeZER0:
To be clear, I don't mind the behavior as long as it doesn't result in needing to be super precise to tap or avoid tapping the element in question! I've found one line/shorter elements do make that sort of thing happen a lot, especially in compact modes. Happy with wherever y'all land on this one :)

sant0s12:
Screenshot time :D

Icons + author:
image

Just autor:
image

Compact:
image

I also made it wrap if all the stuff doesn't fit in one line (it's a bit annoying that the dot hangs out alone at the end sometimes but I haven't looked into it):
image

Also, I think going to the user's profile/the community page when you tap the text is nice and I feel like it's unlikely that you would do it accidentally. But I don't have strong opinions either way.

CTalvio:
Very nice!

The user should be more greyed out in compact the same way as in normal, and I thin the community icon should be just a tad smaller, if you're making a single line the intended look

hjiangsu:
Looking good! I'm not sure if its just me, but it feels off that the user is more greyed out than the community/instance. I feel like in terms of importance, both the user and the community are roughly equals (if you've taken the time to enable usernames to show up on the cards)

Nevertheless, I'll let this minor styling choice up to the community to decide on. Once thats decided, and once the merge conflicts are fixed, I can merge this in!

machinaeZER0:
I know what you mean - I'd filed another feature request a while ago to consider making the community name a dedicated color for this reason, which would match the styling of the web clients a bit more closely. I could definitely see attributing a themed color to community name, author name or both if others felt the same!

hjiangsu:
@sant0s12 Could I get you to fix the merge conflicts and perform `dart format . -l 200`? Once those are done, I think I can merge this in if you'd like!

sant0s12:
Should be good to go :)

hjiangsu:
Merging in!

PR #384, opened 2023-07-17
Improve link previews

micahmo:
This PR attempts to fix link previews. Currently they seem to rarely/never work. I believe this is because we're solely relying in `ExtendedImage`, so they can only work if the link points directly to an image. Therefore, I pulled in [link_preview_generator](https://pub.dev/packages/link_preview_generator) to help us with this. Seems to work nicely in the majority of cases.

### Notes
* I changed a bunch of `borderRadius`es to match the hard-coded values in link preview generator. 
https://github.com/ghpranav/link_preview_generator/blob/fce54c8f67ddfb89bf62e870e00d4f466605b166/lib/src/widgets/link_view_large.dart#L85-L86
* There are also some more auto-formatting changes.

#### Normal

https://github.com/thunder-app/thunder/assets/7417301/d66234d9-3a1b-4fab-b934-041658bd62fe

#### Compact

https://github.com/thunder-app/thunder/assets/7417301/59c62d15-a7fe-4255-8d03-b2eeccb1be0f

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -17,8 +17,10 @@ class ImagePreview extends StatefulWidget {
   final int? postId;
 
   const ImagePreview({
-    super.key, required this.url,
-    this.height, this.width,
+    super.key,
+    required this.url,
+    this.height,
+    this.width,
     this.nsfw = false,
     this.isGallery = false,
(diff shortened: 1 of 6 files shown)


hjiangsu:
Looks good!

PR #370, opened 2023-07-15
Link preview for compact view

CTalvio:
Also removes the "thumbnails" on text posts, allowing long titles to overflow to fewer lines

image

image

PR #385, opened 2023-07-17
Optionally show text post indicator

micahmo:
As discussed in #370, this PR brings back the option (off by default) to show a placeholder icon for text-only posts (to bring them into uniformity with media and link posts).

https://github.com/thunder-app/thunder/assets/7417301/34b23301-8bc0-4d8c-bf0e-71fee310ce80

### DIFF
--- a/lib/community/widgets/post_card.dart
+++ b/lib/community/widgets/post_card.dart
@@ -181,6 +181,7 @@ class _PostCardState extends State<PostCard> {
                     ? PostCardViewCompact(
                         postViewMedia: widget.postViewMedia,
                         showThumbnailPreviewOnRight: state.showThumbnailPreviewOnRight,
+                        showTextPostIndicator: state.showTextPostIndicator,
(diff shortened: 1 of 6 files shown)


machinaeZER0:
Yay!

machinaeZER0:
It's been gone so long I've forgotten (lol) - did tapping the indicator work like other posts, where it let you view content without marking it as read?

micahmo:
> did tapping the indicator work like other posts, where it let you view content without marking it as read?

No, I'm afraid not. It just opens the full post. Since there's no media or link to open, it just opens the full post, which always mark it as read (current). Is this something you'd want?

machinaeZER0:
Just curious! Couldn't remember the original functionality :) I could see a case being made for it but I don't feel strongly about it

hjiangsu:
I have merged in #384 so you can adjust the border radius value and we should be good to go!

micahmo:
All set, thank you!

PR #386, opened 2023-07-17
Show posts from all instances by default

micahmo:
As discussed in Matrix chat, we agreed to set the default feed type to `All` in order to surface posts from across the Lemmyverse. 😊


hjiangsu:
Nice 👍

Issue #361, opened 2023-07-14
Tapping comment from profile should show comment in limited context


SquawkWoof:
**Is your feature request related to a problem? Please describe.**

Currently, tapping a comment from your profile (comment history) loads the entire comment thread for the associated post.  This makes it cumbersome to locate the comment and view replies.

**Describe the solution you'd like**

Tapping a comment from your profile (comment history) should display the comment thread for the associated post in limited context: the comment itself, all replies (children and grandchildren), and the immediate parent comment (if applicable).

**Describe alternatives you've considered**

Automatically scrolling to the comment within the full thread could be a reasonable alternative.

**Additional context**

n/a

micahmo:
Related to #329.

KarkanAlzwayed:
@hjiangsu this also applies to clicking on a reply to my comment in the inbox. Clicking the reply doesn't bring it up, it just loads the whole thread, and I'd have to sift through the whole thread to find it. Didn't want to open a new issue since I think the two issues are similar

hjiangsu:
Hey! A fix for this has already been applied, but has not yet been released as a pre-release version!

KarkanAlzwayed:
> Hey! A fix for this has already been applied, but has not yet been released as a pre-release version!

Is the fix supposed to be in the alpha 1+13? If so, it still not bringing up the comment. It doesn't bring up the thread that comment is in, but not the comment itself. I still have to go find the comment, but it's easier to find now.

hjiangsu:
If I recall correctly, it will load the specific comment thread where your comment resides in, but it does not yet scroll your comment into view!

The fix for the inbox not showing the comment is not in alpha 1+13. It'll most likely come with the next pre-release version alpha 2+13

KarkanAlzwayed:
> If I recall correctly, it will load the specific comment thread where your comment resides in, but it does not yet scroll your comment into view!
> 
> The fix for the inbox not showing the comment is not in alpha 1+13. It'll most likely come with the next pre-release version alpha 2+13

My man, thank you so much for the hard work and the quick replies. Brining the comment itself automatically would be legendary. Let me know if you need any testing done. Again, thank you.

Issue #329, opened 2023-07-11
Selecting comment in inbox shows full comment section


TheSaneWriter:
**Description**  
When I select a comment in my inbox, it takes me to the full comment section, forcing me to search for the original comment.

**How to Reproduce**  
1. Go to your user inbox
2. Select a comment reply
3. Full comment section of that post opens.

**Expected Behavior**  
When I select a comment in my inbox, typically in other apps it will open a focused view of the post, the selected comment, and the replies to that comment, without the other post comments.

**Device & App Version:**  
 - Device: OnePlus 9 5G
 - OS: Android 13
 - Version: v0.2.1+11

CTalvio:
There's been some discussion on implementing jump to comment and context view. It's still in the phase of figuring out how to do it in a good way.

Definitely something that is planned to be addressed.

PR #396, opened 2023-07-17
Initial support for viewing comment context.

ajsosa:
https://github.com/thunder-app/thunder/issues/361
https://github.com/thunder-app/thunder/issues/329

This is going to need some scrutiny. Don't like the increase in data being passed around but not sure how to improve. So please tear it apart with some good feedback.

I've basically added the ability jump to a comment's context when a comment is clicked from a user's comment's list. I also added a button that is clickable to then load in the full comments list.

To get the full comment load to work with the fancy animation, I had to rearrange `comment_view` a bit to get the animation to work smoothly and keep both the end of comments message and the load full comments button together with the comments themselves.

Future improvements would be auto scroll to highlighted comment and also maybe add a spinning wheel animation while you wait for the full comment list to load in from the bottom.

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -113,9 +113,14 @@ class PostBloc extends Bloc<PostEvent, PostState> {
 
           CommentSortType sortType = event.sortType ?? (state.sortType ?? defaultSortType);
 
+          int? parentId ;
+          if (event.selectedCommentPath != null) {
+            parentId = int.parse(event.selectedCommentPath!.split('.')[1]);
+          }
+
           List<CommentView> getCommentsResponse = await lemmy
               .run(GetComments(
-            page: 1,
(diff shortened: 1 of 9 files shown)


ajsosa:
Found a bug. If the selected comment is very deep, it will be collapsed. Expanding it will remove the load full comments button.

hjiangsu:
Just giving an update - I'm looking at this currently to just double check everything! Don't worry about the conflicts, I've already fixed them on my end

hjiangsu:
I think overall, it looks pretty good. I'm also not a fan of passing around the data so much, but I think this would be fixed by refactoring PostBloc into post and comments rather than having to handle both (which is another task by itself). For now though, I think its fine to leave it as it is!

I'll merge it in for now, and when we refactor some of this overall PostBloc code, we can revisit how to make it better

PR #397, opened 2023-07-17
Fixed perpetual spinning icon for comment loading

ajsosa:
Fixed a bug where we would perpetually show spinning icon indicating we were loading new comments when infact we had reached the end. Lemmy API may send the same comments for different pages so we keep track of what comments we have with a map to have an accurate count of comments so we know when we've reached the end.


hjiangsu:
Nice catch there - I noticed it too but you beat me to it 😅

PR #399, opened 2023-07-18
Mark read/unread swipe gesture

micahmo:
This PR adds a new post swipe gesture for marking posts as read/unread.

### Notes
* I am totally open to a different color for the gesture.

### Demo

https://github.com/thunder-app/thunder/assets/7417301/39c60bae-22e4-4372-8326-323e694b64d0

### DIFF
--- a/lib/core/enums/swipe_action.dart
+++ b/lib/core/enums/swipe_action.dart
@@ -1,1 +1,15 @@
-enum SwipeAction { upvote, downvote, reply, save, edit, none }
+enum SwipeAction {
+  upvote(friendlyName: 'Upvote'),
+  downvote(friendlyName: 'Downvote'),
+  reply(friendlyName: 'Reply'),
+  save(friendlyName: 'Save'),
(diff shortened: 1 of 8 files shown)


lib/core/enums/swipe_action.dart
```diff
@@ -1 +1,15 @@
-enum SwipeAction { upvote, downvote, reply, save, edit, none }
+enum SwipeAction {
+  upvote(friendlyName: 'Upvote'),
```
hjiangsu, line 2: I would change this from `friendlyName` to `label` to make it more clear!

hjiangsu:
I think it looks pretty good! Just one comment about the variable name for the enum, but otherwise good!

I do think we should refactor the common swipe icons/colours though as it'll get confusing later on :D

lib/core/enums/swipe_action.dart
micahmo: Thanks, that's much better!

PR #400, opened 2023-07-18
Improve block community toast

### LINKED ISSUE #285: Community title + info is cut off in Blocked popup message
Screenshot attached - for some/all of the communities which use additional descriptive text beyond their community name, it looks like that info is being pulled when you use the block functionality:
Screenshot_20230708-140548

Currently the popup does not account for text this long, which continues to draw off the screen instead of wrapping. I propose that (since we presumably know what we're blocking already) the popup would only show the community name, and not the community name + description text.

This feeeeeeeeeeels like it might be something I could try and make a PR for, but (as I've stated once or twice) I'm not a coder and I've never done a PR. That said, if someone wanted to point me in the right direction as to where I could look in the code, I could try and implement?

micahmo:
Fixes #285. In addition to truncating text, I also tweaked the styling of the snackbar.

### Demos

https://github.com/thunder-app/thunder/assets/7417301/3b576263-6710-4a7e-85a1-d3db6dce679f

https://github.com/thunder-app/thunder/assets/7417301/58e2dee9-6ba0-4015-9686-88e9a226b05a

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -73,19 +73,27 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
               content: Row(
                 mainAxisAlignment: MainAxisAlignment.spaceBetween,
                 children: [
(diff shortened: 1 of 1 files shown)


hjiangsu:
Just rebased it to `develop`, but otherwise it looks good!

PR #401, opened 2023-07-18
Minor optimization: Minimized DateTime.now() calls and comment time calcs moved to post_bloc

ajsosa:
Minor optimization. We call DateTime.now() just once and pass it to comment cards rather than call it once for every card. Also we now compute the time elapsed since comment posted/edited in post_bloc when we initially fetch comments to avoid doing those calculations in a Build.

### DIFF
--- a/lib/core/models/comment_view_tree.dart
+++ b/lib/core/models/comment_view_tree.dart
@@ -10,7 +10,10 @@ class CommentViewTree {
   /// The depth of the comment. It starts from 0, which is a direct reply to the post
   int level;
 
+  String datePostedOrEdited;
+
   CommentViewTree({
+    this.datePostedOrEdited = "",
     this.commentView,
     this.replies = const [],
     this.level = 0,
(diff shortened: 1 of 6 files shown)


hjiangsu:
Hmm, one suggestion for this - instead of just having minutes ` int sinceCreated = widget.now.difference(widget.commentViewTree.commentView!.comment.published).inMinutes;`

I have a helper function which helps determine the time difference in `date_time.dart`. You can check out `formatTimeToString` function which is used for determining the post card times! This might be better since it accounts for minutes, hours, days, weeks, etc.

ajsosa:
So `int sinceCreated` was something I preserved from the original logic. It's only used for determining if a post is new by checking if it's less than 15 minutes old. The actual post time is still being formatted by `formatTimeToString` in `comment.dart` when we build the tree view.

I went ahead and refactored that `sinceCreated` bit by renaming and doing the new calculation right then and there to make it clearer what the value is actually for.

hjiangsu:
Ahh I see - in that case, we can leave it then!

Issue #404, opened 2023-07-18
Scrolling is laggy in comment threads when one or more comments contain embedded images


SquawkWoof:
**Description**  

When viewing through a comment thread with one or more comments contain embedded images, scrolling becomes very laggy.

**How to Reproduce**  

1. Open a thread with one or more embedded images 
2. Scrolling will be laggy for the entire thread

**Expected Behavior**  

Scrolling to be smooth

**Screenshots**  

n/a

**Device & App Version:**  

Lenovo Tab 8 HD, Android 10, Lightning v0.2.1+12

**Additional Context**  

n/a

hjiangsu:
Do you have a link to a post thread which can be used to reproduce the issue? That would help with profiling and making sure that fixes are applied properly!

SquawkWoof:
@hjiangsu: sure, I've added a link in the original post

CTalvio:
Is this different from #395?

ajsosa:
https://github.com/thunder-app/thunder/pull/408 Should definitely help with overall performance and smoothness of the comment section. But there is still an issue with lots of images/big images causing the comment section to teleport.

hjiangsu:
@SquawkWoof Could you check and see if you're still encountering this issue with the latest version? Thanks!

PR #408, opened 2023-07-18
Performance bug fix for comment view rebuilding on every scroll event.

ajsosa:
The scroll back to top button logic depended on a `setState` inside of a scroll event listener. This was causing that `setState` to trigger constantly as the user scrolls through a post. This in turn marks the current widget and it's children as `dirty` which triggers a complete rebuild.

Added some if statements to prevent `setState` from being continuously called and instead only called when needed.

Removed unused auth related stuff.

### DIFF
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -67,9 +67,17 @@ class _PostCardListState extends State<PostCardList> {
     if (_scrollController.position.pixels >= _scrollController.position.maxScrollExtent * 0.7) {
       widget.onScrollEndReached();
     }
-    setState(() {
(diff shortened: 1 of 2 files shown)


ajsosa:
Just made the same fix for the main post view.

micahmo:
Nice! I've been noticing some laginess in scrolling lately. Hopefully this helps.

It wouldn't help #404 would it?

hjiangsu:
Oof, thanks for finding that out! Hopefully, it'll improve the situation a bit.

> It wouldn't help https://github.com/thunder-app/thunder/issues/404 would it?

I think this is most likely a different issue overall - it might be something to do with the images not being offloaded from memory, so a closer look will need to be done here

Issue #409, opened 2023-07-19
2 column mode bug: swipe gesture voting has misaligned icons.


ajsosa:
#### Description
When using 2 column tablet mode, the swiping gesture on posts has the vote and reply icons misaligned. 

#### Examples
![Untitled](https://github.com/thunder-app/thunder/assets/41923324/f8bd1e69-0dea-4eea-9def-f1b4b3b6bbc1)
![Untitled2](https://github.com/thunder-app/thunder/assets/41923324/aca4bb0e-b3ab-4522-a802-0e9a74672dcd)

PR #410, opened 2023-07-19
#409: Fixed placement of vote and reply button during swipe action on post.

ajsosa:
https://github.com/thunder-app/thunder/issues/409

There is still a separate visual issue where you can swipe the card over the other in 2 column mode.


hjiangsu:
Looks good!

PR #411, opened 2023-07-19
Login keyboard improvements

micahmo:
This is a tiny tweak which improves the keyboard behavior when navigating through the login fields. The user should be able to navigate and initiate the login directly from the keyboard. Note that I didn't have "Password" navigate to "TOTP". I figured if you have to fill in that field you'll be switching apps anyway.

### Before

https://github.com/thunder-app/thunder/assets/7417301/78e49cdf-8ce8-451e-92f4-3df53618f6c1

### After

https://github.com/thunder-app/thunder/assets/7417301/9e55d25b-8460-48a3-afab-ad66af6fe267


hjiangsu:
Looks good!

Issue #343, opened 2023-07-12
Incorrectly assuming all links with /c/ are Lemmy links


ajsosa:
#### Issue
Clicking on links with paths that start with `/c/rest-of-path` get incorrectly treated as a Lemmy link. 

#### Example
https://catbox.moe/c/rk6abk

So when clicked within app, Thunder tries to open it as if it were a community.

#### Reference
https://lemmy.world/post/1369784

hjiangsu:
I believe I need some stronger regex rules for figuring out lemmy links - linking this since it seems like we can follow lemmy-ui's regex: https://github.com/LemmyNet/lemmy-ui/blob/61255bf01a8d2acdbb77229838002bf8067ada70/src/shared/config.ts#L38

micahmo:
> it seems like we can follow lemmy-ui's regex: https://github.com/LemmyNet/lemmy-ui/blob/021677316d290062a8c3bafdbec668dd8ebb3b85/src/shared/config.ts#L31-L39

Unfortunately that regex is pretty limited in what it matches. It requires the full `community@instance` name.

![image](https://github.com/thunder-app/thunder/assets/7417301/5f067e03-451f-4d7e-83ea-924c484ca9eb)

ajsosa:
Yeah that's probably the simplest solution. Do a ping to check if the community exists and have that decide how to open link.

micahmo:
This fix should be in the latest general release.

PR #416, opened 2023-07-19
Improve community navigation

micahmo:
This PR includes two improvements for community navigation.
1. Ensure that the `communityId` is populated, which allows us to perform account-specific actions.
2. Ensure that communities with `@` are handled properly (i.e., not as emails).

I used this post as an example with all the varieties I could think of.

https://lemmy.fmhy.ml/post/1169884

### Demo

https://github.com/thunder-app/thunder/assets/7417301/6541106e-5946-44fa-9d4c-d68b06a09a9f

**Note** that this does not affect #343, which still needs to be addressed separately (i.e., how to detect whether a `/c/` link is actually to a Lemmy community).


CTalvio:
Were you able to fix exclamation marks not being detected as part of the link, when not explicitly formatted as one? It's really only a cosmetic issue, but it's annoying the hell out of me.

micahmo:
> Were you able to fix exclamation marks not being detected as part of the link, when not explicitly formatted as one?

No, not as part of these changes. In fact, the only reason `@` are recognized as a link by the markdown parser is that they look like email addresses. In fact, part of this fix is to remove the `mailto:` that's implicitly prepended. In order to include the `!`, we would need to do our own parsing.

hjiangsu:
Looks good!

PR #417, opened 2023-07-19
Improve link previews more

micahmo:
This PR includes a few improvements to the link preview.
1. Use `thumbnailUrl` (as generated by Lemmy backend) from the API before resorting to `LinkPreviewGenerator`.
2. Set `LinkPreviewGenerator` cache to 0 so that it doesn't consume as much memory.
3. Set the loading indicator for `LinkPreviewGenerator` so it doesn't just pop in.

@ajsosa Since you found the original memory issue, maybe you can test these changes.


hjiangsu:
We can take this one step further if you'd like for efficiency purposes. When we parse the post, we can check if there is an existing `postView.post.thumbnailUrl`. If there is, we can completely skip `LinkInfo linkInfo = await getLinkInfo(url);`

Then we can adjust this line `media.add(Media(mediaUrl: linkInfo.imageURL!, mediaType: MediaType.link, originalUrl: url));` so that it picks between the thumbnail or the image retrieved from the url

https://github.com/thunder-app/thunder/blob/1a0da4bd41242c20db4d111b5d836320d5824532/lib/utils/post.dart#L129-L150

micahmo:
> When we parse the post, we can check if there is an existing `postView.post.thumbnailUrl`. If there is, we can completely skip `LinkInfo linkInfo = await getLinkInfo(url);`

It looks like `getLinkInfo` is only called if we're showing full height images. And in that case, don't we want the original image, not the thumbnail?

hjiangsu:
I think we can continue to use the thumbnail image, because the thumbnail would have the same ratio as the original image. I think the thumbnails are also high-quality enough that we don't have to worry about pixelation (you can double check that if you want!)

Just skipping that step `LinkInfo linkInfo = await getLinkInfo(url);` saves us from having to make more network calls than needed - I also think thumbnail images are usually not as important as the actual images since its mostly there for visual purposes (e.g., a user can't save or preview the thumbnail images)

micahmo:
Ok I think I get what you're asking, but let me just check before I commit. Something like this?

image

hjiangsu:
Yup, that's pretty much it! I would also add that into the `else` clause near the bottom (2nd line from the bottom) too so that it can also try to fallback to using the thumbnail

micahmo:
All set!

Issue #345, opened 2023-07-12
[Development] Standardize linting and auto-formatting of project files


hjiangsu:
**Is your feature request related to a problem? Please describe.**
This is for development practices and workflow enhancements. It would be best if we are able to standardize the auto-formatting and linting processes for the Thunder. This aims to help with a few things:
- Easier time to look over the actual changes when reviewing PRs as we can focus on the logical changes, rather than changes that occur due to auto-formatting that VSCode provides
- Linting will allow us to keep the codebase a bit cleaner, and helps remove unnecessary imports and changes

**Describe the solution you'd like**
- To standardize linting, we could have a pre-commit hook which runs `dart fix` to automatically apply fixes to any dart files
- Standardizing auto-formatting will need a more robust solution. For those using VSCode, we could create a workspace settings file which would apply to everyone using VSCode. These workspace settings should be agreed upon together.
  - Other solutions may be required for those who are using another IDE to develop in

The main workspace settings that I feel like are applicable are the following rules (these are my own relevant settings):
```json
"dart.lineLength": 200,
"[dart]": {
    "editor.formatOnSave": true,
    "editor.rulers": [ 200 ],
},
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

hjiangsu:
Pulling in @micahmo @ajsosa @bactaholic @CTalvio for suggestions and discussion - Feel free to pull in others which might benefit from this discussion as well

micahmo:
I am definitely in favor of this, and the starting settings you supplied seem to work great (by default, the `lineLength` is far too short, in my opinion). Do any of the other recommended settings here look useful?

https://dartcode.org/docs/recommended-settings/

The only other thing that comes to mind for standardized formatting is `.editorconfig`. The benefit is that a lot of IDEs support it, so we wouldn't be tied to VS Code. However, I'm not sure if Dart supports it.

hjiangsu:
I actually have all of those settings that you linked! But I think for this particular purpose, the settings I mentioned above is the minimum needed and the rest are based on your personal preference

> The only other thing that comes to mind for standardized formatting is .editorconfig. The benefit is that a lot of IDEs support it, so we wouldn't be tied to VS Code. However, I'm not sure if Dart supports it.

I think formatting on save is something that happens on a per IDE basis, so I'm not sure if `.editorconfig` will work in this case (although it might help with setting the indentation spacing)

micahmo:
[...]

PR #418, opened 2023-07-19
Save images to its own folder

njshockey:
This might be a me problem but saving downloaded images to the root Pictures folder is kind of annoying. It's bound to get clogged up eventually. So I changed it to save the pictures to [internal_storage]/Pictures/ThunderImages. To do this, I needed to use gallery_saver instead of image_gallery_saver. image_gallery_saver doesn't support albums and doesn't have any way to get the path that the image is being saved to, so changing the dependency was the only way to do that. The function that's called when the download button is clicked is the only place where image_gallery_saver is used as far as I can tell (it doesn't give any new errors anywhere else). To support the new dependency, I also had to change the minimum Android sdk version to 21, which is 5.0 Lollipop. This version of Android is nearly 9 years old so I doubt it'll be an issue.

But I should note that I have no way of testing this on iOS. It works on Android in both emulator and on my phone (Galaxy S8). I had to do a slight hack and make the albumName "Pictures/ThunderImages" since otherwise it just makes a new folder in the root internal storage. The path may need to be changed on iOS as well but I have no way of testing it so I have no idea if its necessary.

This is unrelated to this PR but I also noticed that there's a lot of unused variables and unused imports in the code base, so someone might want to use `dart fix` to try and clean those up.


ajsosa:
Currently we actually do all of our merges into the develop branch which then gets merged into main for a major release. Could you please change the PR for merging into the develop when you get the chance? Thanks!

njshockey:
How would I do that?

njshockey:
Never mind, I figured it out. Resolving the conflicts now.

njshockey:
I resolved the merge conflicts. Let me just double check that this still works real quick.

njshockey:
Alright, this should be good to go now. Thanks for clarifying that!

hjiangsu:
Hey! Thanks for the contribution - I'll check out the code when I have time later.

> But I should note that I have no way of testing this on iOS.

I can help test on iOS

> This is unrelated to this PR but I also noticed that there's a lot of unused variables and unused imports in the code base, so someone might want to use dart fix to try and clean those up.

Yup, this is known right now, and I'm trying to figure out a better solution to handle general workflow and development stuff here: #345 If you have any suggestions, feel free to drop it in that discussion!

CTalvio:
I'd really like this to be a setting, eventually. Personally I'd vastly prefer a folder named "Thunder" but that is generic enough that it might overlap with an actual album name someone has.

This is an improvement, but a save location setting should be the final goal.

An additional feature that I'm sure some would appreciate, would be that of the app saving images to folders named according to to the community they were posted in.

njshockey:
> I'd really like this to be a setting, eventually.

I was thinking the same thing but wasn't sure if people would even care lol.

> Personally I'd vastly prefer a folder named "Thunder" but that is generic enough that it might overlap with an actual album name someone has.

I just decided to be consistent with how it saves to pictures on desktop. The code for desktop is right below the code that saves on mobile.

hjiangsu:
@njshockey When you have the chance, could you run `dart format . -l 200` to make sure that the formatting stays consistent with the project?

Or alternatively, if you're using vscode, you can pull the changes from the latest `develop` branch which contains the vscode workspace settings for auto-formatting!

njshockey:
Yeah, I can probably do that in a few hours.

hjiangsu:
Hmm, just took a look at the code, and I think there's some code thats missing here which was handling tap-zoom gesture that was added previously.

I'm thinking this could be because the branch you were working on was on the `main` branch `njshockey:main` rather than `develop`. I'll see if I can adjust it so that it includes the new changes made to the image viewer!

hjiangsu:
Alright, I've updated the code to contain all the changes from the most recent `develop` changes here!

I also adjusted the name of the folder to `Thunder` instead of `ThunderImages`. If you could just double check the changes I've made, and run some tests to make sure it still all works, that would be great!

Once you confirm that everything still seems to be working as intended, I'll merge this in

njshockey:
Everything worked for me on Android. I could double tap to zoom in and it saved to the right place. (Again, I can't test iOS.)

hjiangsu:
Alright sounds good! I briefly did a test on iOS and it seemed to work!

bikasv:
I know it's a closed ticket, but what I've on Android is that the images are not saved in **sdcard > Pictures > Thunder** folder but rather **sdcard > Pictures > Pictures > Thunder** folder.

Can this be addressed?

CTalvio:
> I know it's a closed ticket, but what I've on Android is that the images are not saved in **sdcard > Pictures > Thunder** folder but rather **sdcard > Pictures > Pictures > Thunder** folder.
> 
> Can this be addressed?

Oh dear, that's definitely a bug, feel free to open up another issue for it.

Issue #368, opened 2023-07-15
Add option to block users from comment long tap


CTalvio:
There is currently no way to block users. The option to do this should be available in both post and comment context menus, as well as profile pages.

CTalvio:
Will come with with #419

Will still need to be added to long tap menus.

Issue #289, opened 2023-07-09
Community (and user?) sidebars


CTalvio:
Communities have sidebars. These often contain information on what a community is for, as well as rules on what you are allowed to post. Sidebars also inherit the content of the parent instance sidebar, which should be accessible below the sidebar content of the current community, which ensures users see the rules of an instance when posting to communities outside their own instance.

I think community sidebars should be accessible in two places:

- On the community page: by tapping the header, a button at the top or a slide gesture on the header

![Screenshot_20230708_152903](https://github.com/hjiangsu/thunder/assets/4365015/746f2481-4cdd-4b38-910f-308dd3f35e10)

My admittedly minimalist preference for hinting to a user that this is how it's done, is a mere arrow pointing to the left. Users will be familiar with sidebars being on the right, and coming in towards the left. Tapping or pulling anywhere on the header should reveal the sidebar.

I would suggest keeping the header, and sliding in the sidebar over the posts list, so as to keep the header present and avoid having to display duplicate information in the sidebar. This will also keep the subscribe button in the top bar, available. The arrow should remain, now pointing to the right, representing the same action closing the bar.

- In the editor when creating a post: when making a post, there should a hit text telling users to make sure they are following the rules of the community they are posting in, as well as a way to check what those rules are without leaving the editor (ie, the ability to open the sidebar in the editor)

I think something similar could be used for profile pages to reveal a user's bio, join date, and account age.

PR #419, opened 2023-07-19
Sidebars

CTalvio:
Adds sidebars accessible on community pages and user profiles.

- #368 
Adds the ability to block and unblock a user on their profile page.

Known problems:
- Currently does not check if a user is already blocked when opening their sidebar.
- Local instance information is not available.

Fixing these will have to be done by someone else, or me, later. I'm done with this for now, but I've gotten it to a very usable state.


CTalvio:
#289

machinaeZER0:
I've been thinking about the sidebar some more and I have a thought/request - would it be possible to change the arrow icon that opens the sidebar to an information icon? The old lowercase i within a circle. 

My thinking here is that while the arrow indicates that something will happen when you tap it, it doesn't inform them as to what specifically will happen. If we use an information icon instead, people will tap on it expecting more information and then be pleasantly surprised by the beautiful sidebar, is my thinking.

If we made this change I don't think we'd necessarily need an arrow icon to close the sidebar either - we don't have one to close the left sidebar, after all. When people are done sidebar-ing they can just swipe the bar away :)

cc @hjiangsu @micahmo and others in case people have further thoughts on this

micahmo:
I had originally suggested [menu_open_rounded](https://api.flutter.dev/flutter/material/Icons/menu_open_rounded-constant.html) which I liked because it indicated both an action (with the arrow) and the result (that it would show a menu).

We could definitely use an info icon though, like [info_outline_rounded](https://api.flutter.dev/flutter/material/Icons/info_outline_rounded-constant.html).

CTalvio:
I ended up not using the icon you found, I also liked it better, but at a size where it looked right, it was too small, and a size where it was big enough to be noticeable, it looked very "bloated".

Replacing the arrow, or adding the info icon, could work.

CTalvio:
Ready to merge! @hjiangsu 

image

I can swap that icon in again if we end up preferring.

Issue #366, opened 2023-07-14
Reduce size of images in comments


CTalvio:
Comments with images currently take a very large amount of space, it would be good to reduce the size of images in comments 30-40% for a more compact overview.

Introducing a maximum allowed height for very tall aspect ratios could also be good.

PR #420, opened 2023-07-20
Tweaks to comment images and buttons

CTalvio:
Limits vertical and horizontal size of images in comments. The padding, height, and width of the comment actions have also been tweaked. The lack of top padding can look quite goofy.

Button height reduced, while width is increased, maintaining a reasonable tap target size while taking up ever so slightly less vertical space.

Screenshot_20230720_032549

Above is current appearance, below is the tweaks this PR applies.

Screenshot_20230720_032520


CTalvio:
#366

machinaeZER0:
This is a huge improvement! Defer to you, but should we indicate in some way that an image is cropped/tapping it will show the full image? Do we currently do that anywhere else in a way we could replicate here?

micahmo:
> should we indicate in some way that an image is cropped/tapping it will show the full image?

I don't know what this would look like UI/UX-wise, but I agree. This get me in Telegram sometimes (where it's unpredictable whether images are cropped or not). You might just think that's the whole thing.

machinaeZER0:
Maybe some sort of linear gradient overlay coming up from the bottom? Probably wouldn't have to be too dark. I think the rounded corners already present do help as well with suggesting that this is a thumbnail, so this would just be a little extra touch to make it (hopefully) obvious

CTalvio:
I wont be adding any further thumbnailing is this PR, moving onto FABs. @hjiangsu feel free to merge

hjiangsu:
Looks good!

CTalvio:
Realised a bonus feature. Comments that are just an image, have almost no area to tap to collapse them.

This fixes that.

PR #421, opened 2023-07-20
Settings tweak

CTalvio:
Adds the same collapse and hide animation to gesture settings as compact/normal switch.
Adds tips on where to find relevant settings for users looking to disable gestures and use buttons instead.

image


ajsosa:
Um was this mean for develop branch?

CTalvio:
> Um was this meant for develop branch?

Yes. WHY DOES IT KEEP DEFAULTING TO MAIN. Even when i branch of dev.

micahmo:
> WHY DOES IT KEEP DEFAULTING TO MAIN.

Because that's the repo's default branch. 😊

image

hjiangsu:
Looks good to me!

PR #422, opened 2023-07-20
Increase relevance of default community search

machinaeZER0:
Hoping this works... with some help from @micahmo and @Benjamint22 I have hopefully updated the community search results to default to Top (Year) by creating a new default search sort const. The current default (Hot) tends to lead with communities that feel less relevant and have way fewer people subbed, so this feels like it will be an improvement for people diving into lemmy for the first time. Best part - we already have sort options in, so users are free to change it to suit their needs!

Please test if you can and let me know if I screwed anything up - this is my second-ever PR and I have much to learn in the ways of coding!


micahmo (approved):
Looks perfect!!

hjiangsu:
WOOO nice job! It looks good to me!

hjiangsu (approved):

PR #423, opened 2023-07-20
Improve nested comment indicator

micahmo:
This PR aims to improve the nested comment experience by default while also introducing some customizability.

There are now two settings with two options each to customize the nested comment indicator. Here is a video demonstrating just the settings.

https://github.com/thunder-app/thunder/assets/7417301/2739fd8d-003d-490b-8e70-2590462fb925

Here are screenshots of every possible combination of settings in both light and dark mode.

|                                                                                                           |                                                                                                           |                                                                                                           |                                                                                                           |
|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
| SPeAOxaRfL | 8GQ0HQJP2g | Cn44jarMgq | QrQ2feiodt |
| DE5Hzlvra2 | 2tRp5g4pSY | GJ0OImbjrM | LNP067hmJp |

### DIFF
--- a/lib/post/widgets/comment_card.dart
+++ b/lib/post/widgets/comment_card.dart
@@ -128,273 +128,302 @@ class _CommentCardState extends State<CommentCard> with SingleTickerProviderStat
     final ThunderState state = context.read<ThunderBloc>().state;
 
     bool collapseParentCommentOnGesture = state.collapseParentCommentOnGesture;
+    NestedCommentIndicatorStyle nestedCommentIndicatorStyle = state.nestedCommentIndicatorStyle;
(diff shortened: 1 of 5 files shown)


hjiangsu:
Looks good! Although now that I'm looking more closely at the code, one thing that maybe we can start thinking about in the future is to see if we can split the long widgets into shorter more condensed sub-widgets. Theres 2 benefits to this:

- It can improve flutter performance because of the way flutter handles rebuilding widgets
- It improves readability, especially for code reviews where wrapping a widget causes a git to mark everything within that wrapped widget as changed

I'll merge this in though as these issues are a separate issue!

PR #426, opened 2023-07-20
About page improvements

micahmo:
This PR includes some small fixes for the About page.

1. Update the GitHub link to reflect the new org.
2. Update the community link to open in-app.
   * I refactored some code so we have a generic way to navigate to a community by name.

### DIFF
--- a/lib/settings/pages/about_settings_page.dart
+++ b/lib/settings/pages/about_settings_page.dart
@@ -5,6 +5,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:thunder/utils/links.dart';
 import 'package:thunder/core/update/check_github_update.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+import 'package:thunder/utils/navigate_community.dart';
 
 class AboutSettingsPage extends StatelessWidget {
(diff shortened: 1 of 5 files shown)


lib/settings/pages/about_settings_page.dart
```diff
@@ -45,10 +46,10 @@ class AboutSettingsPage extends StatelessWidget {
                     'GitHub',
                     style: theme.textTheme.titleMedium?.copyWith(fontWeight: FontWeight.w600),
                   ),
-                  subtitle: const Text('github.com/hjiangsu/thunder'),
+                  subtitle: const Text('github.com/thunder-app/thunder'),
```
micahmo, line 49: @hjiangsu If you intentionally wanted to keep a reference to your name/username in the app (which you deserve), I would put something like "developed by hjiangsu" beneath the app title on the about page, rather than using the old GitHub link.

hjiangsu: I think we can change it to the org name, it makes more sense to do that.

If anything I was thinking we could just have a separate section where we can attribute main contributors of the project (at this point, this would include you, @ajsosa, @CTalvio, @machinaeZER0  and maybe others that have helped with this project!)

PR #435, opened 2023-07-20
Haptic feedback for comment long press

ajsosa:
Added haptic feedback when long pressing on a comment to bring up Actions drawer.


CTalvio:
They still lack the inkwell effect?

Issue #407, opened 2023-07-18
Ability to view and manage blocked communities


machinaeZER0:
Saw this request on the lemmy board so logging here - now that we're able to block communities in the app, it would be cool to be able to view and manage this in settings! Would be especially useful in the event that you accidentally block a community you want to un-block. Already possible to do this on the web client, but an in-app solution would be handy in the long term!

PR #437, opened 2023-07-21
Accounts settings for blocked entities

micahmo:
For #407

This PR introduces an account settings page. To start, the page contains the user's currently blocked people and communities. They may be unblocked by being removed from the list. 

### Notes
* This is my first time doing a brand new page and an end-to-end bloc, so hopefully things make sense.
* There is some crossover with the `CommunityBloc` and `UserBloc`, but the new events are specifically within the context of account settings, so I think it's ok for them to be separate.
* Load animations, progress indicators, error handling, undo, etc., should all be supported.

### Demo

https://github.com/thunder-app/thunder/assets/7417301/4e62858c-ce01-4cd5-b163-8f849a8c1e18

### DIFF
--- a/lib/user/bloc/user_settings_bloc.dart
+++ b/lib/user/bloc/user_settings_bloc.dart
@@ -1,0 +1,121 @@
+import 'package:bloc/bloc.dart';
+import 'package:bloc_concurrency/bloc_concurrency.dart';
+import 'package:equatable/equatable.dart';
+import 'package:lemmy_api_client/v3.dart';
+import 'package:stream_transform/stream_transform.dart';
+import 'package:thunder/account/models/account.dart';
+import 'package:thunder/core/auth/helpers/fetch_account.dart';
+import 'package:thunder/core/singletons/lemmy_client.dart';
+
+part 'user_settings_event.dart';
+part 'user_settings_state.dart';
+
+const throttleDuration = Duration(seconds: 1);
+const timeout = Duration(seconds: 3);
+
+EventTransformer<E> throttleDroppable<E>(Duration duration) {
+  return (events, mapper) => droppable<E>().call(events.throttle(duration), mapper);
+}
+
+class UserSettingsBloc extends Bloc<UserSettingsEvent, UserSettingsState> {
+  UserSettingsBloc() : super(const UserSettingsState()) {
+    on<GetUserBlocksEvent>(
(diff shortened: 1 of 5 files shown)


hjiangsu:
Just took a look at this, and I think it looks pretty good! This user settings page can also be the start of syncing up the web ui settings with Thunder as well, so I like the overall naming for the bloc

One thing that we might be able to improve on in the future is maybe abstracting the logic of showing the toast notifications. We're starting to use that a lot in different areas, so it makes sense to abstract it.

For example, instead of having to write:
```dart
SnackBar snackBar = const SnackBar(
  content: Text('Succesfully blocked!'),
  behavior: SnackBarBehavior.floating,
);

WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
  ScaffoldMessenger.of(context).clearSnackBars();
  ScaffoldMessenger.of(context).showSnackBar(snackBar);
});
```

We might just be able to abstract it into its own function and pass in the body:
```dart
Widget toastBody = Text('Succesfully blocked!');

showToastNotification(context: context, body: toastBody);
```

We don't have to do it in this PR, but just something to think about

micahmo:
Regarding toasts, I agree. I think we have at least three variations:

1. Show toast
2. Clear existing toasts and show (maybe we should always do this instead of number 1)
3. Show toast with action (this can take the icon and action as parameters)

hjiangsu:
Yeah, I think we'll keep it to 2 and 3. In this case, we could have a call to show toast notifications as

```dart
Widget toastBody = Text('Successfully blocked!');

function onAction() {
  // Do something
}

Icon actionIcon = Icon(Icons.abc);

showToastNotification(context: context, body: toastBody, actionIcon: actionIcon,  action: () => { onAction() });
```

Also, I just noticed a typo: should be Successfully instead of Succesfully!

micahmo:
> action: () => { onAction() }

I'm assuming functions are objects in Dart, so could we save an invocation and just do `action: onAction`?

> Successfully instead of Succesfully

Fixed!

hjiangsu:
> I'm assuming functions are objects in Dart, so could we save an invocation and just do action: onAction?

Possibly? I'm not too sure myself without doing some experimentation. I know in JS that you can do that though

hjiangsu:
Anyways, I think I'll merge this in for now, and we can think about the refactoring for a future issue?

Issue #431, opened 2023-07-20
Thumbnail badges to clarify post type


machinaeZER0:
I don't know if this would be for both compact and normal modes, but for compact in particular I think it would be great to add a badge that shows the type of content that's been posted (text/image/link/video). Once we figure out the best way to pull thumbnails for external links, this will be particularly good for showing that something is a link without overly obscuring the thumbnail. 

I was inspired by a screenshot from the lemmy app Beans today:
![1689854393309.png](https://github.com/thunder-app/thunder/assets/3624023/6b60fb90-1a43-4d6d-8af0-50a4dcaf32e5)

I think we could do something similar and have it look really gorgeous in Thunder!

micahmo:
Relay had something similar with little color tabs to indicate the media type (no tab = link, gray = photo, green = gallery, blue = gif). On the one hand, colors are less discoverable. On the other, once you know what they mean, they keep the UI pretty clean.

Right now, we assume any thumbnail is an image, unless it has the link overlay. But I definitely like the idea of trying to streamline that a bit.

CTalvio:
This would only become really important once we have support for video, so users can know when to expect audio. Although by default audio should even then be muted, same as Relay.

machinaeZER0:
I'm some of the screenshots/videos I've seen as we progress pulling thumbnails for external links, they've been shown partially obscured with an overlay and our hyperlink icon right in the center - the idea here is to avoid that if possible, while also paving the way for future post types to utilize the same layout.

PR #438, opened 2023-07-21
Added blur to external link previews

ajsosa:
Added blurring to externally scraped images. Icon placements will probably have to be adjusted. Specifically in compact mode where the warning triangle symbol is behind the external link symbol. 

Mainly wanted to get this out even though the icon situation wasn't great since blurring NSFW images is an important feature.

### DIFF
--- a/lib/shared/link_preview_card.dart
+++ b/lib/shared/link_preview_card.dart
@@ -1,3 +1,5 @@
+import 'dart:ui';
+
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


machinaeZER0:
Thanks for prepping this, definitely important now that external link images are a thing! If someone ends up looking into #431 it would solve the icon overlap in compact mode... :3

hjiangsu:
Looks good!

Issue #291, opened 2023-07-09
Improved feed switcher


CTalvio:
I would like to do away with the hamburger menu as a main navigation element for browsing content on Lemmy.

By its nature, it centres interactions around the top left, the least reachable of corners. Even with the icon bar gesture to bring it out, Subbed/Local/All switching will involve reaching for the top left. I'm also not a fan of having the list of subscribed communities be cramped into a limited subset of the available screen real estate. This is why the hamburger menu is oft used as a main screen switcher, or as a settings list. Things that are less frequently accessed. I think Thunder should consider using it for the items currently in the settings tab.

In it's stead I would like Thunder to implement a pull-up combination feed/sorting switcher. Pulling up anywhere on the icon bar would pull up a tab, like the one now used for the sort switcher, containing both the sorting options, and the feeds. 

Continuing to scroll past this, would reveal the list of your subscribed communities. In the future, the ability to have certain communities appear first by favouring them, will also work very nicely this way.

Relay has something like this, but it is not exactly what I'm suggesting. Relay combines it with a subreddit-only-search field. I think Thunder should at most include a shortcut to the search page, with community search pre-selected (once thunder is able to search more than just communities).

Relay can still illustrate what I mean:

https://github.com/hjiangsu/thunder/assets/4365015/36e5e229-a453-4b3b-87bb-60148e1dfe85

This _could_ be combined with search, but I find search should remain a tool of it's own, as going forward it will need to be able to browse for communities, posts and comments interchangeably. I'd want this feed switcher to be for accessing the three feed types, and communities, specifically.

machinaeZER0:
This would be neat! I agree that the less one has to interact with the top corners of the screen, the better. Hadn't seen that functionality in Relay, very cool implementation!

phareous:
Is there a way switching between instances could also be incorporated into this? Right now it’s several clicks to get to a different instance feed

CTalvio:
> Is there a way switching between instances could also be incorporated into this? Right now it’s several clicks to get to a different instance feed

I would not be against this.

Issue #174, opened 2023-07-02
Easily accessible NSFW filter on/off


dotPeddy:
**Is your feature request related to a problem? Please describe.**

NSFW content can take up a lot of space in your "feed", depending on the day and what you're viewing. I like access to NSFW, but not always and it also depends on where you are physically (public transportation For example). 

**Describe the solution you'd like**

It would be nice with a checkbox or something in the main view, to filter NSFW content on/off. Depending on the day, NSFW posts can take a lot of space on your feed.

An easily accessible on/off would be really nice. 

**Describe alternatives you've considered**

I've considered just completely turning off NSFW :p 

**Additional context**
No additional context

hjiangsu:
My initial thoughts on this - there can be a new filter button at the top of the feed (beside the sort options). This filter button then allows you to toggle NSFW on/off, and also include other functionality like toggling read posts on/off

PR #443, opened 2023-07-21
#174 Hide NSFW posts togglable option

ajsosa:
https://github.com/thunder-app/thunder/issues/174

Added a hide NSFW posts as an option. Currently requires a manual refresh of feed when selected. Ideally we'll do away with this solution and implement it as suggested in the above issue or it could be handled with what is suggested here https://github.com/thunder-app/thunder/issues/291


hjiangsu:
Looks good!

PR #447, opened 2023-07-22
User/Community Sidebar Redesign + New Stats

bactaholic:
Howdy.

Redesigned the side bar for the user and community and added some fun new stats. Total additions:

-Left-Side Titled Dividers between sidebar sections on user and community sidebars
-Combined Total Posts/Comments Counter
-Combined Total Posts/Comments Score
-Activity Section with...
-Average Posts/mo
-Average Comments/mo
-Average Total Contributions/mo
-"Disable All Score Counters" switch in General Setting

karmaOff
karmaOn
Screenshot_20230722_110826


CTalvio (approved):
Looks good to me! Might want to keep changes to build.gradle and pubspec out of commits, but @hjiangsu should have no trouble leaving those out when merging I think.

machinaeZER0:
Gorgeous! Love the title formatting, very slick!

hjiangsu:
Yeah, I think it looks good but do revert back the changes with `build.gradle` since that'll affect release builds on my end!

One other thing I would say is (if you feel up to it) to start adding in some localization to some of the strings like "Stats".

If you want to see an example of how I did it, this is the commit: https://github.com/thunder-app/thunder/commit/450f6da94005af23c3554f3c364b82ed50108ac1

CTalvio:
I did come up with another thing as well.

The language for your posts/comments/contributions stats is similar to the users per month/week/day for communities.

I think you should switch your stats to "a month" or change the community ones to "/week" etc. just so both are consistent with the other.

bactaholic:
> I did come up with another thing as well.
> 
> The language for your posts/comments/contributions stats is similar to the users per month/week/day for communities.
> 
> I think you should switch your stats to "a month" or change the community ones to "/week" etc. just so both are consistent with the other.

Ahhhhh yes I like that better, more clean. Will change that.

bactaholic:
Good to go

hjiangsu:
@bactaholic I'm going to hold off on merging this until after the settings refactor is done so that you can switch over to the new way of adding settings if thats okay with you!

bactaholic:
> @bactaholic I'm going to hold off on merging this until after the settings refactor is done so that you can switch over to the new way of adding settings if thats okay with you!

Yeah makes sense, will await that and update it when the refactor is done

hjiangsu:
The refactor is up, so you can go ahead and make the changes whenever you have time!

bactaholic:
@hjiangsu Good to go

lib/settings/pages/general_settings_page.dart
```diff
@@ -58,6 +58,7 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
   bool showEdgeToEdgeImages = false;
   bool showTextContent = false;
   bool showPostAuthor = false;
+  bool disableScoreCounters = false;
```
hjiangsu, line 61: I feel like the default value for this should be true. We want users to opt-in to seeing the post scores to limit karma farming in Thunder if possible!

lib/settings/pages/general_settings_page.dart
```diff
@@ -213,6 +218,7 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
       hideNsfwPreviews = prefs.getBool(LocalSettings.hideNsfwPreviews.name) ?? true;
       hideNsfwPosts = prefs.getBool(LocalSettings.hideNsfwPosts.name) ?? false;
       useDisplayNames = prefs.getBool(LocalSettings.useDisplayNamesForUsers.name) ?? true;
+      disableScoreCounters = prefs.getBool(LocalSettings.disableScoreCounters.name) ?? false;
```
hjiangsu, line 221: Change this to `true` to reflect the default value as well

lib/thunder/bloc/thunder_bloc.dart
```diff
@@ -117,6 +117,7 @@ class ThunderBloc extends Bloc<ThunderEvent, ThunderState> {
       bool showEdgeToEdgeImages = prefs.getBool(LocalSettings.showPostEdgeToEdgeImages.name) ?? false;
       bool showTextContent = prefs.getBool(LocalSettings.showPostTextContentPreview.name) ?? false;
       bool showPostAuthor = prefs.getBool(LocalSettings.showPostAuthor.name) ?? false;
+      bool disableScoreCounters = prefs.getBool(LocalSettings.disableScoreCounters.name) ?? false;
```
hjiangsu, line 120: Change this to `true` by default as well

lib/thunder/bloc/thunder_state.dart
```diff
@@ -29,6 +29,7 @@ class ThunderState extends Equatable {
     this.markPostReadOnMediaView = false,
     this.disableFeedFab = false,
     this.showInAppUpdateNotification = true,
+    this.disableScoreCounters = false,
```
hjiangsu, line 32: This one to `true` as well

hjiangsu (commented):
I think it looks good overall, just one thing with making the scores disabled by default unless enabled by the user!

hjiangsu:
One potential thing we can do here in the future is to refactor this to make it a bit easier to work with. For example, creating a new widget for each of the "header" titles since they're all the same logic except for the text of the header.

We could also modularize some other parts of the community/user sidebar since they share a lot of things together. For example:

```dart
 Row(
  children: [
    Padding(
      padding: const EdgeInsets.symmetric(horizontal: 8.0, vertical: 2),
      child: Icon(
        Icons.score_rounded,
        size: 18,
        color: theme.colorScheme.onBackground.withOpacity(0.65),
      ),
    ),
    Text(
      '${NumberFormat("#,###,###,###").format(totalContributionsPerMonth)} Average Contributions/mo',
      style: TextStyle(color: theme.textTheme.titleSmall?.color?.withOpacity(0.65)),
    ),
  ]
);
```

Most of that logic is pretty similar except `Icons.score_rounded` and `'${NumberFormat("#,###,###,###").format(totalContributionsPerMonth)} Average Contributions/mo',`. We could make this into its own widget and pass in the icon and text to render out!

bactaholic:
@hjiangsu Sounds good, i'll make it a widget on its own and update this, hopefully I'll have time later today after my day job.

hjiangsu:
If you'd like, you can just address the comments I've made on the PR first and I can merge that in! We can separate the refactor into a separate PR :D

bactaholic:
@hjiangsu I made disableScoreCounters true by default as requested, problem is I just realized the settings toggle can't be turned off to enable them now.

I swear the settings switches are the bane of my existence, I always have trouble with them and its ends up being something dumb and small i'm overlooking.
I've just pushed the latest changes (with the bug) to my fork. Do you see any obvious reason the settings toggle would be stuck in the on position?

bactaholic:
Oh, forgot to mention I made the sidebar data its own widget(s) as well. Had to make two, one that includes a icon, label, metric, scoreLabel, and scoreMetric and a 2nd that only contains the icon, scoreLabel and scoreMetric, since one requires an extra line of text and values.
named user_sidebar_activity and user_sidebar_stats but I can change those names if needed

bactaholic:
> > I made disableScoreCounters true by default as requested, problem is I just realized the settings toggle can't be turned off to enable them now.
> 
> Hmm, this might just be a case where the value is not being updated properly when it switched on/off in the settings? @bactaholic does the toggle not work even after closing and reopening the app?

Just reopened it, still doesn't work. It's stuck in the on position. The button looks fine to me, and I tried searching the entire directory for "disableScoreCounters =" and looked at every instance it's called to make sure it's not being overridden anywhere. I wasn't able to find anything.

lib/settings/pages/general_settings_page.dart
```diff
@@ -359,6 +365,13 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
                           iconDisabled: Icons.person_off_rounded,
                           onToggle: (bool value) => setPreferences(LocalSettings.useDisplayNamesForUsers, value),
                         ),
+                        ToggleOption(
+                          description: 'Disable All Score Counters',
+                          value: disableScoreCounters,
+                          iconEnabled: Icons.score_rounded,
+                          iconDisabled: Icons.score_rounded,
+                          onToggle: (bool value) => setPreferences('setting_disable_score_counters', value),
```
micahmo, line 373: I'm not sure if this will solve your issue, but even if not, this should be 
``` dart
setPreferences(LocalSettings.disableScoreCounters, value),
```

bactaholic: This does fix it, thank you! I hate chasing down errors with these settings switches. Luckily each time I do I learn something new.

bactaholic:
Alright, @hjiangsu got it, thanks to @micahmo. Turns out in one line I was still referencing the old name, "settings_general_disable_score_counters", hence why the switch did nothing.
This has been fixed.

I got lunch break in 1 hour, will resolve the existing conflict then (probably from an update to the develop branch), then this can finally be merged.

bactaholic:
@hjiangsu could have sworn I saw a conflict but I must have misread it. Got the format so it passes checks, button functions, scores disabled by default. Good to go.

hjiangsu (approved):
Looks good! There are some more suggestions but I don't want to hold up the PR from being merged because its at a good state.

For perhaps a future PR, we can address some of the following:
- Adding localization strings for the different metrics
- Double checking accessibility for voice over/talk back
- Changing the option from "Disable score counters" to "Enable score counters"

PR #448, opened 2023-07-22
Bug Fix: Comment edge colors follow comment on gesture.

ajsosa:
Rearranged comment_card layout so the colored left edge columns follow the card when using gestures.


ajsosa:
There is a very minor graphical thing where the child card can cover a the left colored line of only it's parent when in thin style mode. But it's super minor and can be improved on later.

PR #454, opened 2023-07-23
Bug Fix: In comment context view, voting on comment then refreshing loaded all comments

ajsosa:
Fixed bug where when viewing comment context, voting on a comment and then refreshing would load all comments instead of the same comment context.

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -233,7 +233,7 @@ class PostBloc extends Bloc<PostEvent, PostState> {
                   commentResponseMap: responseMap,
                   commentPage: 1,
                   commentCount: responseMap.length,
-                  hasReachedCommentEnd: getCommentsResponse.isEmpty || getCommentsResponse.length < commentLimit,
+                  hasReachedCommentEnd: getCommentsResponse.isEmpty || commentTree.length < commentLimit,
                   sortType: sortType),
             );
           }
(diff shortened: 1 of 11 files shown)


ajsosa:
Cleaned up the fix a bit to make use of the data in post state. Also fixed similar issues with deleting comments, saving comments, and deleting.

There is still a bad bug when deleting/restoring in a comment context. The comment tree gets a bit messed up. Will investigate further.

hjiangsu:
> There is still a bad bug when deleting/restoring in a comment context. The comment tree gets a bit messed up. Will investigate further.

@ajsosa Should I hold off on merging this PR until this is fixed?

ajsosa:
@hjiangsu Yeah give me a few more days to see if I can iron some of this stuff out.

hjiangsu:
Sounds good! Let me know whenever you feel ready :D

ajsosa:
@hjiangsu I kind of went off the rails a bit with this one ha ha. I'll try not to do this again.

But basically I fixed various issues with commenting in general and then some.

#### Commenting in comment context
- Creating new comment to post outside of context will fetch all comments and remove highlighting.
- Creating new post to existing context will just append to existing tree without fetch.
- Deleting/restoring existing comment will update existing tree without fetch.

#### Commenting in full post
- Deleting/restoring existing comment will update existing tree without fetch.

#### Comments general
- Added spinning circle animation on comment card while waiting for comment to be deleted/restored.
- Added vote count to comment card in comment list for the User view.

#### Other fixes
- Fixed another edge case of the loading more comments infinite spinning circle.
- Fixed infinite spinning circle when loading a user's posts in the User view. This would happen due to the post count not always matches with the actual count of posts your instance has. Sometimes it doesn't have all of them.

hjiangsu:
Hmmm, it looks good overall but I do have a couple of questions!

> Creating new post to existing context will just append to existing tree without fetch.
Deleting/restoring existing comment will update existing tree without fetch.

What happens if deleting/restoring a comment fails? How does Thunder fall back to the previous state and how is the user notified of the failure?

ajsosa:
The snackbar is supposed to pop up with the failure and the comment tree left unchanged with the spinning wheel gone. Unfortunately the snackbar is broken right now. I decided not to work on that in this PR because it's a completely separate problem and this PR is already really big. It should be noted, the failure state currently without these changes is exactly the same. Nothing happens because the toast is off screen due to the FABs even if FABs are disabled. Even worse, there isn't any feedback that anything is happening at all.

I'm currently investigating how to fix the broken snackbar which would solve the error problem.

hjiangsu:
Alright, sounds good!

PR #458, opened 2023-07-24
Show community subscription indicator

micahmo:
This PR adds an icon next to subscribed communities. Note that this is only shown in Local/All feeds, not Subscribed (where it's implied every community is subscribed) or within a community itself (where you can see the status at the top).

https://github.com/thunder-app/thunder/assets/7417301/ad42490b-6fff-41f1-9fd1-10507b4a351c


hjiangsu:
Looks good!

PR #460, opened 2023-07-24
Fix Show Post Author icon

micahmo:
Followup to #381. This uses our standard "rounded" icon for the "Show Post Author" setting. It also adds a separate icon to represent the "disabled" state since most (all?) of our other settings have that.


hjiangsu:
Nice

PR #462, opened 2023-07-24
Show instance taglines

micahmo:
This PR adds community taglines to the top of the feed. This behavior now matches the web and Jerboa.

### Example (programming.dev)

| image |
| - |

| image |
| - |

#### Instances with no taglines have no change.

| image |
| - |

### DIFF
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -237,6 +237,13 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
             postIds.add(post.postView.post.id);
           }
 
+          // Fetch any taglines from the instance
+          FullSiteView fullSiteView = await lemmy.run(
+            GetSite(
+              auth: account?.jwt,
+            ),
+          );
+
           emit(
             state.copyWith(
               status: CommunityStatus.success,
@@ -250,6 +257,7 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
               subscribedType: subscribedType,
               sortType: sortType,
               communityInfo: getCommunityResponse,
+              taglines: fullSiteView.taglines,
             ),
           );
         } while (tabletMode && posts.length < limit && currentPage <= 2); // Fetch two batches
--- a/lib/community/bloc/community_state.dart
+++ b/lib/community/bloc/community_state.dart
@@ -17,6 +17,7 @@ class CommunityState extends Equatable {
     this.communityName,
     this.communityInfo,
     this.blockedCommunity,
+    this.taglines = const [],
   });
 
   final CommunityStatus status;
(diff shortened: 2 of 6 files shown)


machinaeZER0:
I know that it doesn't do this on web, but if this is being used for telling people potentially important stuff should we consider theming it in a way that pops out more? Would that be possible/something that we should consider?

micahmo:
@machinaeZER0, sure! I didn't want it to get too annoying (as it will be at the top of the feed every time you open the app / refresh), but we can certainly try to emphasize it a bit. Maybe a darker weight and a color?

| image |
| - |

machinaeZER0:
Hmm, yeah - it's too bad it can't be dismissible/collapsible since people theoretically would only need to see it once, and then not again unless the content had changed.

Would it look super out of place if the background was slightly themed instead of the text itself? Or if it was presented over a large chip? (Is it still a chip if it's this big?)

micahmo:
> it's too bad it can't be dismissible/collapsible

It could be! Would that be worth trying?

> Would it look super out of place if the background was slightly themed instead of the text itself? Or if it was presented over a large chip?

What do you think?

| image |
| - |

| image |
| - |

machinaeZER0:
I like that! I think if possible it would be good to give the text a bit more padding on all sides. If it is possible to swipe it away or x it out I think that would be great - I don't think we have an x to close in the app yet so if that feels out of place maybe just swiping away makes the most sense/won't require additional padding?

hjiangsu:
I personally like it with the chip look, but as @machinaeZER0 mentioned, have extra padding on all sides.

Another idea with this is to have a way to collapse it so that you only see the first line of text (in case there are times where its a extremely long tagline and it takes up half of the screen's height. Or, be able to swipe it away as @machinaeZER0 mentioned!

micahmo:
Thanks for the feedback. Let me add the extra padding as you both suggested.

In terms of dismissing, I'm sure that's possible, but I think we would have to remember which tagline you dismissed so we don't show it again (until/unless it changes). I like the idea of collapsing to a single line and allow it to be expandable. I think that's a good compromise of always showing it but not being annoying. Let me run with that.

machinaeZER0:
I like that collapsing idea, good thinking! I may be overthinking, but to me it feels like a chip would be swipe-able, where if it were a pane it would make more sense that it would collapse. What do you guys think? Either way I like that the background would be colored in an eye catching way, as currently mocked up!

hjiangsu:
>I like that collapsing idea, good thinking! I may be overthinking, but to me it feels like a chip would be swipe-able, where if it were a pane it would make more sense that it would collapse

I think in this case, its less of a chip and more of a card. For a card, I think it makes sense to allow expanding/collapsing

https://m3.material.io/components/cards/guidelines#652a58b2-e429-4409-a3c8-4cc08fa8ae33

machinaeZER0:
I think this looks great! I wouldn't be mad if it was entirely dismissible, but I think as is it's a great implementation :) nice job!

PR #463, opened 2023-07-24
Small image viewer improvements

micahmo:
This PR introduces a couple improvements to the image viewer.
 - The open and close times are reduced further. Currently this still feels a little sluggish. Also, close always feels slower than open, so I made it relatively smaller.
 - `ExtendedImage` cache is not immediately cleared. Currently every time we scroll through or open a post and go back, etc., the image preview flashes white and reloads. By allowing caching, we have a much smoother experience.

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -47,8 +47,8 @@ class _ImagePreviewState extends State<ImagePreview> {
       // TODO This is probably where BlocProvider breaks
       PageRouteBuilder(
         opaque: false,
-        transitionDuration: const Duration(milliseconds: 200),
-        reverseTransitionDuration: const Duration(milliseconds: 200),
+        transitionDuration: const Duration(milliseconds: 100),
+        reverseTransitionDuration: const Duration(milliseconds: 50),
         pageBuilder: (BuildContext context, Animation<double> animation, Animation<double> secondaryAnimation) {
           String heroKey = generateRandomHeroString();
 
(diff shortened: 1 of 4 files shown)


hjiangsu:
Hmm, we need to be a bit careful about allowing cache to stay - this can potentially lead to regressions in performance and/or memory leaks.

Before merging in this change, we'll have to check how the performance is with the cache on vs off! I can do a bit of tests on my end as well to double check the behaviour

micahmo:
> this can potentially lead to regressions in performance and/or memory leaks

I think this can only help performance. Right now we have to download the image any time the preview or full image viewer is built. This should save a lot of bandwidth. But I agree about memory. I'm not sure if there's a cache limit for `ExtendedImage` and/or if we can clear it ourselves.

> I can do a bit of tests on my end as well to double check the behaviour

Let me know what you find!

micahmo:
@hjiangsu I just hit another use case today where having no cache is really annoying. If you're scrolling in a post view near the top, every time you get in range of the post media, it will refetch and resize and redraw, pushing your comments down. Gets frustrating. 😆


https://github.com/thunder-app/thunder/assets/7417301/2631c5d1-3248-432c-a78f-541ef60f4e31

hjiangsu:
Hey! So I did a few profiling tests for memory usage, and it seems to be relatively okay with the changes that you made. I did notice something else which could be unrelated to this change. Theres a point in time that the app freezes for a brief second, and there's a huge spike in memory usage - it doesn't seem like its related to `extended_image` but I'll have to do more investigation on it.

I _dont_ think its related to this, so I'll go ahead and merge the changes in!

> I just hit another use case today where having no cache is really annoying. If you're scrolling in a post view near the top, every time you get in range of the post media, it will refetch and resize and redraw, pushing your comments down.

This is more related to not having the size dimensions for the images. So what ends up happening is that because there are no size dimensions, flutter can't pre-allocate the correct amount of space required to show the image and you see that layout jank in the video. Having the cache will help a little bit, but won't solve the issue entirely (it'll just be a bit less frequent)

micahmo:
> I'll have to do more investigation on it.

Sounds good! I would like to try some profiling too as I often get a bit of jank here and there. 

> Having the cache will help a little bit, but won't solve the issue entirely (it'll just be a bit less frequent)

Fair enough!

PR #468, opened 2023-07-25
READY TO MERGE: Post card changes, add badges

CTalvio:
- #431
- Move favorite and pinned icons to end of title
- Pinned posts have a green title
- #511

Lots of small changes to both post view types


machinaeZER0:
Sounds like for now this is on hold, but copying over my comments from Matrix/ #431: 

- I think I'd put the badge in the bottom right since the top corners often have stuff like faces - less likely to overlap something crucial in the bottom corner
- When the time comes to implement this more fully (which sounds like it'll be in a larger rewrite) I'd consider tweaking the thumbnail layout so that we could offset the badge - this obscures the image less while still giving us plenty of space to use badge icons of a pleasant size
- Noting your callout that whatever we land on for compact should match an implementation in normal view (aside from potential scaling tweaks)

I know ours won't look exactly like in Bean (defer to what feels right for Thunder), but pasting them as an example here. Thanks for taking an exploratory stab at this!

micahmo:
Great start! Agree with @machinaeZER0's comment.

CTalvio:
Screenshot_20230804_044821
Screenshot_20230803_011013

Progress. I'm moving the saved/pinned icons to the title, they make sense there, and the lower part of the cards is getting cramped.

micahmo:
Nice progress! I don't mind saved/pinned in the title, but maybe after would be better? I think (a) this would match other UIs, (b) there wouldn't be a problem of overflow since we wrap, and (c) it would keep the beginning of the title in a consistent place (no UI clutter to skip over before reading the title). Also I wonder if the pin icon could be a tiny bit smaller. 😊

CTalvio:
Span doesn't handle the icon well at the end, it has be made super tiny to fit.

Maybe the whole title could be tinted green for pinned posts, like the webUI?

machinaeZER0:
If you don't mind me weighing in - I do like the idea of putting the pin at the end of the title, but if that's not possible I wouldn't worry about it. Tinting the text green would be my vote, since (to your point) it matches the webui and doesn't feel too "loud" (though if the pin icon stays as well I would move it next to the title - looks like right now the Saved star places itself in between?)

Alternatively, would it be worth trying something with chips in this situation? If a pinned title becomes a green chip/badge, would that do anything to solve the issue where the pin icon has to be made super small to fit? Probably way more work unless something can be carried over from the work on username display in comments (and it might be unnecessarily loud)

CTalvio:
image

I kinda like it when the pin/star is small and at the end. ALSO BADGES!

machinaeZER0:
Oh wow!

machinaeZER0:
I think the link one looks perfect - I might shrink the other icons a little within their respective badges, personally. Do they all have slightly different badge colors, or just the link one? 

Thank you for working on that!

micahmo:
> image
> 
> I kinda like it when the pin/star is small and at the end. ALSO BADGES!

This looks absolutely amazing! 🎉

CTalvio:
Screenshot_20230806_002424
Screenshot_20230806_003055

Allright, this is ready to go!

lib/community/widgets/post_card_metadata.dart
```diff
@@ -30,87 +30,91 @@ class PostCardMetaData extends StatelessWidget {
...
+                    Padding(
+                      padding: const EdgeInsets.only(left: 2.0),
+                      child: IconText(
+                        textScaleFactor: state.contentFontSizeScale.textScaleFactor,
+                        icon: Icon(
+                          Icons.public,
+                          size: 15.0,
+                          color: readColor,
+                        ),
+                        text: Uri.parse(hostURL!).host.replaceAll('www.', ''),
```
micahmo, line 111: In the crazy unlikely chance that `www` is contained elsewhere in the URL, do you want to use `replaceFirst` here?

lib/community/widgets/post_card_view_compact.dart
```diff
@@ -48,89 +50,217 @@ class PostCardViewCompact extends StatelessWidget {
...
+                  ),
                 ),
-              ),
         ],
       ),
     );
   }
 }
+
+class TypeBadge extends StatelessWidget {
```
micahmo, line 202: Should this go in its own file?

micahmo (approved):
LGTM!

lib/community/widgets/post_card_view_compact.dart
CTalvio: I don't see a downside to that.

hjiangsu:
Looks good! The only thing is to address the merge conflicts here (I think it most likely has to do with changing the textScaleFactor to use the correct one)

E.g., use `metadataFontSizeScale` instead of `contentFontSizeScale` for the post card metadata text!

hjiangsu:
@CTalvio Could you double check the formatting - it didn't pass the checks!

PR #473, opened 2023-07-25
Add visual swipe gesture picker

micahmo:
This PR introduces new styling for the swipe actions. The motivations are (a) save some space, (b) give a clear visual indicator of what's what (since I always forget!) and (c) pave the way for 4 options on each side.

https://github.com/thunder-app/thunder/assets/7417301/69a43e04-9e87-42e4-833d-dfb4d7b5231d

Since the above recording was taken, a hint text and inkwell animation were added to make it more clear that the swipe actions are interactable.

| image |
| - |


machinaeZER0:
GORGEOUS!

hjiangsu (approved):
It looks really good! I like the refactor you made with the swipe icons and colours!

Issue #266, opened 2023-07-07
Screen goes black after posting comment


krestenlaust:
**Description**  
The instant I press send, the screen goes black and the application closes. When I open the app again the screen is just black, and I have to force close the app to restart it.

The comment *is* submitted and visible, once restarted.

**How to Reproduce**
I think this is an issue oddly specific to my phone, since it's so simple to reproduce.
1. Make a comment
2. Screen goes black

**Screenshots**  
Difficult to document.

**Device & App Version:**  
 - Device: OnePlus 6T
 - OS: Android 11
 - Version: 0.2.1-alpha.1+11

krestenlaust:
This is 100% consistent and happens every time. I'm not sure what options I have for providing logs

hjiangsu:
I have not been able to reproduce this on my end unfortunately, so it's hard for me to figure out the root cause at the moment

krestenlaust:
Strange. It doesn't do it now. It's happened to me probably 10 times already in the past. I haven't upgraded either.

hjiangsu:
There is no logging at the moment, but maybe that might prove useful to have if its do-able. Previously, the logging was done through Sentry as it detects all errors/exceptions, but I've removed that in the latest version

krestenlaust:
Just wanted to let you know, that this bug still affects me. It's probably half of the time I post.

hjiangsu:
Is this with the latest pre-release? v0.2.1-alpha.4+13?

micahmo:
@krestenlaust Since you're using an alpha, I'm assuming you're installing from GitHub? Thunder does not support in-app updates at this time. You might check out [Obtainium](https://github.com/ImranR98/Obtainium) which is a cool app that can install and keep other apps up-to-date based on GitHub releases.

krestenlaust:
Thunder used to support in-app update notifications. I might look at Obtainium to replace the lost functionality.

micahmo:
> Thunder used to support in-app update notifications.

Hmm, the code does seem to be there, but it's never worked for me. 😊 @hjiangsu, is this something we officially support? Should we see if it still works?

But yes, Obtainium does the job nicely as well.

hjiangsu:
> Hmm, the code does seem to be there, but it's never worked for me.

It should still be working. I wonder if it stopped because of the migration from my personal GitHub account to the organization account? This might be something we need to double check

micahmo:
[...]

PR #478, opened 2023-07-25
Fix version parsing

micahmo:
For in-app update notifications, our current version parsing was a simple string comparison. This hasn't been working well lately, especially with our extended usage of semver. The [version](https://pub.dev/packages/version) package helps with that.

Ref: https://github.com/thunder-app/thunder/issues/266#issuecomment-1650335600

Issue Number: #615


micahmo:
It also occurred to me that we're only grabbing the latest full release, which is another reason @krestenlaust wasn't getting update notifications. Now that the app is on Google Play and TestFlight, maybe the in-app notifications should include alphas and be off by default. Thoughts on that?

krestenlaust:
Agree, I think it would be good with an option to opt-out of alpha releases, but allow people who want alpha versions like me, to still be notified.

micahmo:
Ok I updated the PR to disable in-app update notifications by default. I am going to make this a draft until we have a new prerelease alpha, then I can verify that we always get the "newest" as opposed to the "latest" (which are only the stable ones).

krestenlaust:
It is my belief that the notifications for full releases should be on by default, and alpha-releases are opt-in

hjiangsu:
We would also need to do testing on iOS devices for this
- TestFlight builds are only for the full releases and not nightlies/pre-releases
- Versioning is different in iOS than it is on Android (iOS only allows numerical values and does not contain any characters)

micahmo:
> TestFlight builds are only for the full releases and not nightlies/pre-releases

Yes, analogous to Google Play Store releases in that way. That's why I was suggesting the in-app notifications could be for prereleases, since we'll assume you'll get stable releases through other channels.

> Versioning is different in iOS than it is on Android (iOS only allows numerical values and does not contain any characters)

Ahh, maybe that's why the string comparison worked for iOS. Maybe we can inject the current full version number somewhere else (even if it's a `const String` we stick somewhere) so we have a consistent thing to check.

micahmo:
Hey all, I'm finally revisiting this issue. I've considered quite a lot of alternatives here, like multiple options for stable/prerelease, automatic in-app updates (including [Android's native one](https://developer.android.com/guide/playcore/in-app-updates)), differentiating between different install sources, etc. For posterity, here are all the packages related to this topic that I found interesting.

* https://pub.dev/packages/in_app_update
* https://pub.dev/packages/upgrader
* https://pub.dev/packages/new_version
* https://pub.dev/packages/store_checker

However, I think in-app update notifications are mainly geared toward those who did not install through a store. Therefore, I think what we have in this PR is very close to what we want, with a couple more additions:
* Rename the setting to imply that it's for GitHub / prereleases.
* Improve semver checking on iOS by putting the full version number (including prerelease moniker) in a `const String` in code. We could also update the CI to verify that this matches `pubspec.yaml`.

In the future, we could do something more fancy, like checking whether the app was installed via the store, and showing the native store updater.

In order to avoid the asynchronous nature of GitHub discussions, I'll move forward with my idea here, but please leave feedback!

micahmo:
This will probably not go in the next GA, so I'll leave it as as draft, but just for future reference, this should be ready to go once #574 is done.

lib/core/update/check_github_update.dart
```diff
@@ -37,3 +37,20 @@ Future<Version> fetchVersion() async {
...
+bool _compareVersions(version_parser.Version first, version_parser.Version second) {
+  // Handle different builds (which are not normally part of semver sorting)
+  int? firstBuild = int.tryParse(first.build);
+  int? secondBuild = int.tryParse(second.build);
+  if (firstBuild != null && secondBuild != null && firstBuild != secondBuild) {
+    return secondBuild > firstBuild;
+  }
+
+  return second > first;
+}
```
micahmo, line 56: Despite using the `version` package, we still have to do some custom handling because our versioning is not purely semver in terms of sorting. Here are the scenarios I tested. In each case, the second should be considered greater than the first.

* Alpha to alpha across releases
  * 0.2.1-alpha.2+13 -> 0.2.1-alpha.1+14
* Alpha to alpha within a release
  * 0.2.1-alpha.0+14 -> 0.2.1-alpha.1+14
* Alpha to release
  * 0.2.1-alpha.4+13 -> 0.2.1+13
* Release to alpha
  * 0.2.1+13 -> 0.2.1-alpha.1+14

micahmo:
I'm all in favor of following semver more closely! I think the problem with using the build number is that it's not included in the sermver comparison. I understand that iOS doesn't support alpha characters. Could we use the prelease moniker with just a number? `0.2.2-1`, `0.2.2-2`? I'd have to make sure sorting works, but that might satisfy all of the requirements.

hjiangsu:
Hmm, that might be possible. I just hope that doesn't cause more confusion for users when describing the pre-release versions (having `alpha` helps distinguish that a version is a pre-release as compared to a general release whereas just having the value might make it less obvious)

micahmo:
> (having `alpha` helps distinguish that a version is a pre-release as compared to a general release whereas just having the value might make it less obvious)

I agree, but since iOS doesn't support alpha characters, I think we have to come up with some other solution, right? Otherwise it'll always be hard for us to grab the "real" version from iOS builds. Since you already suggested using just a single number for nightlies, I was thinking we could do that the "right" way according to semver.

Would it help if tapping the version number copied it to the clipboard? Then users might have an easier time sending us their version number.

hjiangsu:
@micahmo I've added a few changes here!
- Updated the branch to reflect latest develop changes
- Tweaked the code slightly to now account for the format `x.y.z-n` for both iOS and Android

Here are the screenshots:

For `0.2.4-1+17`:
simulator_screenshot_403F91FE-2910-419A-9E93-949B9876E574

image


For `0.2.4+17`:
simulator_screenshot_20307B8C-C3A8-436A-A1BD-0A85E06485B8
image

micahmo (commented):
Alrighty, this looks great! The version is displayed exactly as expected on Android.

In terms of checking against GitHub, I tested the following scenarios (faking the API response), and they worked great (where `->` implies we were prompted to upgrade)
* 0.2.4-2 -> 0.2.5-1
* 0.2.4-1 -> 0.2.4-2
* 0.2.4-1 -> 0.2.4
* 0.2.4 -> 0.2.5-1

And since the build number is gone, I was able to remove some of my hacky comparison stuff too.

I think we're good to go here! Although `pubspec.yaml` currently doesn't match `globals.dart`, so the CI should fail. 😉

PR #482, opened 2023-07-26
Tappable author and community

micahmo:
Since #466 removed the ability to tap authors/communities in compact mode, I am bringing back that feature via an option (off by default). I also added inkwell animations so that the tap is more obvious.

https://github.com/thunder-app/thunder/assets/7417301/8982d821-e1fc-4220-a7da-b06877f03532


hjiangsu:
Looks good! Thanks :D

PR #483, opened 2023-07-26
Match swipe picker inkwell border

micahmo:
This is a super tiny fix on top of the recent swipe gesture picker. Previously, the inkwell animation was square, going outside the bounds of the row. Since `InkWell` has its own `borderRadius` property, it can match exactly.

### Before

https://github.com/thunder-app/thunder/assets/7417301/8b0e73f0-c5d8-478a-bf0f-ac0f742cc153

### After

https://github.com/thunder-app/thunder/assets/7417301/f27a8626-420c-4d34-9d7d-91d8f49ffd0d


CTalvio:
This change should be introduced in other places where the inkwell effect is used. Image thumbnails come to mind, at least.

PR #485, opened 2023-07-26
Use spinner for downloading media to share

micahmo:
As discussed in Matrix chat, we would like to use snackbars less when there is an ongoing activity, and prefer inline spinners. Here I have done that for the share button within the image viewer. A snackbar is still shown if/when there is an error.

https://github.com/thunder-app/thunder/assets/7417301/dba67ad0-afea-4841-a7d4-86b792cada97

### DIFF
--- a/lib/shared/image_viewer.dart
+++ b/lib/shared/image_viewer.dart
@@ -373,7 +373,7 @@ class _ImageViewerState extends State<ImageViewer> with TickerProviderStateMixin
                                   )
                                 : Icon(
                                     Icons.share_rounded,
-                                    semanticLabel: "Comments",
+                                    semanticLabel: "Share",
                                     color: Colors.white.withOpacity(0.90),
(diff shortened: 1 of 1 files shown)


lib/shared/image_viewer.dart
```diff
@@ -370,14 +359,24 @@ class _ImageViewerState extends State<ImageViewer> with TickerProviderStateMixin
...
+                            icon: isDownloadingMedia
+                                ? SizedBox(
+                                    height: 20,
+                                    width: 20,
+                                    child: CircularProgressIndicator(
+                                      color: Colors.white.withOpacity(0.90),
+                                    ),
+                                  )
+                                : Icon(
+                                    Icons.share_rounded,
```
hjiangsu, line 375: Could you change this to "Share" instead of "Comments"? I think that was something that snuck in from a previous PR!

hjiangsu (approved):
Just one quick change, otherwise it looks good!

PR #490, opened 2023-07-26
Resepect comment deleted property in reply modal

ajsosa:
This is a fix for this reported bug.

https://lemmy.world/post/2213300

### DIFF
--- a/lib/inbox/bloc/inbox_bloc.dart
+++ b/lib/inbox/bloc/inbox_bloc.dart
@@ -85,9 +85,9 @@ class InboxBloc extends Bloc<InboxEvent, InboxState> {
             return emit(
               state.copyWith(
                 status: InboxStatus.success,
-                privateMessages: privateMessageViews,
-                mentions: personMentionViews,
-                replies: commentViews,
+                privateMessages: removeDeleteMessages(privateMessageViews),
+                mentions: removeDeleteMentions(personMentionViews),
+                replies: removeDeleteReplies(commentViews),
                 showUnreadOnly: !event.showAll,
                 inboxMentionPage: 2,
                 inboxReplyPage: 2,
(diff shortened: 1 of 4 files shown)


hjiangsu (approved):
Looks good! I wonder if it might be better to move this at the bloc level or even at the API level rather than here.

If its at the bloc/API level, we can intercept all comments and we won't have to worry about implementing this everywhere we potentially fetch comments

ajsosa:
I think that's a good idea. I'll take a look at doing that tomorrow. It may require tweaking some of the other comment interactions since they now avoid doing fetches now.

hjiangsu:
I guess I can hold off on merging this until you take a look at the other solution? Or if you'd like, I can merge this in for now and we can implement my suggestion at a later date

ajsosa:
@hjiangsu After implementing your suggestions, we still weren't showing deleted by creator in the inbox. That's because for that we just output whatever is coming from the API straight. I think if we want to hide deleted comments without having to think about it, it has to be done in the `lemmy_api_client`. Thoughts?

ajsosa:
Something else I'm wondering about. For inbox specifically. Should we even present anything in that view that has been deleted? Though I'm not sure if that could cause a situation where your inbox notification shows you have something and then when you go to that view, there is nothing because it was deleted.

ajsosa:
Went ahead and added some functions for stripping out mentions, messages, and replies from the inbox so they never appear in the inbox view.

hjiangsu:
>  I think if we want to hide deleted comments/messages without having to think about it, it has to be done in the lemmy_api_client. Thoughts?

Hmm, I wonder if there's ever going to be a use-case where we'll need to original non-deleted comment. I _feel_ like its best to keep it within Thunder since the `lemmy_api_client` shouldn't directly modify information coming in from the API because that serves as the single source-of-truth which everything else relies upon

> Should we even present anything in that view that has been deleted? Though I'm not sure if that could cause a situation where your inbox notification shows you have something and then when you go to that view, there is nothing because it was deleted.

What does the web ui do? I think we should try to keep it as similar to the web ui as possible so that there's less confusion overall. If the web-ui hides it, then we can hide it, and if the web-ui has a placeholder for it, then we can keep a placeholder there and disable all actions for that deleted comment

ajsosa:
Reverted inbox to show deleted mentions, messages, and replies but made sure to strip out the content.

hjiangsu:
Looks good! Thanks :D

PR #492, opened 2023-07-27
More robust community navigation

micahmo:
This PR further improves community navigation. The motivation here was (a) to add support for community links in the form `instance.tld/c/community@otherinstance.tld`, and (b) to fix the issue that we treated any URL with `/c/` as community links (#343). As a result, the community link parsing has been overall improved. I used this post as a testing ground: https://reddthat.com/post/932411

Note that for `/c/` links we will do a quick ping before navigating to the link. _This will slow down navigation initially only for links with `/c/`_; however, successful instances will be added to an in-memory cache to at least speed up subsequent checks in the current session. Unless anyone has any clever ideas, I don't see any way around this.

### DIFF
--- a/lib/utils/instance.dart
+++ b/lib/utils/instance.dart
@@ -1,3 +1,5 @@
+import 'dart:collection';
+
 import 'package:lemmy_api_client/v3.dart';
 
 String? fetchInstanceNameFromUrl(String? url) {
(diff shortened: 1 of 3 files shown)


lib/utils/instance.dart
```diff
@@ -9,19 +11,40 @@ String? fetchInstanceNameFromUrl(String? url) {
...
+/// Matches community@instance.tld
+/// Puts community in group 2 and instance.tld in group 3
+/// https://regex101.com/r/1VrXgX/1
+final RegExp instanceName = RegExp(r'^!?(https?:\/\/)?((?:(?!\/c\/c).)*)@(.*)$');
+
+/// Checks if the given text references a community on a valid Lemmy server.
+/// If so, returns the community name in the format community@instance.tld.
+/// Otherwise, returns null.
+Future<String?> getLemmyCommunity(String text) async {
+  var fullCommunityUrlMatch = fullCommunityUrl.firstMatch(text);
```
hjiangsu, line 33: I'm wondering if we can replace `var` with a specific type here just to reduce the amount of times we use var

I think using var skips out on some of the useful dart's static type checking

hjiangsu (commented):
Just had one discussion point that we can think about!

lib/utils/instance.dart
micahmo: Done!

hjiangsu:
And approved!

PR #494, opened 2023-07-28
Focus search on nav tap; remember state

micahmo:
Related to #117, this PR introduces a quick way to show the keyboard on the search page by pressing search on the nav bar a second time. This is a very common pattern in apps, and I think it's less intrusive than immediately showing the keyboard upon switching to the search page.

As part of these changes, I (originally unintentionally) ended up making search more stateful. I ended up liking this, so I leaned into it by making the widget keepalive. That way, you won't lose search results by navigating to another page. The last step to make this work is to detect when the instance (account) changes, so that we can reset search (since the results would be invalid).

So to summarize...
* Focus search text box by tapping on nav bar
* Remember search results
* Reset search when changing accounts

https://github.com/thunder-app/thunder/assets/7417301/4119591f-ef81-436d-884f-72026a2e8f8c


machinaeZER0:
Looks great, I like the double-tap behavior! Will have to try that out in other apps and see which ones have adopted that :)

micahmo:
> Will have to try that out in other apps and see which ones have adopted that

I just went through a bunch of apps on my phone, and maybe it's not as common as I thought. 😆 But I think that's because a lot of apps put their search bar at the top instead of the bottom nav. Regardless, I think it's a nice design that makes the feature accessible without being annoying. Plus we've already established that pattern with the jump-to-top of the feed. Finally, if we extend our search page to include filters, the user will probably want to interact with those before typing.

hjiangsu:
I think this looks good! One use case I would double check is when the user is not logged in to any accounts. They should still be able to subscribe and search for communities!

micahmo:
> One use case I would double check is when the user is not logged in to any accounts. They should still be able to subscribe and search for communities!

Yes, that works fine! One weird quirk is that you will still be connected to the instance of the last account after you sign out. But that was true before. I think we need a separate instance switcher for anonymous mode.

However, I did discover another issue which is that search resets when subscribing. I need to distinguish between a full account switch and just a new subscription when resetting. I'll fix.

hjiangsu:
> I think we need a separate instance switcher for anonymous mode.

Yeah, this was something I was working on in the very beginning but never got around to it 😅

> I need to distinguish between a full account switch and just a new subscription when resetting. I'll fix.

Sounds good! Thanks :D

micahmo:
Ok I think this is ready to go! I tested logging in, logging out, switching accounts, subscribing/unsubscribing, etc. The only time the search state should get reset is when there's a real account switch. Otherwise it should be persistent. Feel free to do additional testing.

https://github.com/thunder-app/thunder/assets/7417301/96e3a36e-d0ad-4cc0-8bc6-38fc953edca8

ajsosa (approved):
Looks good :)

PR #495, opened 2023-07-28
Support exact community searches

micahmo:
This PR introduces the ability to search for a community by exact name. Sometimes I know exactly what community I want to add, but there are too many in the search results to sift through. This should support all types of valid community references.

https://github.com/thunder-app/thunder/assets/7417301/9965e8de-a20a-48e7-b079-b02455c7ac76

### DIFF
--- a/lib/search/bloc/search_bloc.dart
+++ b/lib/search/bloc/search_bloc.dart
@@ -9,6 +9,7 @@ import 'package:thunder/account/models/account.dart';
 import 'package:thunder/core/auth/helpers/fetch_account.dart';
 
 import 'package:thunder/core/singletons/lemmy_client.dart';
+import 'package:thunder/utils/instance.dart';
 
 part 'search_event.dart';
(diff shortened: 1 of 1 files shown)


CTalvio:
I assume this does still not allow the triggering of your host instance to federate a given community? That's only possible while logged in even the webUI, and still something I haven't seen another client support.

micahmo:
> I assume this does still not allow the triggering of your host instance to federate a given community?

That's a good question. Is there a small instance I could test this with? I skipped auth for this `GetCommunity` call, but I could add it back if it triggers this.

CTalvio:
Easiest way to test this is probably to create a community on one instance, and then to try federating it to another.

You can delete it once you're done.

micahmo:
I think I'm getting rate limited on community creation. 😆 But I _think_ it works once auth is there. Worked a couple times for me at least. Pushing those changes.

PR #496, opened 2023-07-29
Add username navigation

micahmo:
This PR supports user navigation from markdown comments. Note that we're not changing the markdown parser to support anything new. Rather anything it already detects as a clickable link that we can determine refers to a user should be navigable now. This includes things like `/u/username@instance.tld` and `instance.tld/u/username@otherinstance.tld`. We even support some "unofficial" ways to refer to users like `@user@instance.tld`.

I used the following post to test: https://reddthat.com/post/932411


hjiangsu:
Looks good!

PR #497, opened 2023-07-29
Fab rework

CTalvio:
Need some extra eyes on this.

Works as intended on feeds, but on posts, bloc changes don't trigger a UI update, and I don't know what's different.


micahmo:
I grabbed your branch, but I can't get the FABs to appear. I'm sure I'm doing something wrong. Any ideas?

https://github.com/thunder-app/thunder/assets/7417301/c9a6a9b6-2807-4ef2-a322-ce2e83afed57

CTalvio:
Does it maybe start off dismissed? I changed the summon gesture to a swipe up from the bottom right corner because that can be activated faster than the long press, try that.

CTalvio:
Yup, that was it.

micahmo:
> Yup, that was it.

Thanks! It works for me now!

> on posts, bloc changes don't trigger a UI update

Are you still having this issue? Actions on posts seems to work fine for me.

In fact, everything looks very good! I think this is very close to mergeable. I see a couple issues that would be nice to get cleaned up first.

 - On posts, dim the background when the additional actions are shown. Right now things are hard to read.
   image
- I don't see New Post when navigating to a community from the app drawer.
- Subscriptions does not work when viewing a community in a navigated (i.e., pushed) page.
- On the main feed, the default action (dismiss read) is duplicated in the additional actions, but in the post view, the main action (reply) is not. I don't have a preference but the behavior should be similar.
- Opening additional actions from a post view and then navigating back causes the additional actions to be opened on the main view, which is unexpected.

I would _not_ hold up this PR for adding the post actions customization to settings. I know that will come later. On/off is good enough for now.

Great job as always! 🎉

P.S. I _love_ the swipe up/down to show/hide the FAB. I don't think I've seen that in any other app. It's so much better than auto showing/hiding on scroll.

CTalvio:
> On posts, dim the background when the additional actions are shown. Right now things are hard to read.

This is what I'm trying to get to work, the code is there, if you open the fab, and trigger a hot reload, you'll see it update the way it's supposed to.

lib/post/pages/post_page.dart
```diff
@@ -75,7 +87,17 @@ class _PostPageState extends State<PostPage> {
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
     final ThunderState thunderState = context.read<ThunderBloc>().state;
```
micahmo, line 89: Try `context.watch` here instead of `read`.

micahmo:
> This is what I'm trying to get to work, the code is there, but the UI won't update. If you open the fab, and trigger a hot reload, you'll see it update the way it's supposed to.

Got it. I think I found the problem. I left a comment on the relevant line of code.

CTalvio:
Alright! That works!

I'm down to merge this now, just so the framework for doing more with the new FAB is in place for anyone to contribute to. I'd like to look at some other stuff for a change so if this gets merged, feel free to build on what I've done so far.

As stated, the idea is to be able to choose whatever primary action you like most, among the ones available, and to swap it out on the fly by long tapping the options while the FAB is expanded.

I also think the swipe up gesture could trigger whatever secondary action is enabled, without the menu, if only one is enabled.

And I agree, whatever action is currently set as primary, should not be in the menu as well.

micahmo:
I would approve. I'll leave the final call to @hjiangsu in terms of whether he thinks this is mergeable as-is or whether the points I mentioned (or others) should be addressed first. I think this comes down to what standard we hold the nightlies to and whether we're ok with known inconsistencies/quirks. BTW, if you want to move on to something else, I'd be happy to finish up the points I mentioned.

hjiangsu:
Sorry it took me so long to go through this PR. It has a lot of changes which is why I tried to go through it as quickly as possible while trying to check for any potential issues!

I would just like to say this is pretty incredible and it seems like the implementation overall was well thought out!

There's just a couple of things I noticed which are really minor and I've noted them as comments. Before I go ahead and give my approval on this, I do want to open up a quick discussion on one architectural portion of the code and I wanna hear you inputs on this matter!

Currently, there are three variables in the thunder bloc which may be suited to be elsewhere: `dismissEvent`, `isFABOpen` and `isFABSummoned`. I feel like `dismissEvent` should reside in community_bloc and `isFABOpen`/`isFABSummoned` should in in the community_bloc/post_bloc.

Theres a few reasons why:
- The variables are more scoped if they are relocated to those places. For example, within the posts page, having `dismissEvent` doesn't make much sense since we can't dismiss posts there
- Having `isFABOpen`/`isFABSummoned` in their respective blocs allows the respective pages to individually handle the state of the FAB. This would most likely deal with the problem that @micahmo mentioned with regards to the additional actions being opened when switching from posts back to the feed

Anyways, those issues aside, I think this looks great! I'm fine with merging this in for the nightlies and just adding in fixes to the various issues as separate PRs before this is released to the general branch. Great job again @CTalvio!

hjiangsu (commented):
Here are the minor comments I had!

CTalvio:
> The variables are more scoped if they are relocated to those places. For example, within the posts page, having dismissEvent doesn't make much sense since we can't dismiss posts there

We will want to though. The idea is to eventually expand dismissEvent to be able to remove content from the feed when that content becomes blocked, wherever the user performs that action, without the need for a complete refresh.

Would moving the FAB summoned state to community/post prevent it from properly remembering whether the FAB is summoned? If I unsummon it, I want it gone, everywhere, until I summon it again, at which point I want it present, everywhere. Would it have its own summon state for each post or just all posts? Regardless you'd have to unsummon and summon it for posts and feeds separately, which is explicitly not what I wanted.

As for moving the open state, I see no issue, aside from that if you exit a post with it open, and then open another, I suspect the fab will then start off open?

micahmo:
> * Having `isFABOpen`/`isFABSummoned` in their respective blocs allows the respective pages to individually handle the state of the FAB. This would most likely deal with the problem that @micahmo mentioned with regards to the additional actions being opened when switching from posts back to the feed

Maybe `isFABOpen` should be scoped and `isFABSummoned` should be global?

hjiangsu:
>  If I unsummon it, I want it gone, everywhere, until I summon it again, at which point I want it present, everywhere. 

Ahh, okay in that case, then it makes sense to keep `isFABSummoned` globally. I do agree that maybe we can scope `isFABOpen` as @micahmo suggested.

> The idea is to eventually expand dismissEvent to be able to remove content from the feed when that content becomes blocked, wherever the user performs that action, without the need for a complete refresh.

Hmm, I think this can be done in other ways potentially, without needing to place `dismissEvent` in the thunder bloc. For example, if you are in the post page and you block some content, you can trigger an event in the community bloc (where the feed related info is located) and the community bloc can be in charge of refreshing the content without having to pass it all the way up to the thunder bloc. I did something similar here where if you upvote a post within a post page, the post card in the community feed will update without having to do a full refresh

I just want to be a bit more strict with adding variables to the thunder bloc because thats the "global" state so to speak, so its best if we keep the global state as slim as possible

hjiangsu:
Regardless, I think these things can be tackled in the future! If you can address those comments I mentioned and fix some of the conflicts, I'll gladly approve it and merge it in for the nightly!

CTalvio:
Done

hjiangsu:
Merging!

PR #501, opened 2023-07-30
Removed sliver of border color on root comments

ajsosa:
Removed sliver of border color that was present on root comments for both thick and thin style comments.


micahmo (approved):
LGTM!

PR #505, opened 2023-07-31
Replacing cards in UI

CTalvio:
image

Working on making user comment list, and inbox, match the UI style of the rest of the app. Will be looking at adding the functionality of buttons, modals, etc. directly to these views.

image

Might even look at getting saved to display both comments and posts, but that currently using the community list builder, so will be a bigger chore...


CTalvio:
This was a draft? Why is it merged?

PR #507, opened 2023-07-31
Add mod identifiers

micahmo:
This PR completes a previously unfinished feature related to mod identifiers (where we previously only identified admins, self, and the OP). This was harder because we didn't have the full community info within the post. Adding that will add slight overhead to post loading (one additional API call).

Also refactored the special user checks.

#### Mod + OP

| image |
| - |

### Admin + Mod + OP

Same as above but when logged into the instance that this user is an admin of

| image |
| - |

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -98,21 +98,24 @@ class PostBloc extends Bloc<PostEvent, PostState> {
           }
 
           PostViewMedia? postView = event.postView;
-          FullCommunityView? community;
+          List<CommunityModeratorView>? moderators;
 
           if (getPostResponse != null) {
             // Parse the posts and add in media information which is used elsewhere in the app
             List<PostViewMedia> posts = await parsePostViews([getPostResponse.postView]);
 
             postView = posts.first;
+
+            moderators = getPostResponse.moderators;
           }
 
-          // Get the full community in which the post resides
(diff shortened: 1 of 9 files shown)


hjiangsu:
I think it looks okay but I wonder if we can optimize this a bit more. From looking at the API, it seems like when we fetch a post (which returns back a `FullPostView`), it gives us a list of moderators which we can use to then to determine whether a comment is made from a moderator.

I don't think we need to change the current implementation, but its just something to consider in the future for optimization purposes!

https://github.com/thunder-app/thunder/blob/58ee5bd1f5d63647fe60c4eaeb1981c6024afee4/lib/post/bloc/post_bloc.dart#L92-L98

hjiangsu:
I think I'll leave it up to you to see if this is worth doing at this time - I'm okay with either or!

If we really want to get into this, we could even do a network check on which API call takes less time/less data (fetching communities vs fetching a full post) and then use that (although it feels overkill for this stage 😅)

hjiangsu (approved):

micahmo:
> I think I'll leave it up to you to see if this is worth doing at this time

Let me give it a quick shot and I'll let you know.

micahmo:
Ok, updated as suggested and refactored to only pass around the list of `moderators`. Let me know what you think!

hjiangsu:
Looks good!

PR #509, opened 2023-07-31
Improve sort picker

micahmo:
This PR improves the sort picker in a couple ways.

1. Adds a back button to the "top" submenu, so that you don't get stuck!
2. Shows all options in the "top" menu by default rather than requiring scrolling. (Scrolling is still possible on small screens.)

https://github.com/thunder-app/thunder/assets/7417301/06e0548b-8264-43df-aad9-9d40659d92e6


hjiangsu:
Looks good!

PR #512, opened 2023-07-31
Bottom comment scrolling improvements

micahmo:
This PR allows comments to be completely scrolled up to the "seems you've reached the bottom" message. This avoids a couple of issues.
 - The FABs could cover the bottom message.
 - Collapsing a comment could cause annoying rescrolling.
 - People may not like to read all the way to the bottom of the screen.

@hjiangsu, can you test this on iOS to see if the screen calculations look good there?

### Before

https://github.com/thunder-app/thunder/assets/7417301/21515fea-57e9-4330-b8af-4f3989b7b3c7

### After

https://github.com/thunder-app/thunder/assets/7417301/b2e65fc0-20f0-4933-a3b2-69b1d2019668

### DIFF
--- a/lib/post/widgets/comment_view.dart
+++ b/lib/post/widgets/comment_view.dart
@@ -9,6 +9,7 @@ import 'package:thunder/post/widgets/comment_card.dart';
 import 'package:thunder/core/models/comment_view_tree.dart';
 import 'package:thunder/post/widgets/post_view.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+import 'package:thunder/utils/screen.dart';
 
 import '../bloc/post_bloc.dart';
 
@@ -52,6 +53,7 @@ class CommentSubview extends StatefulWidget {
 }
 
 class _CommentSubviewState extends State<CommentSubview> with SingleTickerProviderStateMixin {
+  final GlobalKey _reachedEndKey = GlobalKey();
   Set collapsedCommentSet = {}; // Retains the collapsed state of any comments
   bool _animatingOut = false;
   bool _animatingIn = false;
@@ -156,6 +158,7 @@ class _CommentSubviewState extends State<CommentSubview> with SingleTickerProvid
                         crossAxisAlignment: CrossAxisAlignment.stretch,
                         children: [
                           Container(
+                            key: _reachedEndKey,
                             color: theme.dividerColor.withOpacity(0.1),
                             padding: const EdgeInsets.symmetric(vertical: 32.0),
                             child: Text(
@@ -164,6 +167,12 @@ class _CommentSubviewState extends State<CommentSubview> with SingleTickerProvid
                               textAlign: TextAlign.center,
                               style: theme.textTheme.titleSmall,
                             ),
+                          ),
+                          SizedBox(
+                            height: _reachedEndKey.currentContext?.findRenderObject() is RenderBox
+                                // Subtract the available screen height from the height of the "reached the bottom" widget, so that it's the only thing that shows
+                                ? getScreenHeightWithoutOs(context) - (_reachedEndKey.currentContext?.findRenderObject() as RenderBox).size.height
+                                : 0,
                           ),
                         ],
                       )
--- a/lib/utils/screen.dart
+++ b/lib/utils/screen.dart
@@ -1,0 +1,5 @@
+import 'package:flutter/material.dart';
+
+/// Returns the available screen height without including OS elements such as the status bar and the nav bar
+/// From: https://stackoverflow.com/a/71895304/4206279
+double getScreenHeightWithoutOs(BuildContext context) => MediaQuery.of(context).size.height - kToolbarHeight - MediaQuery.of(context).padding.top - kBottomNavigationBarHeight;


hjiangsu:
I can definitely test this out on iOS to see how it works! However, I do wonder if users will think this is a bug if we allow the scrolling to happen all the way to the top. Also, how would this work on a post with no comments or just a few comments?

(On a completely different note, I think it would be fun to add in an easter egg at the bottom of the comment thread for really long comment threads)

micahmo:
I totally get where you're coming from. The main thing I'm trying to address here is how the scrolling jumps when you collapse a comment near the bottom. We could add space just for the bottom message above the fab, but you could collapse comments with more height than that.

Here's how Relay does it. Maybe they add the height of the collapsed comment to the padding upon collapsing?

https://github.com/thunder-app/thunder/assets/7417301/66a156a8-1a02-42a9-b4fc-e877246b1548

I'd also be interested to hear what others think. @machinaeZER0?

machinaeZER0:
Heyyyyy. Definitely a good call to avoid the FAB obscuring content (good thinking!). Personally the current implementation works exactly how I'd expect it to (besides the FAB overlap) - if you're already at the bottom, collapsing a comment seems to re-size the view so that any content above is being pulled **down** into view, rather than still trying to roll up and expose more of the bottom of the screen. I like that the end text is always right at the bottom, unless there is so little content on screen right from the start that the "you've reached the bottom" doesn't actually reach the bottom (in which case, if it were possible I might not even display the "you've reached the bottom" message?

Let me know if that makes any sense - not sure if it's possible, but feels like the "natural" behavior to my brain :) my vote would just be to scroll up more if the FAB is present, really!

machinaeZER0:
I originally wrote a lot more than that and then realized the behavior I was describing was basically the current behavior 😆 glad I checked

micahmo:
Hey @machinaeZER0, thanks for your thoughtful response as always! I am totally fine with both your and @hjiangsu's point that showing too much whitespace at the bottom may seem very unusual to users.

However, I'd like to take one more stab at demonstrating the behavior that I don't like. If you still don't think it's an issue, I'm happy to drop this in favor of both of your opinions!

What I find jarring is that the behavior of tapping on a comment is inconsistent. When I'm in the middle of the comments, the content is pulled up. But once I get towards the, the content gets pulled down. This is very weird especially if (a) I still wanted to read the top-level comment (now I've lost my place) and (b) I don't know I'm near the bottom (since I may be collapsing more than fits on the screen at the time). If the collapsed content is partway off the screen, I could get a combination of pulling down and pushing up. And re-expanding the comment now pushes down the content (instead of going back up where it was). So I think it's all kind of inconsistent. 😆 Here's is an example. Pretend you're reading the comment when I collapse it and maybe that will help.

https://github.com/thunder-app/thunder/assets/7417301/1495e485-6bf7-4347-b000-255eb601241e

machinaeZER0:
Ahh, I see what you mean - I forgot that it will sort of "split the difference" if collapsing will reach the end of the page, even if the end of the page isn't in view yet! Would it be overly complicated if doing this one time (exposing the bottom of the screen) scrolled content up (creating that potential additional whitespace), but then after that any additional collapses (without first scrolling back up) would use the current behavior? In theory this would mean no "surprises" on first reaching the bottom through a collapse, but once you can see the bottom things "collapse upward" as they do now to not create a crazy amount of whitespace?

micahmo:
> Would it be overly complicated if doing this one time (exposing the bottom of the screen with a thread collapse) scrolled content up (creating that potential additional whitespace), but then after that any additional collapses (without first scrolling back up) would use the current behavior?

That's not a bad idea. It eliminates the most jarring behavior (splitting the difference). However, to me, having any kind of different behavior is jarring, especially when I'm trying to read the comment. I find the whitespace preferable to to losing my spot. And the whitespace would only be there if you collapsed a bunch of comments at the bottom.

Anyway, I don't want to impose my will or my expectations on here, and I think you're probably the best UX person we have, so I'm happy to defer.

machinaeZER0:
WellLllLllLl shucks, hahah. I try! I don't feel super strongly about the whole thing, though I guess I do enjoy the current behavior from a "minimizing whitespace" point of view, and it feels like (if it's possible/not a pain to implement) the proposed change would give people at least one "buffered" collapse toward the bottom before shifting content in the opposite direction.

I guess I'd ask if you feel like you/the typical user would be reading comments and also collapsing content at the same time? To me that feels like the main situation in which something jarring like this would occur, but given your most recent video I may not be thinking this through adequately... Presumably it's happened to you at least once or we wouldn't be here, so I think it's worth mulling over!

machinaeZER0:
Also HEY - didn't you come up with that absolutely gorgeous menu for swipe actions?? you're the real UX master around here, hahah

ajsosa:
@micahmo I wonder if it's possible to have that extra white space existing at the end to accommodate collapsing comments while still being able to limit the scrolling that far. Does the scroll controller provide that kind of control?

micahmo:
> I wonder if it's possible to have that extra white space existing at the end to accommodate collapsing comments while still being able to limit the scrolling that far.

I would have to look into it (might require some hacky workarounds), but I would like that.

micahmo:
I'm going to make this a draft for now. I want to look at the comment collapsing animation anyway since I think the performance can be improved, so maybe I'll tackle this then.

micahmo:
It looks like bottom scrolling is better now (maybe someone else improved elsewhere). The "reached the end" message scrolls up about halfway, which is good enough not to be covered by the FAB. The other solutions suggested here would be totally different than what I implemented, so I'm going to go ahead and close this PR.

PR #518, opened 2023-08-02
Adds additional localization strings to Thunder, and added temporary language files for Swedish/Finnish

hjiangsu:


micahmo (approved):
LGTM!

PR #520, opened 2023-08-02
Move image-to-comments button

micahmo:
This is a tiny change which moves the comments button to the right of the image viewer. Since this seems like common functionality, it's nice to have it closer to the thumb for right-handed users.

| image |
| - |


hjiangsu:
I agree with this as well!

PR #522, opened 2023-08-02
Add inkwell effect to comment usernames

micahmo:
This is a tiny UI improvement which adds the inkwell effect to the tappable usernames on comments. I think this is a nice pattern to have for any tappable element.

https://github.com/thunder-app/thunder/assets/7417301/907c8634-b7e5-4b79-9bc5-aa414dcf9012

### DIFF
--- a/lib/post/widgets/comment_header.dart
+++ b/lib/post/widgets/comment_header.dart
@@ -61,7 +61,7 @@ class CommentHeader extends StatelessWidget {
                   preferBelow: false,
                   child: Row(
                     children: [
-                      GestureDetector(
+                      InkWell(
                         onTap: () {
                           account_bloc.AccountBloc accountBloc = context.read<account_bloc.AccountBloc>();
                           AuthBloc authBloc = context.read<AuthBloc>();
--- a/lib/post/widgets/comment_header.dart
+++ b/lib/post/widgets/comment_header.dart
@@ -61,30 +61,31 @@ class CommentHeader extends StatelessWidget {
                   preferBelow: false,
                   child: Row(
                     children: [
-                      InkWell(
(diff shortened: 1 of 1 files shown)


hjiangsu:
Hmm, is there a way to round out the inkwell effect? It looks a bit jarring to see a square where most of the other UI elements are slightly rounded!

micahmo:
> adding a bit of padding on the sides of the inkwell so that it matches the padding that you have when you tap on the user with a chip

The reason I didn't do this is because it would add actual padding to the layout of the comment header, which I didn't want to change since we like how it looks now.

Note I also didn't add padding to the InkWell for tapping on usernames/communities in #482 for the same reason, and we were happy with that.

What do you think?

micahmo:
> I feel like I still personally prefer to have some padding around the inkwell

Sure, let's see how it looks! I think it can be done of of two ways.

1. Offset the username to allow for the padding. Personally, I don't think this looks good.

https://github.com/thunder-app/thunder/assets/7417301/d64b30b1-21ec-42ea-9f2a-aea4bf83e5dd

2. Offset the inkwell to allow for the padding. I think this looks better, but still a little funny that the inkwell goes to the left of the comment margin.

https://github.com/thunder-app/thunder/assets/7417301/4b61003e-56f4-489d-afc9-57f66b208741

What do you think? 1, 2, as-is, or something else?

hjiangsu:
I think 2 looks much better! I don't like the look of 1 either

micahmo:
Committed 2!

micahmo:
I just realized that this kind of messes up the inkwell on the comment. 😞

| image |
| - |

It doesn't seem to happen on posts, so I think it's fixable.

PR #523, opened 2023-08-02
Increase image load transition

micahmo:
This is a small fix which decreases the animation time once an image has finished loading. This should make the app overall feel more snappy.

### Before

https://github.com/thunder-app/thunder/assets/7417301/d2e897ad-ac43-43e8-bb4a-54b42e0f1e88

### After

https://github.com/thunder-app/thunder/assets/7417301/25073ff0-725c-4eec-a58e-a38191c368bc

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -111,7 +111,7 @@ class _ImagePreviewState extends State<ImagePreview> {
             width: widget.width ?? MediaQuery.of(context).size.width - 24,
             fit: BoxFit.cover,
             cache: true,
-            clearMemoryCacheIfFailed: false,
+            clearMemoryCacheWhenDispose: false,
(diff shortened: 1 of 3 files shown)


lib/shared/image_preview.dart
```diff
@@ -111,7 +111,7 @@ class _ImagePreviewState extends State<ImagePreview> {
             width: widget.width ?? MediaQuery.of(context).size.width - 24,
             fit: BoxFit.cover,
             cache: true,
-            clearMemoryCacheIfFailed: false,
+            clearMemoryCacheWhenDispose: false,
```
micahmo, line 114: This fixes an unintended change from here:

https://github.com/thunder-app/thunder/pull/463/files#diff-f771bff454f9ff659025f245c61fbecc7e01a6be62d0ef4c21ada798dd536851R103

hjiangsu:
I'll approve it but again will have to double check if this has any impact on performance through some profiling!

Issue #517, opened 2023-08-01
Accessibility Regressions


shortwavesurfer2009:
**Description**  
The following are not read or no longer read with talkback

**How to Reproduce**  
1. Go to System Settings > Accessibility > Talkback and turn it on
2. Open Thunder
3. Tap the items indicated

**Expected Behavior**  
should br read by talkback

**Screenshots**  
- [x] ![Screenshot_20230801-164726_Thunder_1_1](https://github.com/thunder-app/thunder/assets/116814522/c8c07dc3-876b-4c6a-b98b-b05023271d63)
- [x] ![Screenshot_20230801-170535_Thunder](https://github.com/thunder-app/thunder/assets/116814522/152aace4-f56a-47a5-b1e2-015a4875ab52)
- [x] ![Screenshot_20230801-170603_Thunder_1](https://github.com/thunder-app/thunder/assets/116814522/da63bfe5-8b46-45bd-a4a9-8ff918bad293)
- [x] ![Screenshot_20230801-172228_Thunder](https://github.com/thunder-app/thunder/assets/116814522/b8f734aa-d45c-45e2-9a13-d0b387f41b95)

**Device & App Version:**  
 - Device: [e.g. iPhone6] Oneplus Nord N200
 - OS: [e.g. iOS8.1] Android 13
 - Version [e.g. 22] 0.2.1-alpha.1+14

**Additional Context**

hjiangsu:
I'll go ahead and close this!

PR #524, opened 2023-08-02
Support accessibility for new gesture picker

micahmo:
Fixing my part of #517.

@shortwavesurfer2009 Maybe you can let me know if these descriptions are good.


shortwavesurfer2009:
`i think "long "long swipe right", "short swipe right", "short swipe left", and "long swipe left" would be better. So an example would read "short swipe right, downvote" or whatever it was currently set to.

micahmo:
@shortwavesurfer2009 Thanks! Updated as you suggested.

shortwavesurfer2009:
Perfect. I just had a look through the commit and that is perfect

hjiangsu:
LGTM :D

PR #532, opened 2023-08-02
Fixed another infinit spin for comment loading.

ajsosa:
Fixed another infinite spin for comment loading edge case. Also we no longer keep repeatedly spamming `GetPostCommentsEvent()` when comment thread is marked as being at the end.


hjiangsu:
🎉

PR #533, opened 2023-08-02
Fix location of PR template

micahmo:
I think I did this wrong. The template file is only supposed to be in a subfolder if you want to use multiple template and reference them by query string. Otherwise it should be in the root of `.github`. 😕

https://stackoverflow.com/a/62613198/4206279


hjiangsu (approved):

micahmo:
Cool, I see the template when creating a PR now! 🎉 I'm going to bring this back to `develop` just for consistency.

Issue #510, opened 2023-07-31
Add font scale options for more elements


sandpaperback:
There are currently font scale options only for Title and Content. Allowing additional elements to be controlled independently will help provide users with finer control over their experience and also improve accessibility.

The biggest piece would be for metadata/interface to scale separately from content. Right now, if a user wants to minimize the space taken up in their timeline by user, community, etc, that scale setting also impacts the font scale for all comments and post detail. It can result in a less than comfortable experience.

Attached are some images to illustrate the idea.

![Screenshot_20230731_081605](https://github.com/thunder-app/thunder/assets/104503487/4c83f46d-e201-494a-8243-26d721534f01)
![Screenshot_20230731_082555](https://github.com/thunder-app/thunder/assets/104503487/d4a884bd-1e22-4ab8-b930-47518e63c421)
![Screenshot_20230731_081701](https://github.com/thunder-app/thunder/assets/104503487/5a65cac9-ff56-439a-ae6b-b70f33e8ee7c)
![Screenshot_20230731_081707](https://github.com/thunder-app/thunder/assets/104503487/1ae8f7e9-239c-4abd-b84c-8200f0ba9bb6)

hjiangsu:
Thanks for creating the issue and the screenshots! That is really useful for us to be able to tackle this on.

I agree that we should have some more granularity in selecting the font sizes for different aspects of the UI. The reasoning for why there are only two options at the moment is to allow a bit more accessibility for those who require larger font sizes in order to use the app. Now that we have that foundation in place, we can start adding in more granularity to all the elements you've mentioned!

hjiangsu:
One thought here: should "content font scale" and "detail content font scale" be the same thing?

I think these should be the options:
- **Title font scale**: handles the title of the post from the feed, and when you are viewing the post itself (same as what you have)
- Rename **Content font scale** -> **Post font scale** or **Post detail font scale**: this would handle the text preview in the feed, as well as the body of the post (this combines content font scale and detail content font scale from your image)
- **Comment font scale**: handles the text within comments (same as what you've described)
- **Metadata font scale**: handles any metadata text (same as what you've described)

@sandpaperback Let me know what you think!

hjiangsu:
Added the options I mentioned in my previous comment, this should be available in the upcoming pre-release

sandpaperback:
Sorry, started several replies and never followed through. Looks good and I'm looking forward to the next release!

hjiangsu:
[...]

Issue #301, opened 2023-07-09
[backlog] OS-specific font scaling


machinaeZER0:
Not high priority, but per the discussion and resolution of issue #59 I'm creating this ticket as a backlog item. It seems iOS and Android handle font scaling differently, so currently (to my understanding) the unified code in Thunder is setting font size in a way to achieve a sort of "happy medium" across devices. In practice, this is currently set up in a way where the base font size on Android is slightly smaller than it was in earlier releases, so that iOS fonts aren't displayed as oddly large.

At some point in the future I believe the goal is to implement dedicated, OS-specific font handling "stuff", which would address this small issue :)

hjiangsu:
Finally working on this!

machinaeZER0:
Oh dang! At this point I'm used to it, hahah. A big win for consistency though!

micahmo:
Yeah that was my concern as well. The font size looks great on Android currently, so I was hoping it wouldn't change much. Looks like most of the changes are for iOS.

machinaeZER0:
Yeah, @hjiangsu's initial workaround set the scaling to .95 I believe (to make iOS look more normal), so if that's changing back to whatever 1 is on Android it shouldn't be too jarring a change! I guess the good news is that if it looks off we'll be able to tweak without worrying about affecting both OS views anymore :)

PR #543, opened 2023-08-03
Platform specific font scaling and more font scaling options

hjiangsu:
## Pull Request Description

Creates a way for font sizes to be scaled according to the platform. This is because it seems like there are differences in the way fonts are scaled between iOS and Android

This also adds in two new font scaling options for more granularity:
- Comment font scaling
- Metadata font scaling

The different font scalings are applied closely to match #510's description

It would be good for anyone on Android to test out the changes to see if the font scaling sizes are appropriate!

## Issue Being Fixed

Issue Number: #301, #510

## Screenshots / Recordings
simulator_screenshot_40D6A831-B0BA-4D28-9D9F-4BEB40E29244

N/A

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Added additional localization strings to Thunder, and added temporary language files for Swedish/Finnish
 - Added manual refreshing to the user account page - contribution from @micahmo 
 - Added inkwell effect when tapping on usernames in comments - contribution from @micahmo 
+- Added additional font scaling options for comments and metadata
 
 ### Changed
 - Removed tap zones for author/community on compact post cards - contribution from @CTalvio
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Snappier image load transition - contribution from @micahmo 
 - Align back button in image preview with the back button in the main pages - contribution from @micahmo 
 - Moved location of comment button within image preview - contribution from @micahmo 
+- Adjusted font scaling to be platform specific
 
 ### Fixed
 - Fixed issue where the community post feed was missing the last post - contribution from @ajsosa
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -3,20 +3,18 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
-import 'package:flutter_markdown/flutter_markdown.dart';
 import 'package:lemmy_api_client/v3.dart';
-import 'package:thunder/account/bloc/account_bloc.dart';
 
+import 'package:thunder/account/bloc/account_bloc.dart';
+import 'package:thunder/core/enums/font_scale.dart';
 import 'package:thunder/community/bloc/community_bloc.dart';
 import 'package:thunder/community/widgets/community_header.dart';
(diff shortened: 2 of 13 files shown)


micahmo (approved):
@hjiangsu I'd like to test this on Android (which will be later today) if that's ok, but otherwise approved!

hjiangsu:
@micahmo I've added a bit more to this PR 😅 - thought the two issues were somewhat linked so decided to do them together

machinaeZER0:
Sorry, I still don't have a test environment set up - would you clarify if the new defaults will still look basically the same as they do now? It looks at a glance like the font size is consistent in a given post, just that the title is bolded. Is that correct?

As an aside, one thing I'm so thankful for in Thunder (a pet peeve with some other apps) is that oftentimes adjusting the font size in an app doesn't tweak line spacing appropriately to accommodate, which often means bigger or smaller text will look weirdly cramped or floating in a sea of empty space. I feel like Thunder really nails font spacing/content padding in a lovely way, and I really appreciate that!

hjiangsu:
>  would you clarify if the new defaults will still look basically the same as they do now (albeit slightly larger overall on Android)?

Yeah, I think everything should look almost the same, although it'll be ever so slightly larger on Android (1.0 scale vs 0.95 from before) Most of the changes should happen on the iOS side of things, where everything will be slightly smaller!

I'll definitely add this as part of a pre-release for people to test out and see if the scaling should be adjusted more!

micahmo:
Hmm, the text is quite a bit larger on Android now. I mean, not to a crazy extent, but having gotten used to the size it was before, I'm not sure I like this. 😅

https://github.com/thunder-app/thunder/assets/7417301/bb0418fa-ca39-4770-8dfe-599bfaad6533

https://github.com/thunder-app/thunder/assets/7417301/802e2b24-6d88-4f73-96da-0a97df169f06

machinaeZER0:
I don't mind it from what I'm seeing, but it is noticeably different for sure (took a while for me to get used to it being smaller, which is why I filed the ticket all those weeks ago). 

I wouldn't want to get toooo granular (especially if it's a pain to code), but instead of having a couple of sizes to choose from, we could let people choose a multiplier value on a slider or something? Then people who are used to .95 would be able to choose it (where I feel like the next smallest size would be a larger jump than that in the current setup). 

Do you think it's something you could get used to, potentially? It should be pulling the base size from your phone settings...

hjiangsu:
Dang, I wonder if we can get an older build to compare with from before the text scaling was added in.

I could just revert the text scale down to the same value as before and that should solve the issue again 😅

micahmo:
> I could just revert the text scale down to the same value as before

I think that would satisfy those who have gotten used to the sizing of things in Thunder. Honestly the sizing being so nice out of the box is one of the things that immediately grabbed me about Thunder. Liftoff is too big, Sync is too small, etc. Obviously we'll have customizations for fine-grained control, but I like the previous values as the defaults. (Just my opinion.)

machinaeZER0:
I still have screenshots from when it changed initially, but really what you're seeing now should effectively match what it was back then:





At this point I've definitely gotten used to .95 on android, and I don't know that anyone's really complained since adding in size options (beyond maybe wanting more of them, which this is already addressing). If Micah likes .95 I'm personally more than happy if it stays that way :) just my two cents!

micahmo:
@hjiangsu Yep, looks perfect to me now! Thanks for that! I guess these kinds of tweaks are fine now that they're platform-specific. 😊

PR #545, opened 2023-08-04
Fix swipe gesture to go back on iOS due to WillPopScope

hjiangsu:
## Pull Request Description

Fixes an issue where the swipe gesture to go back would not work on iOS (and maybe Android) because WillPopScope issue: https://github.com/flutter/flutter/issues/14203

This should also be checked on Android to see if swipe to go back works properly

(review with whitespace removed)

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -57,6 +57,11 @@ class _PostPageState extends State<PostPage> {
   @override
   void dispose() {
     _scrollController.dispose();
+
+    if (context.read<ThunderBloc>().state.isFabOpen) {
+      context.read<ThunderBloc>().add(const OnFabToggle(false));
+    }
+
     super.dispose();
   }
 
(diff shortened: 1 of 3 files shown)


lib/post/pages/post_page.dart
```diff
@@ -57,6 +57,11 @@ class _PostPageState extends State<PostPage> {
   @override
   void dispose() {
     _scrollController.dispose();
+
+    if (context.read<ThunderBloc>().state.isFabOpen) {
+      context.read<ThunderBloc>().add(const OnFabToggle(false));
```
micahmo, line 62: Hmm, this doesn't seem to work. The problem of the FAB staying open when navigating back from a post is now happening again (whether I press the UI back button, the OS back button, or swipe).

hjiangsu:
I'll close this since it seems like more thought will need to be placed here

PR #553, opened 2023-08-04
Experimental: Post & Comments are now kept alive

ajsosa:
This is an experimental change that improves scrolling on a post's view. The post and comments are all set to be kept alive even when off screen. This improved performance by not having to rebuild layout constantly at the cost of memory. Also addresses the jittery scrolling when a post has a large image.

~~Additionally I removed the no longer needed scroll listener that was used for the original FAB. This removes the state changes that would occur at the end of the comment section which also caused jank frames.~~

Please don't merge yet. I want people to test this out and get feedback.


ajsosa:
This has essentially become an experimental branch that is exploring ways of improving performance to the comment list. NOT TO BE MERGED!!!

hjiangsu:
Closing this as it's no longer relevant!

PR #554, opened 2023-08-04
Removed no longer used scroll listener.

ajsosa:
Removed no longer used scroll listener which was triggering state changes at the end of the comment section.


micahmo (approved):
Nice!

Issue #527, opened 2023-08-02
Accessibility: long-press to expand FABs


shortwavesurfer2009:
The FABs should be expandable with a long press as swipes and screen readers such as talkback do not work well together. When the FAB expands the focus for screen readers should be placed on the top most item, followed by the next down, and so on. The very bottom option needs to be "close" or "collapse"
![Screenshot_20230802-103720_Thunder](https://github.com/thunder-app/thunder/assets/116814522/9fadd2d9-ae78-4f18-86d0-e56495635e05)

hjiangsu:
I think the original plan here was to allow for a long press to trigger a secondary action in the future. I think we can maybe accommodate both use-cases by doing this:

Introduce a new setting: Long press action on FAB

The available options for these new settings would be:
- Expand/collapse FAB which would fulfill this accessibility issue
- Trigger secondary action (in which case, you could choose your secondary action from a list of actions)

shortwavesurfer2009:
> I think the original plan here was to allow for a long press to trigger a secondary action in the future. I think we can maybe accommodate both use-cases by doing this:
> 
> Introduce a new setting: Long press action on FAB
> 
> The available options for these new settings would be:
> 
>     * Expand/collapse FAB which would fulfill this accessibility issue
> 
>     * Trigger secondary action (in which case, you could choose your secondary action from a list of actions)

Seems fine to me. What would the default be set to?

CTalvio:
Default action could be either nothing, or expand. The long press won't have anything communicating what it does to most users, so should only trigger a secondary action for users who manually set one, and hence know what it will do.

shortwavesurfer2009:
> Default action could be either nothing, or expand. The long press won't have anything communicating what it does to most users, so should only trigger a secondary action for users who manually set one, and hence know what it will do.

Personally i think it should be set to expand by default as this would best help for accessibility.

PR #557, opened 2023-08-05
Feature/fab customization

micahmo:
## Pull Request Description

This PR adds support for customizable short-press and long-press actions on both the feed and post FABs.

There are a ton of possible configurations now (my video just demonstrates a few), but I did test as well as I could, despite my demo only showing a few options.

I also think some of the generalization could help us get towards reorderable actions.

This PR also introduces inkwells for settings and the ability to click anywhere on the row to change a toggle.

## Issue Being Fixed

Issue Number: #527 (maybe others)

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/caa99a85-271b-415a-b1bb-e9fbebc33973

## Checklist

- [ ] Did you use localized strings where applicable? -- No, the settings don't seem to have been converted yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Added manual refreshing to the user account page - contribution from @micahmo 
 - Added inkwell effect when tapping on usernames in comments - contribution from @micahmo 
 - Long-pressing on FAB shows extended actions - contribution from @micahmo
+- Added support for customziable short-press and long-press FAB actions - contribution from @micahmo
 
 ### Changed
 - Removed tap zones for author/community on compact post cards - contribution from @CTalvio
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -19,6 +19,7 @@ import 'package:thunder/community/pages/create_post_page.dart';
 import 'package:thunder/community/widgets/community_drawer.dart';
 import 'package:thunder/community/widgets/post_card_list.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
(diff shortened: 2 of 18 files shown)


hjiangsu:
I think it looks pretty good overall! Just one point of feedback (that we can do in the future as a separate PR maybe):
- Instead of using the local settings string throughout the code to determine the FAB action, we can create a new enum for FAB actions (FABAction/FabAction or some other name) which would work similarly to SwipeAction enum
  - When setting the setting for the single/long press actions, we would save the name of the FABAction enum name
  - We can then translate the value back from a string (stored in SharePreferences) to a FABAction enum which can then be used everywhere else in the code

You can take a look at how SwipeAction is handled throughout the codebase to get examples! Here is an example of how swipe action is translated from the string to the enum:
```dart
SwipeAction leftPrimaryPostGesture = SwipeAction.values.byName(prefs.getString(LocalSettings.postGestureLeftPrimary.name) ?? SwipeAction.upvote.name);
```

Doing this would then allow us to extend the FAB action enum (e.g., to be able to fetch the associated icon) and keep things a bit cleaner! Anyways, let me know what you think about this. We can still merge this in first and then potentially apply the feedback I mentioned if you'd like!

micahmo:
I agree, that would be an big improvement. Like the `SwipeActions` refactoring I did, it would make things more object-oriented and maintainable going forward. I'll see if I can knock it out.

micahmo:
Alright, I think this is ready to go!

hjiangsu:
Nice! It looks good to me, thanks for adding the improvements!

hjiangsu (approved):

PR #561, opened 2023-08-06
Deferred comment indicator improvements

micahmo:
## Pull Request Description

This PR adds a few QoL improvements to the deferred comment indicator.

1. Localize and handle pluralization.
2. Lighten color so it's not mistaken for a comment.
3. Add InkWell effect.
4. Better handling of errors.
   - Show toast for errors.
   - Don't show spinner forever.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/c6b281e5-c59f-4fe4-856b-e48693213c66

https://github.com/thunder-app/thunder/assets/7417301/41d92504-0dfc-4e9b-9315-622a96926209

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -220,7 +220,6 @@ class _PostPageState extends State<PostPage> {
                                 ],
                               ),
                               backgroundColor: theme.colorScheme.onErrorContainer,
-                              behavior: SnackBarBehavior.floating,
                             );
                             WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
                               ScaffoldMessenger.of(context).clearSnackBars();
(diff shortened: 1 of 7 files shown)


lib/post/pages/post_page.dart
ajsosa, line 1: Are we moving away from floating toasts? If that's the case, we should probably make them `fixed` across the board so it's consistent. Thoughts?

micahmo: > Are we moving away from floating toasts?

Yes I think so. In this case, it didn't appear until I removed floating. Also the non-floating ones automatically bump the FAB.

> If that's the case, we should probably make them `fixed` across the board so it's consistent. Thoughts?

I agree. @hjiangsu suggested we do an overall snackbar refactor so we don't have these same lines duplicated everywhere. Then we could have one place that controls the style.

hjiangsu: Yeah, I think we can do a refactor of the snackbar stuff in another PR to consolidate the codebase and to keep things consistent for all toast notifications!

hjiangsu (approved):
I think it looks good!

micahmo:
Resolved conflicts!

PR #569, opened 2023-08-07
Prioritize OP

micahmo:
## Pull Request Description

This is a tiny UI fix which prioritizes (color, icon) of the Original Poster over being self, admin, or mod. I believe the OP identification should always be consistent and not change based on the person's role, and I have found myself often getting confused that this is not the case.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| image |
| - |

## Checklist


machinaeZER0:
I agree with this change - I like theming (and would love more uses for the selected accent color), but consistency helps ground the user in their experience, and having important information persist across theme changes helps a lot with that!

micahmo:
> I like theming (and would love more uses for the selected accent color), but consistency helps ground the user in their experience, and having important information persist across theme changes helps a lot with that!

@machinaeZER0 This actually isn't related to theming. (The screenshot I posted is outdated now due to other theming changes.) Rather this is related to which identifier is prioritized when a user falls into multiple categories. For example, in the screenshot I showed, the user is both the OP, an admin, and a mod. Prior to my change, the admin badge would be first, and the chip would be the admin color. However, I was finding it hard to remember who the OP was. The best example of this was in the Lemmy dev asklemmy. Both of the devs were in there, but one was the OP. However, they were the same color. So I thought it might help if the OP were consistently themed, regardless of their other roles.

machinaeZER0:
Ohhh, gotcha - that makes sense too! OP taking "role precedence" sounds right to me :) Sorry for misunderstanding!

PR #571, opened 2023-08-07
Refactor snackbars

micahmo:
## Pull Request Description

This PR is just for code cleanup. It refactors all snackbar usage into a single place, with various options for different usages. I also localized any snackbar text strings that were not already.

## Issue Being Fixed

Issue Number: N/A -- discussed in a few places

## Checklist

- [ ] Did you update CHANGELOG.md? -- N/A, just a code thing


lib/account/pages/login_page.dart
```diff
@@ -129,27 +131,11 @@ class _LoginPageState extends State<LoginPage> {
...
-              SnackBar snackBar = SnackBar(
-                content: Text('Login failed, please try again (${state.errorMessage ?? 'Unknown'})'),
-                behavior: SnackBarBehavior.floating,
-              );
-
-              WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-                ScaffoldMessenger.of(context).clearSnackBars();
-                ScaffoldMessenger.of(context).showSnackBar(snackBar);
-              });
+              showSnackbar(context, AppLocalizations.of(context)!.loginFailed(state.errorMessage ?? 'Unknown'));
```
hjiangsu, line 134: Maybe we can use `missingErrorMessage` localization message for the default here

lib/community/pages/community_page.dart
```diff
@@ -122,44 +123,17 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
...
-                        ),
-                      ),
-                    ),
-                ],
-              ),
-              behavior: SnackBarBehavior.floating,
-            );
-            WidgetsBinding.instance.addPostFrameCallback((timeStamp) => ScaffoldMessenger.of(context).showSnackBar(snackBar));
+            if (state.blockedCommunity?.blocked == true) {
+              showSnackbar(context, AppLocalizations.of(context)!.successfullyBlockedCommunity(state.blockedCommunity?.communityView.community.title ?? 'UNKNOWN'), undoAction: () {
```
hjiangsu, line 131: Maybe we can use `missingErrorMessage` localization message for the default here too!

lib/community/pages/community_page.dart
```diff
@@ -122,44 +123,17 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
...
+                context.read<CommunityBloc>().add(BlockCommunityEvent(communityId: state.blockedCommunity!.communityView.community.id, block: false));
+              });
+            } else {
+              showSnackbar(context, AppLocalizations.of(context)!.successfullyUnblockedCommunity(state.blockedCommunity?.communityView.community.title ?? 'UNKNOWN'));
```
hjiangsu, line 135: Same thing here with `missingErrorMessage` localization message

lib/shared/snackbar.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:flutter/material.dart';
+
+enum SnackBarMode { normal, warning }
+
+void showSnackbar(BuildContext context, String text, {SnackBarMode mode = SnackBarMode.normal, bool clearSnackBars = true, Duration? duration, void Function()? undoAction}) {
```
hjiangsu, line 5: A couple of things!

1. `undoAction` could just be renamed to `action`, and we could allow for a custom icon for the `action` to make it more generic. Or an alternative could be to rename it to `trailing` and make it accept a `Widget` similar to how `AppBar` does it
3. I think I prefer `type` instead of `mode`. Then for the enum, it could be 

```dart
enum SnackBarType { info, warning }
```

lib/user/pages/user_settings_page.dart
```diff
@@ -34,78 +36,33 @@ class _UserSettingsPageState extends State<UserSettingsPage> {
...
-
-              WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-                ScaffoldMessenger.of(context).clearSnackBars();
-                ScaffoldMessenger.of(context).showSnackBar(snackBar);
-              });
+              if (state.status == UserSettingsStatus.failure) {
+                showSnackbar(
+                    context,
+                    state.status == UserSettingsStatus.failure
+                        ? AppLocalizations.of(context)!.failedToUnblock(state.errorMessage ?? 'UNKNOWN')
```
hjiangsu, line 43: See previous comments about using `missingErrorMessage`!

hjiangsu (commented):
I think it's pretty good overall, just a couple of points!

lib/shared/snackbar.dart
micahmo: Good points. This could be made super generic by accepting a leading and a trailing widget. But I don't want the caller to do _too_ much work. 😆 So maybe a good compromise would be a leading icon, a trailing icon, and an action. That would also remove the need for the enum.

hjiangsu (approved):
LGTM!

PR #573, opened 2023-08-07
Fix Post Title Font Sizing

hjiangsu:
## Pull Request Description

With the new posts look, there was a change to the post title which made it much smaller on iOS devices (unsure about Android devices). This PR addresses this and attempts to adjust the font to match what it currently is.

Would need to double check on Android side of things to make sure the post title text is still roughly the same as before

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
LGTM! No issues on Android.

PR #574, opened 2023-08-07
Add code-based version number and CI to validate

micahmo:
## Pull Request Description

While working on #478, we discovered/remembered that iOS does not have access to the full version/prerelease/build number. My idea to combat this problem is to put the full version number directly in code where it can be consistently checked. The downside to this approach is that it's easy to forget to update the code in addition to `pubspec.yaml`. Therefore, this PR also introduces an addition to the build pipeline which verifies that the version number specified in the code always matches the pubspec.

## Issue Being Fixed

Related to PR #478.

## Demo

Here are examples of the CI failing and passing, respectively.
* https://github.com/micahmo/thunder/actions/runs/5788122874/job/15686441460#step:13:10
* https://github.com/micahmo/thunder/actions/runs/5788126185/job/15686453480#step:13:3


micahmo:
Looks like the [flutter-version-number-action](https://github.com/marketplace/actions/flutter-version-number) workflow is broken since we're using a package called `version` and it's getting confused. 🤦

micahmo:
Ok I "fixed" it by putting a space on the package line. Crazy but it works lol. Now the CI is failing for the right reason.

image

hjiangsu:
Nice find! I'll go ahead and merge it in

hjiangsu (approved):

micahmo:
Awesome! Keep in mind that the CI is not (correctly) failing because the versions don't match. Not sure if you're planning on fixing that or if you want me to.

hjiangsu:
Just fixed it :D - should no longer be failing (at least on develop)

PR #576, opened 2023-08-07
set memory cache to be cleared when image is disposed to improve memo…

hjiangsu:
## Pull Request Description

Sets the image cache to be cleared from memory when the image is disposed. This will help reduce memory consumption a little bit by allowing the RAM to clear out images when not needed. I have increased the cacheExtent previously to help accommodate for the issue where we see a lot of images being reloaded

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -111,7 +111,7 @@ class _ImagePreviewState extends State<ImagePreview> {
             width: widget.width ?? MediaQuery.of(context).size.width - 24,
             fit: BoxFit.cover,
             cache: true,
-            clearMemoryCacheWhenDispose: false,
+            clearMemoryCacheWhenDispose: true,
             cacheWidth: ((MediaQuery.of(context).size.width - 24) * View.of(context).devicePixelRatio.ceil()).toInt(),
           ),
           TweenAnimationBuilder<double>(
--- a/lib/shared/image_viewer.dart
+++ b/lib/shared/image_viewer.dart
@@ -233,7 +233,7 @@ class _ImageViewerState extends State<ImageViewer> with TickerProviderStateMixin
                             mode: ExtendedImageMode.gesture,
                             extendedImageGestureKey: gestureKey,
                             cache: true,
-                            clearMemoryCacheWhenDispose: false,
+                            clearMemoryCacheWhenDispose: true,
                             initGestureConfigHandler: (ExtendedImageState state) {
                               return GestureConfig(
                                 minScale: 0.8,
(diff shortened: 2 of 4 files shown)


hjiangsu:
@micahmo Can you do some tests to see if this is behaving okay (e.g., images are not reloading when scrolling back up)

micahmo:
The images are flushed from cache pretty quickly now. I don't have to scroll far for them to need reloading. 😞

https://github.com/thunder-app/thunder/assets/7417301/e67623ff-0b09-4240-a62e-b2e7a730231d

hjiangsu:
Hmm, do you think it's okay? At least when scrolling through the feed normally, it the loading indicators should be less noticeable

micahmo:
I just tried a bunch of the other apps, and none of them seem to evict their cache like this. Once a post in the feed is loaded, I can scroll way down and come back and its image is already there. But if we're trying to squeeze out better memory performance, this is fine. I actually think removing the loading indicator (and having a dissolve) would go a long way to making it look less like things are constantly loading.

micahmo (approved):

hjiangsu:
I think this is a temporary fix until we figure out how to limit the cache size and also keep memory usage within check. I do agree that maybe a dissolve would make it less like its loading

hjiangsu:
One more thing @micahmo - could you try running it on your actual device or on a profile/release mode? I'm doing that right now and I don't get a lot of loading indicators when scrolling through the feed

micahmo:
Tried it on my physical device in profile mode. I would say it's about the same, although it doesn't "feel" as bad because the loading is much faster.

https://github.com/thunder-app/thunder/assets/7417301/728f5d3f-f944-4d42-b92e-bab3f0eb839a

hjiangsu:
I guess it could be phone dependent. We can continue to tweak it after! I'll merge this in otherwise if you're okay with it :D 

I'll try to upload a recording from my phone to show you how it looks like

hjiangsu:
https://github.com/thunder-app/thunder/assets/30667958/1cafd717-2d2f-4736-a1e1-968034747698

micahmo:
> I'll merge this in otherwise if you're okay with it :D

Yep, go for it!

micahmo:
> trim.D5B8738E-216C-401E-95C4-DC8AEFC1B32F.MOV

Wow that looks great and really smooth too. I think we're missing out. 😆

Also I noticed some of the preview progress indicators use the native iOS looking one and some use the more Android looking one. My guess is the difference is between link previews and image previews. Again, I think it would be nice to abandon these spinners altogether (in this case).

hjiangsu:
> Wow that looks great and really smooth too. I think we're missing out. 😆

I think part of it is the high refresh rate thats enabled by default for iOS (but not Android for some reason)

> Also I noticed some of the preview progress indicators use the native iOS looking one and some use the more Android looking one. 

These are just differences in which image preview is loading (LinkPreviewGenerator vs the image fetching logic that we have before)!

PR #577, opened 2023-08-07
Font text scale now respects system font size and scales against the system size

hjiangsu:
## Pull Request Description

Adjusted the font sizing again, this time to respect the system's font size and scale it according to the systems set font size. For example, if you set your system font size to large, then the scaling factor will scale according to that size.

**Example:**
System Scale Factor: 1.2
Thunder Text Scale Factor: 0.95
Final Text Scale Factor: 1.2 * 0.95

Again, this should be checked in Android to ensure that its still somewhat consistent!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
LGTM on Android!

micahmo (approved):

Issue #580, opened 2023-08-08
Accessibility: long-press FAB menu foes not get focus with talkback


shortwavesurfer2009:
**Description**  
when the FAB is long pressed and the options appears the talkback focus should be pulled to the first element ("back to top by default)

**How to Reproduce**  
1. Go to system settings > accessibility > talkback and turn it on
2. Open thunder
3. Double tap and hold FAB to open its options

**Expected Behavior**  
Talkback focus should be pulled to the top item in the list of actions

**Screenshots**  

**Device & App Version:**  
 - Device: [e.g. iPhone6] OnePlus Nord N200
 - OS: [e.g. iOS8.1] Android 13
 - Version [e.g. 22] v0.2.1-alpha.2+14

**Additional Context**

PR #581, opened 2023-08-08
Accessibility: Focus top FAB extended button

micahmo:
## Pull Request Description

1. Focus the top extended FAB button when the FAB is opened via TalkBack.
2. Remove extraneous `semanticLabel`s that were causing duplicate descriptions to be read (because TalkBack also finds the `Text` labels).

## Issue Being Fixed

Issue Number: #580

## Screenshots / Recordings

N/A

## Checklist

N/A

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -280,7 +280,6 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
                                           title: FeedFabAction.dismissRead.getTitle(context),
                                           icon: Icon(
                                             FeedFabAction.dismissRead.getIcon(),
-                                            semanticLabel: FeedFabAction.dismissRead.getTitle(context),
                                           ),
                                         ),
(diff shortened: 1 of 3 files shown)


hjiangsu (approved):
LGTM!

shortwavesurfer2009:
How was this done? The symantic label should be on the button itself so as to not be read as "unlabeled" or "button. If theactual text strings are causing dups then the text strings should be skipped

shortwavesurfer2009:
Well, we do not want the text label and the button being seperate elements for sure otherwise element 1 would say "back to top" and be the text. Element 2 would say "back to top" and be the button. Either the button only needs to be read and skip the text element or the button and text element need to be put together into a single element. Let's see how this chsnge goes in alpha.3+14 and if it needs to be changed we can address it at that time.

micahmo:
Yes, please let me know! I will be happy to make the change if needed. Based on my testing I hope it'll just work as is, but of course we will try to make it right regardless.

PR #592, opened 2023-08-09
Fix animation issues with dismiss read

CTalvio:
## Pull Request Description

Fixes dismiss animation often failing to play when used after many new posts have been marked read. Also dismiss is now staggered. Cool!

## Screenshots / Recordings

[Kazam_screencast_00002.webm](https://github.com/thunder-app/thunder/assets/4365015/a9586998-0e5d-4268-8024-3a540e8172d4)

## Checklist
N/A

### DIFF
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -63,7 +63,8 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
   bool _showReturnToTopButton = false;
   int _previousScrollId = 0;
   bool disableFabs = false;
-  bool showRead = true;
+
+  Set toRemoveSet = {};
 
   late final AnimationController _controller = AnimationController(
     duration: const Duration(seconds: 1),
@@ -245,9 +246,9 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
                 } else {
                   PostViewMedia postViewMedia = widget.postViews![(widget.communityId != null || widget.communityName != null) ? index - 1 : index];
                   return AnimatedSwitcher(
-                    switchInCurve: Curves.ease,
                     switchOutCurve: Curves.ease,
-                    duration: Duration(milliseconds: postViewMedia.postView.read ? 400 : 0),
+                    duration: const Duration( milliseconds: 0 ),
+                    reverseDuration: const Duration( milliseconds: 400 ),
                     transitionBuilder: (child, animation) {
                       return FadeTransition(
                         opacity: Tween<double>(begin: 0.0, end: 1.0).animate(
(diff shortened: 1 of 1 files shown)


CTalvio:
Before:

[Kazam_screencast_00001.webm](https://github.com/thunder-app/thunder/assets/4365015/7f8db95d-3078-4d10-987b-363948c629c7)

micahmo (approved):
Nice, I had noticed this too. LGTM

lib/community/widgets/post_card_list.dart
```diff
@@ -359,15 +360,25 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
 
   Future<void> dismissRead() async {
     if (widget.postViews != null) {
-      setState(() {
-        showRead = false;
-      });
-      await Future.delayed(const Duration(milliseconds: 400));
+      if (!(widget.postViews!.length > 15)) {
```
hjiangsu, line 363: Just a quick question - why are there two calls to `widget.onScrollEndReached();`? One here and one on line 380

lib/community/widgets/post_card_list.dart
```diff
@@ -359,15 +360,25 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
...
+      }
+      for (var post in widget.postViews!) {
+        if (post.postView.read) {
+          setState(() {
+            toRemoveSet.add(post.postView.post.id);
+          });
+          await Future.delayed(const Duration(milliseconds: 60));
+        }
+      }
+      await Future.delayed(const Duration(milliseconds: 800));
```
hjiangsu, line 374: What is this for again? Is it to wait for all the animations to finish before resetting the state?

hjiangsu (commented):
Just left a few comments!

lib/community/widgets/post_card_list.dart
CTalvio: Yes. For some reason it does not wait for the loop to finish. There's probably a better way to do this, but I didn't figure one out yet.

CTalvio: And the 60 ms wait is what causes the dismissals to be staggered.

lib/community/widgets/post_card_list.dart
hjiangsu: Ahh okay - I think we can make the second call a bit more robust in this case. I actually had to do this to resolve a issue where posts would not load on tall screens.

In my case, I had to do it when the state was first initiated, but for your case, we can make do without `WidgetsBinding.instance.addPostFrameCallback((_) {}`

```dart
// Check to see if the initial load did not load enough items to allow for scrolling to occur and fetches more items
WidgetsBinding.instance.addPostFrameCallback((_) {
  bool isScrollable = _scrollController.position.maxScrollExtent > _scrollController.position.viewportDimension;

  if (context.read<CommunityBloc>().state.hasReachedEnd == false && isScrollable == false) {
    widget.onScrollEndReached();
  }
});
```

lib/community/widgets/post_card_list.dart
hjiangsu: Ahh okay okay, sounds good. I do think theres a better way but we can leave it as it is for now!

lib/community/widgets/post_card_list.dart
CTalvio: Done

Issue #587, opened 2023-08-08
Instance taglines do not support markdown


micahmo:
**Description**  
Instance taglines do not support markdown.

**How to Reproduce**  
1. Log into an instance which has a tagline containing markdown (currently, [geddit.social](https://geddit.social) fits the bill).
2. Look at the tagline above the feed and notice that it shows raw markdown.

**Expected Behavior**  
The markdown is rendered correctly.

**Screenshots**  
Seen in the video here: https://github.com/thunder-app/thunder/issues/583#issuecomment-1669784998

**Device & App Version:**  
N/A

**Additional Context**  
N/A

PR #593, opened 2023-08-09
Improve tagline support

micahmo:
## Pull Request Description

This PR improves tagline support in a couple ways.

1. Adds support for markdown rendering.
2. Adds support for multiple taglines, showing them at random (which is what the web seems to do).

Note: Previously we were using `expandable_text` to do the "show more/less" thing. Now that we're using a markdown renderer, I had to make a custom expander. It's not quite as sophisticated, so the cutoff point is based on characters, not lines, meaning that it may potentially be in the middle of a line. Not the end of the world!

## Issue Being Fixed

Issue Number: #587

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/802ee00d-b434-40ca-b39a-c870a242c8fa

https://github.com/thunder-app/thunder/assets/7417301/d8fbafde-3e44-4a51-8778-fb6782961274

## Checklist

- [ ] Did you update CHANGELOG.md? -- Already covered


hjiangsu (approved):
LGTM! I like the approach with the cache for taglines :D

micahmo:
> I like the approach with the cache for taglines :D

Haha yeah, I guess it's kind of like a throttle for just that value/widget.

> Feel free to add in a changelog entry for this since taglines are released now

Done!

> What text scaling does the taglines follow at the moment? I think the best one to follow would be the metadata font scaling but let me know your thoughts!

It's using `CommonMarkdownBody`, which doesn't currently expose a way to set the text scaling. Internally it uses one of two scales based on whether `isComment` is set. Since I didn't set that, it would be the same size as other things like post body, sidebar, etc. Would you like me to expose the scaling directly, use the comment size, or is it good as is?

hjiangsu:
> It's using CommonMarkdownBody, which doesn't currently expose a way to set the text scaling

Hmm, I think we can hold on that for now then. Refactoring `CommonMarkdownBody` to take in textScalingFactor could just be for a separate PR, and we can make a note to adjust the scale factor here to use the metadata value

micahmo:
Sounds good! Fixed conflicts and made a note about `textScalingFactor`.

Issue #596, opened 2023-08-09
Use more distinctive colours for image/link icons


tom-james-watson:
I find it difficult to distinguish between the "image" and "link" icons on posts, i.e. the little icon in the bottom right of the thumbnails on posts.

For example, in the below screenshot, it's not obvious that post 2 and 3 are of a different type to post 1, 4 and 5, because they are both two shades of brown.

<img src="https://github.com/thunder-app/thunder/assets/3852719/be6e870d-ee1d-4144-8edc-0ecf1e83359c" width=300 />

This seems to happen whether I use material you or not.

It would be better if these used more distinctive colours, i.e. if images used a secondary colour from the theme instead of a slightly different shade of the primary colour.

tom-james-watson:
I think they should be derived from the theme, that way they always fit with the rest of the app theme.

I believe that at the moment they are both using derivatives of the primary colour of the theme, but IMO maybe one should use a secondary colour, or maybe one secondary and one tertiary.

tom-james-watson:
If I can get past https://github.com/thunder-app/thunder/issues/598 I'd be more than happy to try and implement this myself.

hjiangsu:
From the Matrix discussions, it seems like it'll be difficult to use `secondary` and `tertiary` colours since some Material You colour themes have very similar secondary/tertiary colours (which would then cause the same issue mentioned originally)

We might have to think of an alternative method which keeps all the colours somewhat distinct while matching the theme

tom-james-watson:
I've made a PR (https://github.com/thunder-app/thunder/pull/603) with what I think is a reasonable way of addressing this. Let me know your thoughts on the PR - thanks!

PR #603, opened 2023-08-09
improve contrast between link and image post icons

tom-james-watson:
## Pull Request Description

This changes the colours used for compact list view posts, that indicate whether it is an image or a link post, to be more easily distinguishable. It does this while sticking with plain derivatives of the primary colour.

## Issue Being Fixed

Can't easily distinguish link and image post icons. It's hard to know which thumbnails should be clicked on to view an image, for example.

Issue Number: https://github.com/thunder-app/thunder/issues/596

## Screenshots / Recordings

For before image, see https://github.com/thunder-app/thunder/issues/596.

image

image

image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - Improve behavior of deferred comment indicator - contribution from @micahmo
 - Text scaling now respects system's font scaling. Text scaling is based off of the system font
 - Improved contrast on user chips and badges - contribution from @CTalvio 
+- Make it easier to distinguish image and link posts in the Compact List View - contribution from @tom-james-watson
 
 ### Fixed
 - Fixed issue where the community post feed was missing the last post - contribution from @ajsosa
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@
 - Improve behavior of deferred comment indicator - contribution from @micahmo
 - Text scaling now respects system's font scaling. Text scaling is based off of the system font
 - Improved contrast on user chips and badges - contribution from @CTalvio 
-- Make it easier to distinguish image and link posts in the Compact List View - contribution from @tom-james-watson
+- Make it easier to distinguish different post types in the Compact List View - contribution from @tom-james-watson
 
 ### Fixed
 - Fixed issue where the community post feed was missing the last post - contribution from @ajsosa
--- a/lib/community/widgets/post_card_type_badge.dart
+++ b/lib/community/widgets/post_card_type_badge.dart
@@ -44,8 +44,8 @@ class TypeBadge extends StatelessWidget {
                     topLeft: Radius.circular(12),
                     topRight: Radius.circular(4),
                   ),
-                  color: theme.colorScheme.tertiaryContainer,
-                  child: const Icon(size: 17, Icons.wysiwyg_rounded),
+                  color: theme.colorScheme.inverseSurface,
+                  child: Icon(size: 17, Icons.wysiwyg_rounded, color: theme.colorScheme.inversePrimary),
                 )
               : postViewMedia.media.firstOrNull?.mediaType == MediaType.link
                   ? Material(
@@ -70,7 +70,7 @@ class TypeBadge extends StatelessWidget {
                         topRight: Radius.circular(4),
                       ),
                       color: theme.colorScheme.primaryContainer,
-                      child: Icon(size: 17, Icons.image_outlined, color: theme.colorScheme.primary),
+                      child: Icon(size: 17, Icons.image_outlined, color: theme.colorScheme.onPrimaryContainer),
                     ),
         ),
       ),
--- a/lib/community/widgets/post_card_type_badge.dart
+++ b/lib/community/widgets/post_card_type_badge.dart
@@ -15,6 +15,14 @@ class TypeBadge extends StatelessWidget {
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
 
+    Color getMaterialColor(Color blendColor) {
+      return Color.alphaBlend(theme.colorScheme.primaryContainer.withOpacity(0.6), blendColor);
+    }
+
+    Color getIconColor(Color blendColor) {
+      return Color.alphaBlend(theme.colorScheme.onPrimaryContainer.withOpacity(0.9), blendColor);
+    }
+
     return SizedBox(
       height: 28,
       width: 28,
@@ -44,8 +52,12 @@ class TypeBadge extends StatelessWidget {
                     topLeft: Radius.circular(12),
                     topRight: Radius.circular(4),
                   ),
-                  color: theme.colorScheme.inverseSurface,
-                  child: Icon(size: 17, Icons.wysiwyg_rounded, color: theme.colorScheme.inversePrimary),
+                  color: getMaterialColor(Colors.green),
+                  child: Icon(
+                    size: 17,
+                    Icons.wysiwyg_rounded,
+                    color: getIconColor(Colors.green),
+                  ),
                 )
               : postViewMedia.media.firstOrNull?.mediaType == MediaType.link
                   ? Material(
@@ -55,11 +67,11 @@ class TypeBadge extends StatelessWidget {
                         topLeft: Radius.circular(12),
                         topRight: Radius.circular(4),
                       ),
-                      color: theme.colorScheme.primary,
+                      color: getMaterialColor(Colors.blue),
                       child: Icon(
                         size: 19,
                         Icons.link_rounded,
-                        color: theme.colorScheme.onPrimary,
+                        color: getIconColor(Colors.blue),
                       ),
                     )
                   : Material(
@@ -69,8 +81,12 @@ class TypeBadge extends StatelessWidget {
                         topLeft: Radius.circular(12),
                         topRight: Radius.circular(4),
                       ),
-                      color: theme.colorScheme.primaryContainer,
-                      child: Icon(size: 17, Icons.image_outlined, color: theme.colorScheme.onPrimaryContainer),
+                      color: getMaterialColor(Colors.red),
+                      child: Icon(
+                        size: 17,
+                        Icons.image_outlined,
+                        color: getIconColor(Colors.red),
+                      ),
                     ),
         ),
       ),
--- a/lib/community/widgets/post_card_type_badge.dart
+++ b/lib/community/widgets/post_card_type_badge.dart
@@ -55,8 +55,12 @@ class TypeBadge extends StatelessWidget {
                         topLeft: Radius.circular(12),
                         topRight: Radius.circular(4),
                       ),
-                      color: theme.colorScheme.secondaryContainer,
-                      child: const Icon(size: 19, Icons.link_rounded),
+                      color: theme.colorScheme.primary,
+                      child: Icon(
+                        size: 19,
+                        Icons.link_rounded,
+                        color: theme.colorScheme.onPrimary,
+                      ),
                     )
                   : Material(
                       borderRadius: const BorderRadius.only(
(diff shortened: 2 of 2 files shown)


CTalvio:
I'm only seeing two colors here, whats the text post color?

tom-james-watson:
I hadn't seen the setting for text post indicator. Will take a look at what I can do.

micahmo:
This looks awesome!

I wonder if it would work with user chips too. ☺

CTalvio:
image

This does not work reliably. Here is a low-contrast example.

tom-james-watson:
My last commit should address that - I updated to use `onPrimaryContainer` for the icon that shows on `primaryContainer`. Will update the screenshots.

CTalvio:
Please note:

The indicators currently uses the primary, secondary and tertiary accent colors. They use the container colors to ensure contrast on both light and dark mode, they previously used the normal accent colours, this did not provide adequate contrast with all themes, do not use these colors.

If you want additional separation of the three colors, and have them be based on the theme colors, you will likely need to write your own color derivator.

tom-james-watson:
I've updated the screenshots with the latest changes, including one using (I think) the example you gave.

tom-james-watson:
> I wonder if it would work with user chips too. ☺

What do you mean by this by the way? I assume it would be a separate issue / PR but maybe I can take a look

CTalvio:
> > I wonder if it would work with user chips too. relaxed
> 
> What do you mean by this by the way? I assume it would be a separate issue / PR but maybe I can take a look

He's referring to the username highlight of OP, self, mod, and admin.

CTalvio:
I like the approach of flipping the colors of some icons to create distinction, however.

How about this:

Media (image, and in the future, video posts): containerColor for background, onContainerColor for icon.
Written media (links and text post): onContainerColor for background, containerColor for background

image

tom-james-watson:
> I'm not sure I like every icon being a combination of unrelated theme presets.

Whether people like it or not I will leave it up to you maintainers, of course. Though having the different types of post be more distinguishable was obviously the intention.

Honestly if I were the one making the decisions, I would just drop the "text post indicator" option completely and then also drop the icon from link posts. That way you're left with only icons on image posts (and in the future video posts), which to me makes sense because these are the only ones where the thumbnail is actually "interactive". It would be similar to how Relay for Reddit works.

CTalvio:
You can disable text post thumbnails in settings.

tom-james-watson:
> How about this:

> Media (image, and in the future, video posts): containerColor for background, onContainerColor for icon.
Written media (links and text post): onContainerColor for background, containerColor for background

Do you mean `primaryContainer` and `onPrimaryContainer`? Those properties you refer to don't exist? Looks like you have a different colour completely vs the link icon though.

CTalvio:
image

```
Padding(
          padding: const EdgeInsets.only(
            left: 2.5,
            top: 2.5,
          ),
          child: postViewMedia == null || postViewMedia.media.isEmpty
              ? Material(
                  borderRadius: const BorderRadius.only(
                    bottomLeft: Radius.circular(4),
                    bottomRight: Radius.circular(12),
                    topLeft: Radius.circular(12),
                    topRight: Radius.circular(4),
                  ),
                  color: Color.alphaBlend(theme.colorScheme.primaryContainer.withOpacity(0.15), theme.colorScheme.onPrimaryContainer),
                  child: Icon(size: 17, Icons.wysiwyg_rounded, color: theme.colorScheme.primaryContainer),
                )
              : postViewMedia.media.firstOrNull?.mediaType == MediaType.link
                  ? Material(
                      borderRadius: const BorderRadius.only(
                        bottomLeft: Radius.circular(4),
                        bottomRight: Radius.circular(12),
                        topLeft: Radius.circular(12),
                        topRight: Radius.circular(4),
                      ),
                      color: Color.alphaBlend(theme.colorScheme.secondaryContainer.withOpacity(0.15), theme.colorScheme.onSecondaryContainer),
                      child: Icon(
                        size: 19,
                        Icons.link_rounded,
                        color: theme.colorScheme.secondaryContainer,
                      ),
                    )
                  : Material(
                      borderRadius: const BorderRadius.only(
                        bottomLeft: Radius.circular(4),
                        bottomRight: Radius.circular(12),
                        topLeft: Radius.circular(12),
                        topRight: Radius.circular(4),
                      ),
                      color: theme.colorScheme.primaryContainer,
                      child: Icon(size: 17, Icons.image_outlined, color: Color.alphaBlend(theme.colorScheme.primaryContainer.withOpacity(0.15), theme.colorScheme.onPrimaryContainer)),
                    ),
        ),
```

CTalvio:
> > How about this:
> 
> > Media (image, and in the future, video posts): containerColor for background, onContainerColor for icon.
> > Written media (links and text post): onContainerColor for background, containerColor for background
> 
> Do you mean `primaryContainer` and `onPrimaryContainer`? Those properties you refer to don't exist? Looks like you have a different colour completely vs the link icon though.

Yeah I wasn't using the exact names, just illustrating my idea, anyway, feel free to use that as a starting point.

tom-james-watson:
Yeah I will play around more because you still get very similar colours sometimes with your code:

image

CTalvio:
Yup, some themes just straight up have the same color for the accents. Usually its the secondary and tertiary that are too similar, but sometimes it's the primary and secondary.

CTalvio:
There's always the option of editing the actual themes. as well.

tom-james-watson:
Been playing around and have two ideas.

One is to just do alpha blending like you suggest and have the three options be blends of primary with red, blue and green. This means you get colours that feel like that they fit in the app and are obviously different, without being too saturated or light or dark:

image

image

Another option that I also think might make more sense is to stop trying to find X different colour combinations for all the post types. What we're really trying to solve here is being able to more easily see which posts allow me to interact with the thumbnail. Therefore, it could just be a case of having one colour for interactive (images and later videos) posts and one colour for non-interactive (links and text):

image

image

Let me know your thoughts.

hjiangsu:
Thanks for making this PR! These are my own thoughts on this:
- I feel like using distinctive colours here seem to be the way to go. I don't think we can cover all of the Material You theming and be able to accommodate for all those use-cases. It'll be difficult to find a combination of primary/secondary/tertiary colours which don't look similar
- Although using static colours mean that it no longer matches the Material You theme, this change would be a deliberate choice. Once users get used to the colours, it would help keep their expectations consistent no matter the colour theme (which is my personal preference for things like this)

If anything, we can potentially revisit this in the future (or you can feel free to try this out) to make some colour derivator which takes in a Material theme and outputs distinctive yet complementary colours. Some potential resources for this:
- https://m3.material.io/theme-builder#/custom (see Extended Colours)
- https://api.flutter.dev/flutter/package-material_color_utilities_blend_blend/Blend/harmonize.html (`harmonize` method which might be useful to generate a colour that matches another given colour)

tom-james-watson:
Thanks for the input! Happy for me to update this PR with some static colours? Any strong opinions on what they are? If not, I will just play around with some sensible combinations (tomorrow).

CTalvio:
FYI, making these colors static is something I will fork the project over, I've been wanting to make the static comment depth colors fit the theme, too.

By all means, make static UI colors an option, but this is a case where I will take aesthetics over functionality.

> Another option that I also think might make more sense is to stop trying to find X different colour combinations for all the post types. What we're really trying to solve here is being able to more easily see which posts allow me to interact with the thumbnail. Therefore, it could just be a case of having one colour for interactive (images and later videos) posts and one colour for non-interactive (links and text):

This is why I suggested making one category dark on bright, and the other bright on dark.

I like the blended colors, primary colors blended towards the current theme seem to work quite nicely, different shades of the same color remain recognizable, but also congruent with the app.

machinaeZER0:
Really liking the more differentiated colors overall! Definitely will help keep the different post types quickly identifiable at a glance.

This would be outside the scope of this PR, but I wonder if we would consider making the badges/icons sliiiightly smaller, if we end up making this change? Since the thumbnails in compact mode are themselves quite small, it would be great to see as much of the image as possible, and if we increase the diversity of the badge colors then users should quickly acclimate to which color means what (in addition to the icons that would still be present). Thoughts on that?

machinaeZER0:
> FYI, making these colors static is something I'd fork the project over, I've been wanting to make the static comment depth colors fit the theme, too.

What will you be calling your fork?

CTalvio:
> > FYI, making these colors static is something I'd fork the project over, I've been wanting to make the static comment depth colors fit the theme, too.
> 
> What will you be calling your fork?

Nothing. I'd just start maintaning my own version and build my own apks. Its not like I want to compete, just meet my own desires first and foremost.

micahmo:
> > I wonder if it would work with user chips too. ☺
> 
> What do you mean by this by the way? I assume it would be a separate issue / PR but maybe I can take a look

@tom-james-watson Special users (OP, admin, mod) are identified by placing their username in a colored container (very similar to the post type indicators). I believe they suffer from the same problem where they're no longer very distinguishable. If you come up with a good mechanism here for the post indicators, I'm hoping you can apply the same change there.

https://github.com/thunder-app/thunder/blob/3a50c73f2eae98cb783d260c8ab3d99d48695c26/lib/post/widgets/comment_header.dart#L282-L285

Example:

| image |
| - |

> Been playing around and have two ideas.
> 
> One is to just do alpha blending like you suggest and have the three options be blends of primary with red, blue and green. This means you get colours that feel like that they fit in the app and are obviously different, without being too saturated or light or dark:

My two cents, this is the best thing I've seen in the thread. The colors are completely distinct while following the theme. Obviously you just posted two examples, but if the same holds true for other theme, I'd say go for it!

> I wonder if we would consider making the badges/icons sliiiightly smaller

+1 to this as well! Obviously we're not asking @tom-james-watson to do extra work when he has already gracious dived into improving the colors. But yes, this is something I'd like to see too.

machinaeZER0:
> > > FYI, making these colors static is something I'd fork the project over, I've been wanting to make the static comment depth colors fit the theme, too.
> > 
> > 
> > What will you be calling your fork?
> 
> Nothing. I'd just start maintaning my own version and build my own apks. Its not like I want to compete, just meet my own desires first and foremost.

Kinda sounds like you'd just like a toggle (similar to the Material You toggle) that themes everything more broadly than Accent Color currently does? Feel like that would satisfy all parties, if implemented thoughtfully :)

machinaeZER0:
> Obviously we're not asking @tom-james-watson to do extra work when he has already gracious dived into improving the colors.

100%! Thank you for taking interest in Thunder, Tom :D

tom-james-watson:
I've given this some more thought and I agree with what looks like the general consensus - that the version with the alpha blends of red/blue/green is the nicest solution to this for now. It means that you both have consistent colouring of post types (if not 100% static), but you still keep the colours harmonious with the rest of the theme. I also switched which colours are used where to make links and images have the most contrast (red vs blue), because most of the time "Show Text Post Indicator" will be off.

I've pushed a commit that introduces this change. I'm happy with how it looks when switching between dark/light and all of the different themes. There are a few screenshots below for reference.

Let me know if you're happy with this!

> 100%! Thank you for taking interest in Thunder, Tom :D

Thank you to you guys for making it! It's been a great experience so far ❤️ 

> +1 to this as well! Obviously we're not asking @tom-james-watson to do extra work when he has already gracious dived into improving the colors. But yes, this is something I'd like to see too.

I'm happy to keep working on a few things like this. I can try and take a look as a follow-up.

image

image

image

image

CTalvio:
The color blend looks so good.

tom-james-watson:
I can maybe look at that after this PR. We could split out these new functions into something more generic.

micahmo (approved):
@hjiangsu makes the final call, but this looks amazing to me!

machinaeZER0:
Beautiful!

machinaeZER0:
One thought - I don't know how others feel, but I think unless it somehow affects readability I'd lean towards the icons themselves always being white, unless it feels correct to have some sort of edge case for particularly bright colored badges? I don't feel super strongly about it, but white icons feel a bit "cleaner" to me I guess.

CTalvio:
I think we should merge this now, then, so users can get used to the red green blue and not have it change on them in the release after this one.

@hjiangsu?

Also yes, I'd been thinking of the color blend option for a while, but had not explored it yet. I only mentioned it in matrix.

Clearly it works great!

hjiangsu:
Already ahead of you! :D

PR #609, opened 2023-08-10
Improve and refactor read post colors

micahmo:
## Pull Request Description

This PR attempts to improve (both in code and in the UI) the color and contrast of read posts in the feed.

~1. I still felt like the contrast of the read indicator was insufficient after the most recent changes, at least in light mode. I've been daily driving `develop` and I'm constantly second-guessing whether something is read or not.~
~2. We have bits of all over the place which set various colors and styles based on the read status. This is hard to maintain, and there are already some things that don't necessarily follow this pattern.~

~I thought it would be much easier to simple tweak the opacity of the whole card. This would ensure...~
~1. The code is all in one place and every widget is automatically affected (including special post indicators, the thumbnail, etc).~
~2. It's easy to tweak things so we can adjust as needed, and we won't have to remember to apply read styling to new widgets.~

The above approach has been abandoned because `Opacity` is bad. See later comments.

> Review with whitespace off.

## Issue Being Fixed

Issue Number: #617

### These are outdated!

| Before | After |
|--------|--------|
| image | image |
| image | image | 

## Checklist

- [ ] Did you update CHANGELOG.md? -- N/A -- we already mention color improvements

### DIFF
--- a/lib/shared/link_preview_card.dart
+++ b/lib/shared/link_preview_card.dart
@@ -35,6 +35,7 @@ class LinkPreviewCard extends StatelessWidget {
     required this.hideNsfw,
     required this.isUserLoggedIn,
     required this.markPostReadOnMediaView,
+    this.read,
   });
 
   final int? postId;
@@ -57,6 +58,8 @@ class LinkPreviewCard extends StatelessWidget {
 
   final ViewMode viewMode;
 
+  final bool? read;
+
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
@@ -73,6 +76,7 @@ class LinkPreviewCard extends StatelessWidget {
           children: [
             if (mediaURL != null) ...[
               ImagePreview(
+                read: read,
                 url: mediaURL ?? originURL!,
                 height: showFullHeightImages ? mediaHeight : 150,
                 width: mediaWidth ?? MediaQuery.of(context).size.width - (edgeToEdgeImages ? 0 : 24),
@@ -81,28 +85,34 @@ class LinkPreviewCard extends StatelessWidget {
             ] else if (scrapeMissingPreviews)
               SizedBox(
                 height: 150,
-                child: hideNsfw
-                    ? ImageFiltered(
-                        imageFilter: ImageFilter.blur(sigmaX: 30, sigmaY: 30),
-                        child: LinkPreviewGenerator(
+                // This is used for external links when Lemmy does not provide a preview thumbnail
+                // and when the user has enabled external scraping.
+                // This is only used in comfortable mode.
+                child: Opacity(
+                  opacity: read == true ? 0.55 : 1,
+                  child: hideNsfw
+                      ? ImageFiltered(
+                          imageFilter: ImageFilter.blur(sigmaX: 30, sigmaY: 30),
+                          child: LinkPreviewGenerator(
+                            link: originURL!,
+                            showBody: false,
+                            showTitle: false,
+                            placeholderWidget: Container(
+                              margin: const EdgeInsets.all(15),
+                              child: const CircularProgressIndicator(),
+                            ),
+                            cacheDuration: Duration.zero,
+                          ))
+                      : LinkPreviewGenerator(
                           link: originURL!,
                           showBody: false,
                           showTitle: false,
-                          placeholderWidget: Container(
-                            margin: const EdgeInsets.all(15),
-                            child: const CircularProgressIndicator(),
+                          placeholderWidget: const Center(
+                            child: CircularProgressIndicator(),
                           ),
                           cacheDuration: Duration.zero,
-                        ))
-                    : LinkPreviewGenerator(
-                        link: originURL!,
-                        showBody: false,
-                        showTitle: false,
-                        placeholderWidget: const Center(
-                          child: CircularProgressIndicator(),
                         ),
-                        cacheDuration: Duration.zero,
-                      ),
+                ),
               ),
             if (hideNsfw)
               Container(
@@ -140,6 +150,7 @@ class LinkPreviewCard extends StatelessWidget {
           children: [
             mediaURL != null
                 ? ImagePreview(
+                    read: read,
                     url: mediaURL!,
                     height: 75,
                     width: 75,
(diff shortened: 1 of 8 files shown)


tom-james-watson:
Nice! For me the text on the light mode in after has too low contrast. 

I think the reason it was easier to see the difference in dark mode before was because there's more difference between the read/unread backgrounds. I think instead of lowering the contrast on the text in light mode, instead try making the read background a bit darker. Worth a try maybe.

micahmo:
@tom-james-watson Thanks for the feedback! I totally see where you're coming from. I guess I was using "contrast" in the wrong way (contrasting the read text vs unread as opposed to read vs background). Relay takes a similar approach to what you're saying by making the background really dark and not touching the foreground.

Personally, I would still like to dim the foreground a bit (I especially like dimming the media previews as I like to scan down the right side and that's a nice quick indicator).

The nice thing about these changes is that it's really easy to tweak things now!

I will play around with this some more when I'm at my computer, and we'll see if anyone else has feedback.

hjiangsu:
As mentioned on Matrix, we should try to find an alternative to using `Opacity`!  Opacity is pretty intensive and could potentially lead to performance regressions. See here: https://docs.flutter.dev/perf/best-practices#minimize-use-of-opacity-and-clipping

micahmo:
Alright, I'm going to revert most of this back to the way it was. As much as it's hard to maintain, I understand it's better for performance to set colors directly on widgets rather than overlaying with an `Opacity`.

One easy win is to ensure that post identifiers (pinned/saved) are also dimmed. I've committed that. I can also continue to play around with contrast.

One of the things I liked best about the `Opacity` overlay was that it affected all text/image/link previews and badges. This will be really hard to go through and do manually. At least for images, [setting transparency is apparently better than `Opacity`](https://api.flutter.dev/flutter/widgets/Opacity-class.html#transparent-image).

micahmo:
Alrighty, I think I'm all set to have this reviewed again. Here is a summary of the changes.

**Changes**
 - Darken/lighten background in light/dark mode (respectively) for more contrast

**Additions**
 - Dim special post indicators (saved/pinned)
 - Dim media preview for
   - Links previews where the thumbnail comes from Lemmy
   - Image previews for image posts
   - Link previews when Lemmy doesn't provide a thumbnail and scraping is off
   - Link previews when Lemmy doesn't provide a thumbnail and scraping is on
     - This is the only place I used `Opacity`, because I can't style the images within `LinkPreviewGenerator`. We already have this off by default with a warning that it hurts performance, so I hope this is ok. If not, I'll take it out!
   - (I also tried to identify each of the above cases with comments)
 - Dim post type badges (foreground and backgroud)

There may be some cases I missed which we can cover when we find them (as this is just cosmetic, it's not the end of the world).

Going forward we just have to remember to do this same thing to any new elements we put on the post cards.

## Demos

| Before | After |
|--------|--------|
| image | image |
| image | image | 

| Before | After |
|--------|--------|
| image | image |
| image | image |

tom-james-watson:
The new background colour looks great!

Personally I think the faded image is a bad idea - I still interact with read posts and I think this makes that a bit more annoying. I don't feel strongly about it though!

Overall looks much better 👍

micahmo:
> The new background colour looks great!

Thanks!

> Personally I think the faded image is a bad idea - I still interact with read posts and I think this makes that a bit more annoying.

Ahh that's the one thing I wanted the most! 😆 I scan down the right-hand side, and having a dimmed preview is the quickest indicator to me that the post is read.

Is this something worth putting in an option?

lib/shared/link_preview_card.dart
```diff
@@ -83,28 +87,34 @@ class LinkPreviewCard extends StatelessWidget {
...
                 SizedBox(
                   height: 150,
-                  child: hideNsfw
-                      ? ImageFiltered(
-                          imageFilter: ImageFilter.blur(sigmaX: 30, sigmaY: 30),
-                          child: LinkPreviewGenerator(
+                  // This is used for external links when Lemmy does not provide a preview thumbnail
+                  // and when the user has enabled external scraping.
+                  // This is only used in comfortable mode.
+                  child: Opacity(
```
hjiangsu, line 93: Is there any way to remove this usage of `Opacity`?

lib/shared/link_preview_card.dart
```diff
@@ -152,10 +163,25 @@ class LinkPreviewCard extends StatelessWidget {
...
                         height: 75,
                         width: 75,
-                        child: hideNsfw
-                            ? ImageFiltered(
-                                imageFilter: ImageFilter.blur(sigmaX: 30, sigmaY: 30),
-                                child: LinkPreviewGenerator(
+                        // This is used for external links when Lemmy does not provide a preview thumbnail
+                        // and when the user has enabled external scraping.
+                        // This is only used in compact mode.
+                        child: Opacity(
```
hjiangsu, line 169: Same comment here as above

hjiangsu (commented):
Just two more comments!

lib/shared/link_preview_card.dart
micahmo: I addressed this in my comment.

> - Link previews when Lemmy doesn't provide a thumbnail and scraping is on
>   - This is the only place I used `Opacity`, because I can't style the images within `LinkPreviewGenerator`. We already have this off by default with a warning that it hurts performance, so I hope this is ok. If not, I'll take it out!

So I think it's up to you whether this scenario (off by default + performance warning) is ok for `Opacity`, or if I should take it out, or if you have any other ideas to style `LinkPreviewGenerator` without forking. 😊

hjiangsu: Ahh I see, I missed that part! Funny that you mention forking `LinkPreviewGenerator`, because we are using a fork of it right now 😆

https://github.com/thunder-app/link_preview_generator

machinaeZER0:
I love this idea! One question - how will this work when using the true black/AMOLED setting?

machinaeZER0:
I love this idea! One question - how will this work when using the true black/AMOLED setting?

EDIT: Also I like the dimmed thumbnails here, personally :)

lib/shared/link_preview_card.dart
micahmo: > Funny that you mention forking `LinkPreviewGenerator`, because we are using a fork of it right now 😆

Aha! I remember @ajsosa was planning on doing that, but I missed that we already did it for one of @coslu's changes.

Alright, I have a PR open there: thunder-app/link_preview_generator#2. If/when that's approved and merged, I can update the package reference here, rip out `Opacity`, and we should be good to go!

micahmo:
> how will this work when using the true black/AMOLED setting?

Here's a before & after using pure black. The difference is more subtle, but I still think it's an improvement (and I think the dimmed previews do even more work here). What do you think?

| Before | After |
|--------|--------|
| image | image |

machinaeZER0:
Looks great to me! :)

tom-james-watson:
Looks awesome!

micahmo:
Should be all set now!

hjiangsu (approved):
LGTM!

PR #614, opened 2023-08-10
Show External Link preview Thumbnails & option change

ajsosa:
Fixed an issue where we weren't displaying external link preview thumbnails when we had them if the show external link previews option was off. The original intent of that option was to limit html scraping. So instead, the option has been completely changed. It is now called scrape missing external link previews. It's purpose is only to scrape the things we are missing and if it's off, we will still show the thumbnails we do get from Lemmy. This option is off by default for performance reasons.

- [✔ ] Did you update CHANGELOG.md?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Added domain for posts linking to external websites - contribution from @CTalvio
 - Added comment navigation buttons - contribution from @micahmo
 - Added full screen swipe to go back on main pages
+- Added new option scrape missing external link previews which is off by default. It's purpose is to attempt to find an image when an external link thumbnail is not available - contribution @ajsosa
 
 ### Changed
 - Removed tap zones for author/community on compact post cards - contribution from @CTalvio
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Added domain for posts linking to external websites - contribution from @CTalvio
 - Added comment navigation buttons - contribution from @micahmo
 - Added full screen swipe to go back on main pages
+- Added new option scrape missing external link previews which is off by default. It's purpose is to attempt to find an image when an external link thumbnail is not available - contribution @ajsosa
```
micahmo, line 29: Should be "its" without apostrophe. 😉

micahmo (approved):
LGTM

BTW @ajsosa to make the checkmark, put an x between the brackets, like this:

```
- [x] completed item
```

- [x] completed item

CHANGELOG.md
hjiangsu: Fixed it

hjiangsu:
I'll go ahead and merge this in, to be released as part of the next nightly to get some immediate feedback on how it feels!

PR #616, opened 2023-08-10
use text for top sort icons

tom-james-watson:
## Pull Request Description

This is a work-in-progress change to switch the icons used for the different kinds of "Top" sorting to just use short bits of text instead.

This is not ready at all to be merged, I just wanted to solicit some feedback before working on this further. If people like the idea then I will work out the nicest way to implement this properly.

What do you think? Do you agree with this change?

**EDIT - the proposal has now changed - please read comments**

## Issue Being Fixed

The icons used for the different kinds of "Top" sorts are not intuitive and it's not clear what they mean.

Issue Number: N/A

## Screenshots / Recordings

image

image

## Checklist

### DIFF
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -224,10 +224,7 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
                                       ));
                                 }),
                             IconButton(
-                                icon: Text(
-                                  "6h",
-                                  style: theme.textTheme.bodyLarge?.copyWith(fontWeight: FontWeight.bold),
-                                ),
+                                icon: Icon(sortTypeIcon, semanticLabel: AppLocalizations.of(context)!.sortBy),
                                 tooltip: sortTypeLabel,
                                 onPressed: () {
                                   HapticFeedback.mediumImpact();
--- a/lib/community/pages/community_page.dart
+++ b/lib/community/pages/community_page.dart
@@ -224,7 +224,9 @@ class _CommunityPageState extends State<CommunityPage> with AutomaticKeepAliveCl
                                       ));
                                 }),
                             IconButton(
-                                icon: Icon(sortTypeIcon, semanticLabel: AppLocalizations.of(context)!.sortBy),
+                                icon: sortTypeIcon != null
+                                    ? Icon(sortTypeIcon, semanticLabel: AppLocalizations.of(context)!.sortBy)
+                                    : Text(sortTypeLabel ?? AppLocalizations.of(context)!.sortBy, style: const TextStyle(fontWeight: FontWeight.bold)),
                                 tooltip: sortTypeLabel,
                                 onPressed: () {
                                   HapticFeedback.mediumImpact();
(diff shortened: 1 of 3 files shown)


micahmo:
I overall like this change! One consideration that comes to mind is how this would work when the app is localized. I'm guessing there are at least some cultures where `[n]h`, `[n]d`, `[n]y` wouldn't make sense. Another thought is whether this could be overlayed onto a calendar icon using a `Stack`, but then things might be too cramped.

While I'm thinking out loud, I've noticed some apps put the full sort type name directly under the main heading. But even if we went in that direction, your icons would be useful in the bottom sheet.

Anyway, sorry for rambling. I like this!

tom-james-watson:
I've not worked with localization in the app yet, but I imagine it should be doable as it's just plain `Text` widgets.

tom-james-watson:
Another option is just drop the icons for these options and do like below. Honestly I prefer that option, actually.

image

tom-james-watson:
> I am generally a proponent of having icons in menus. If nothing else, it gives the user something to latch into subconsciously for quickly picking the desired option, rather than having to read all the options.

Generally I agree, but only if the icon is obvious as to what it represents. My latest push updates the PR to the icon-less variant for "Top" options. The other sort options stay the same. Some more screenshots:

image

image

image

image

Would be interested to hear some more opinions - @machinaeZER0, @CTalvio, @hjiangsu ?

micahmo:
Agreed, I will defer to the team on this. Thanks for your help!

machinaeZER0:
I personally like keeping icons, to @micahmo's point, though I think we could definitely workshop some of them to be more clear (feel like the one/six/twelve hour ones could maybe be changed to some sort of clock icon in the future, with the amount of time highlighted?)

Either way I'd definitely keep the "top" part written out in each option, since we're not hurting for space here :) the icons feel especially relevant since they do get displayed at the top of the screen, as opposed to just a generic sorting icon, otherwise I might feel differently!

Defer to the rest of the group on a final call, just my two cents!

micahmo:
Hey I was thinking about this PR some more. I think we should split the talking points into two distinct areas.

1. What is displayed in the AppBar
2. What is displayed in the modal picker

For what's displayed in the AppBar, I totally see the argument for using the text description of the current sort mode. In fact, as I poke around other apps, a common paradigm I see is to list the sort name on the left, and have a generic sort/filter icon on the right. 

Here's Sync for reference, but almost all of the apps are similar to this. I'm not saying we have to do this, but it's definitely a pattern.

image

For number 2, I'm having a hard time justifying the removal of the icons. Whether we keep the original icons or go with your number/letter idea (which I think was more about satisfying number 1), I don't see how removing them helps things. Aside from making the list look pretty, there's also the subconscious association we talked about where icons can help users quickly latch onto a selection without reading all the text.

Anyway, this just popped back into my head recently so I wanted to write it down. 😊

tom-james-watson:
Yeah I think I actually prefer something like what Sync does. Relay does something similar, too. Maybe I can try mocking that up - if it worked like that then there would be no need for this change.

machinaeZER0:
I kinda like the icon showing the sort option at a glance, but depending on the icon it's definitely not super user-friendly at a glance. I think standardizing it to a generic Sort icon makes perfect sense!

tom-james-watson:
I've implemneted the discussed suggestion in https://github.com/thunder-app/thunder/pull/616, so closing this in favour of that.

PR #624, opened 2023-08-11
Handle deferred comment loading issue

micahmo:
## Pull Request Description

I've been having issues with deferred comments not loading. I found the same issue with the same comments on the web, so it's not strictly a Thunder problem, but I wanted to dig in anyway. It turns out that we occasionally get comments where the none of the children are direct descendants of the parent. For example, the parent path may be `0.10` and the child with ID `30` may be `0.10.20.30`. I'm not sure why it happens (maybe comment `20` got deleted?), but I thought it might be nice to handle this case.

Note that since we do actually get the child comments in the response, we could go a step further and insert placeholders for the missing comments so the results are still displayed. But since this seems like a backend issue (which the web client also fails to handle), I think at least showing a nice error message is an improvement.

## Issue Being Fixed

Issue Number: N/A

### Before (web and app)

https://github.com/thunder-app/thunder/assets/7417301/99746564-57a7-4fe3-a546-9c17927a7b1d

https://github.com/thunder-app/thunder/assets/7417301/115451c8-c913-4fb6-b9b4-35c3af8f951f

### After

https://github.com/thunder-app/thunder/assets/7417301/c4030b52-9be7-4ccb-b726-567d3895d514

## Checklist

- [ ] Did you use localized strings where applicable? -- No, none of these error messages are localized yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@
 - Fix improper back button handling - contribution from @micahmo 
 - Fixed feed page reaching the end in some cases where NSFW content is turned on
 - Fixed issue where external link thumbnails weren't being displayed due to show external link previews option being off which was only intended to prevent html scraping - contribution from @ajsosa
+- Handle issue where some deferred comments won't load - contribution from @micahmo
 
 ## 0.2.1+13 - 2023-07-25
 ### Added
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -285,6 +285,15 @@ class PostBloc extends Bloc<PostEvent, PostState> {
             throw Exception('Error: Timeout when attempting to fetch more comments');
           });
 
+          // Determine if any one of the results is direct descent of the parent. If not, the UI won't show it,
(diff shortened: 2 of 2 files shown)


machinaeZER0:
Is "unable to load direct children" our language, or Lemmy's? If possible I might change it to something simpler, like "unable to load comments". "Direct children" feels odd to me I think, even if that is the technical verbiage

micahmo:
It's our language (you can see the code here: https://github.com/thunder-app/thunder/pull/624/files#diff-d652fa040b3a56cfa0e6ef2042cc3a8fdd54216b6251776edbac9bad9fc52559R293). The problem here is moreso that we're directly showing the error (`Exception`) to the user. I think eventually we'll match all errors (from Lemmy or us) to a more human-friendly description.

For now, I was hoping the raw error would reference "direct child" because that's the very specific error that we're handling here, and if people report it, we'll know exactly what going on.

machinaeZER0:
That's fair, helping keep error reports clear/specific is a good goal!

hjiangsu (approved):
LGTM!

PR #374, opened 2023-07-15
Fix double back

micahmo:
This is an attempt to fix an issue discovered by @machinaeZER0 and demonstrated [here](https://github.com/thunder-app/thunder/pull/338#issuecomment-1636641301). (There are also some formatting changes introduced by workspace settings.)

Please [review without whitespace](https://github.com/thunder-app/thunder/pull/374/files?diff=split&w=1).

Issue #626, opened 2023-08-12
Possible regression in Back action behavior/animations?


machinaeZER0:
I'm not 100% sure if this is just me or not, but it seems like some behavior regressions have occured in the most recent releases regarding the Back action:

- swiping back from post view seems to be triggering the comment swipe animation more often (which is weird, I feel like it was addressed recently and it felt like I could not longer reproduce)
- swiping back has started sometimes triggering twice again - this seems most likely to trigger when swiping back on post content, which brings me back to my feed, but also gives me the "swipe back again to exit" popup

Is this happening for anyone else? Any idea what might have changed to cause this? Let me know if I can help troubleshoot! Using the latest Android alpha now :)

micahmo:
There were a bunch of changes to the back behavior, so this isn't too surprising. I can't repro triggering the gestures (maybe because I'm pretty consistently swiping from the very edge of the display), but I can repro the double back. I'll see if I can fix that (again 😆).

machinaeZER0:
I swipe with gusto, hahah. Please let me know if I can be of assistance!

micahmo:
@machinaeZER0 Let me know how things look in the next build!

machinaeZER0:
Thank you! I'll keep an eye out!

hjiangsu:
Just curious, but this this still happening? I think the fix was added back in the general release (0.2.3+16)

machinaeZER0:
On a quick test with the last nightly build I'm currently not able to get the "press back again to exit" popup to accidentally trigger, but I am able to trigger the start of a comment swipe fairly consistently. It still seems to only happen if I swipe from the left side, and I'm swiping with some amount of gusto. I think(?) it's happening primarily on parent comments?

hjiangsu:
Hmm, just curious, but is your swipe to go back gesture happening from the left or the right side of the screen? (for example, do you swipe from left-to-right or from right-to-left when swiping back)

machinaeZER0:
Left to right!

Issue #312, opened 2023-07-10
Add option for full screen swipe-to-go-back


sageDieu:
**Is your feature request related to a problem? Please describe.**
I tend to prefer swipe actions for navigation, and in Apollo and other similar apps I'm usually able to turn on an option to swipe anywhere on the page to go back to the previous. 

**Describe the solution you'd like**
For a quick example check out the HackerNews app HACK: https://apps.apple.com/us/app/hack-for-hacker-news-reader/id1464477788

Enable the setting for "full screen swipe to go back" and then go load a post and swipe anywhere on the screen to go back to the list of posts.

micahmo:
Related to #198.

jordan1776:
Personally I would rather have partial screen swipe to go back rather than full screen so you could both have comment gestures and swipe back at the same time. This  type of action is already used for the side menu on the feed view.

micahmo:
> you could both have comment gestures and swipe back at the same time

One idea: Relay uses right-to-left for back and left-to-right for post/comment gestures.

thepaperpilot:
As another relay user, I'd also like to be able to swipe right to go back. Now that the swipe gestures are customizable, I'd love to have something that just entirely disabled swiping right on comments to allow going back instead

jordan1776:
>, I'd love to have something that just entirely disabled swiping right on comments to allow going back instead

Maybe the swipe settings should have up to 4 on the right side for emulating Relay and 2 on the left for Apollo, with the option to set the 2nd, 3rd, and 4th actions to none. And if both left side actions are set to none, use the full screen swipe.

Fmstrat:
This needs to be considered for iOS vs Android. Newer versions of Android have a default swipe back action. iOS will probably steal this in the future, but for now, if this is implemented it may need to be platform specific.

hjiangsu:
For swiping right to go back, iOS already has that natively (I can swipe right from any screen to go to the previous screen) I didn't fully realize that this was not the case for Android (I'm guessing it depends on the OS version or variant)

The one thing I'm wondering about with the full screen swipe is how it interferes with swipe gestures. Would swipe gestures on comments/posts need to be turned off for this to work or is there a way that both of them can work together?

micahmo:
> is there a way that both of them can work together?

Not saying this is the best answer, but one thing we discussed in this thread is having left-to-right go back and right-to-left for gestures. We might want up to four gestures per side in that case though.

hjiangsu:
[...]

PR #627, opened 2023-08-12
Add SwipeablePageRoute dead zone to not interfere with Android back gesture

micahmo:
## Pull Request Description

This PR attempts to fix part of #626 by introducing a swipe dead zone. This should help to avoid cases where a swipe could both activate the `SwipeablePageRoute` as well as the Android back gesture. Note that the value of `45` is the one one that previously fixed this issue.

Previous fix: https://github.com/thunder-app/thunder/pull/374/files#diff-0bcedfb901776bf2647be3a194eb8e5e9d3cce4282c31e36699533493ade671bR159

Note there are still general problems with the `SwipeablePageRoute` on Android (see #312), but this shouldn't make that better or worse.

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] Did you update CHANGELOG.md? -- N/A -- fixing regression

### DIFF
--- a/lib/community/widgets/post_card.dart
+++ b/lib/community/widgets/post_card.dart
@@ -265,6 +265,7 @@ class _PostCardState extends State<PostCard> {
 
     await Navigator.of(context).push(
       SwipeablePageRoute(
+        backGestureDetectionStartOffset: 45,
         canOnlySwipeFromEdge: disableFullPageSwipe(isUserLoggedIn: authBloc.state.isLoggedIn, state: thunderBloc.state, isPostPage: true),
         builder: (context) {
           return MultiBlocProvider(
--- a/lib/inbox/widgets/inbox_mentions_view.dart
+++ b/lib/inbox/widgets/inbox_mentions_view.dart
@@ -45,6 +45,7 @@ class InboxMentionsView extends StatelessWidget {
               // To to specific post for now, in the future, will be best to scroll to the position of the comment
               await Navigator.of(context).push(
                 SwipeablePageRoute(
+                  backGestureDetectionStartOffset: 45,
                   canOnlySwipeFromEdge: disableFullPageSwipe(isUserLoggedIn: authBloc.state.isLoggedIn, state: thunderBloc.state, isPostPage: true),
                   builder: (context) => MultiBlocProvider(
                     providers: [
--- a/lib/inbox/widgets/inbox_replies_view.dart
+++ b/lib/inbox/widgets/inbox_replies_view.dart
@@ -59,6 +59,7 @@ class _InboxRepliesViewState extends State<InboxRepliesView> {
               // To to specific post for now, in the future, will be best to scroll to the position of the comment
               await Navigator.of(context).push(
                 SwipeablePageRoute(
+                  backGestureDetectionStartOffset: 45,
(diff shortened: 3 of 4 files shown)


hjiangsu (approved):
LGTM!

hjiangsu:
I think this change may have inadvertently caused issues on iOS where the following happens (on the post page, not sure about the other areas where this change was introduced):
- When LTR gestures are enabled for comments, it causes swipe to go back to be inconsistent and really hard to trigger
- When swiping to go back from the post itself (not on top of a comment), its behaviour is not consistent and your finger must be farther from the side of the screen to trigger it (which is weird behaviour on iOS since system gestures happen at the edge of the screen)

Would replacing `backGestureDetectionStartOffset` with `backGestureDetectionWidth` work in this case?
https://pub.dev/documentation/swipeable_page_route/latest/swipeable_page_route/SwipeablePageRoute-class.html

hjiangsu:
Also found a related issue on the package repo: https://github.com/JonasWanke/swipeable_page_route/issues/30

micahmo:
Hmm interesting. The issue you linked was exactly our problem in #626, and introducing the `backGestureDetectionStartOffset` seemed to work perfectly on Android. At the edge of the screen, the OS handles the back gesture, and past that, `SwipeablePageRoute` handles the gesture in-app.

I wonder if a much simpler solution would be to put that one property in a conditonal. Like `backGestureDetectionStartOffset: Platform.isAndroid ? 45 : 0`. That's off the dome, so it might not be exactly right. But that would keep our Android fix without regressing anything on iOS. Thoughts?

hjiangsu:
Hmm, I think the conditional can be a quick hotfix to address the issue of not being able to swipe back for iOS, but the issue regarding accidentally triggering comment swipes when attempting to go back will still be an issue with this solution (for Android).

If the user swipes from the edge, that is not captured by the `SwipeablePageRoute`, and has the possibility of being captured by the comment, causing the comment swipe to trigger

For better visualization, I drew up this quick diagram 😅

image

micahmo:
> the issue regarding accidentally triggering comment swipes when attempting to go back will still be an issue with this solution (for Android)

Thanks for the diagram, and good point, that was not entirely solved by the original fix.

I was never able to reproduce the comment issue, so maybe you should just make the change and let @machinaeZER0 test both issues in the next prerelease. 😊

hjiangsu:
I think the best way to fix the issue regarding the swipe triggering "back" twice for Android is to figure out the root cause from the package, and resolve it there instead of here. Although I don't know how difficult that would be 😅

Maybe one solution for now (that you and @machinaeZER0  can test) is to do the following:

```dart
backGestureDetectionStartOffset: Platform.isAndroid ? 45 : 0,
backGestureDetectionWidth: 45,
```

This will fix the issue on iOS of not being able to swipe back (while giving a bit more dead-zone area to perform the swipe back gesture without triggering comment swipe gestures). However, for Android, it will still keep the offset for the start position of the detection which may or may not alleviate some of the problems

machinaeZER0:
I haven't built my own apk before but I can try and take a swing at it if it would be helpful to test! I can (and probably should) look into recording my screen to show how I'm replicating...

PR #629, opened 2023-08-12
Reduce image viewer pan inertia

CTalvio:


micahmo (approved):
This feels better to me (at least on the emulator). Let's get it in the build and get some quick feedback.

PR #643, opened 2023-08-15
Redesign Create Comment Screen

coslu:
## Pull Request Description

This PR redesigns the UI of the create comment screen and adds some additional improvements to the create post screen.

* The create comment screen is no longer displayed as a bottom modal sheet. It is now a page that matches the design of the create post page. 
* Allows to upload images in comments. All image buttons in the create post- and comment pages turn into a `CircularProgressIndicator` while an image is being uploaded.
* When replying directly to a post, the contents of the post are shown in the 'replying to' area. 
* When the create comment page is opened, the text field has focus and the keyboard opens automatically. 
* When switching back from preview, the previous cursor location is maintained and the text field gains focus. In create post page, body text field gains focus.
* Create post- and comment pages show an indicator for the currently active account with avatar, display name, user name and instance.
* Prevents opening the create post- and comment pages when not logged in (was possible through FAB).

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Screenshot_20230815_184921 

## Checklist


lib/community/pages/create_post_page.dart
```diff
@@ -105,12 +105,24 @@ class _CreatePostPageState extends State<CreatePostPage> {
           listener: (context, state) {
             if (state.status == ImageStatus.success) {
               _bodyTextController.text = _bodyTextController.text.replaceRange(_bodyTextController.selection.end, _bodyTextController.selection.end, "![](${state.imageUrl})");
+              setState(() => imageUploading = false);
             }
             if (state.status == ImageStatus.successPostImage) {
               _urlTextController.text = state.imageUrl;
+              setState(() => postImageUploading = false);
```
micahmo, line 112: This is probably a question that @hjiangsu can answer better, but are we mixing styles here by calling `setState` from a `BlocListener`? I believe the builder will be called again when the state changes, so can we check `state.status` in the `IconButton` rather than using a separate property and rebuilding again?

lib/post/pages/create_comment_page.dart
```diff
@@ -0,0 +1,355 @@
...
+                setState(() {
+                  isLoading = true;
+                });
+              }
+              if (state.status == InboxStatus.failure) {
+                setState(() {
+                  isLoading = false;
+                  ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+                    content: Text(AppLocalizations.of(context)!.unexpectedError),
+                    duration: const Duration(days: 1),
```
micahmo, line 164: This snackbar will appear for a whole day?? 

Also a quick note about all snackbars, I'm working on a refactor in #571, so depending on who goes first, I'll include this in my changes or vice versa!

lib/post/pages/post_page.dart
```diff
@@ -188,7 +190,7 @@ class _PostPageState extends State<PostPage> {
                                       : singlePressAction == PostFabAction.changeSort
                                           ? () => showSortBottomSheet(context, state)
                                           : singlePressAction == PostFabAction.replyToPost
-                                              ? replyToPost
+                                              ? () => replyToPost(context)
```
micahmo, line 193: This should already be in `develop`, so once you update your branch, hopefully there will be fewer changes here.

https://github.com/thunder-app/thunder/blob/develop/lib/post/pages/post_page.dart#L191

lib/utils/image.dart
```diff
@@ -121,3 +126,15 @@ Size getWEBPImageDimensions(Uint8List bytes) {
...
+
+void uploadImage(ImageBloc imageBloc, {bool postImage = false}) async {
+  final ImagePicker picker = ImagePicker();
+  XFile? file = await picker.pickImage(source: ImageSource.gallery);
+  try {
+    Account? account = await fetchActiveProfileAccount();
+    String path = file!.path;
+    imageBloc.add(ImageUploadEvent(imageFile: path, instance: account!.instance!, jwt: account.jwt!, postImage: postImage));
+  } catch (e) {
+    null;
```
micahmo, line 138: Do we want to propagate up some error handling?

micahmo (commented):
Looks lovely!

lib/post/pages/create_comment_page.dart
coslu: Right, there I was just trying to make it stay indefinitely until the user gets rid of it. I guess it might not be needed. Should I just remove the duration? I am also aware of the refactor, I can change my snackbars once that is merged.

micahmo: Ahh I see what you're going for. Currently all of the snackbars auto-dismiss, so this would be a new pattern. I might not even think to dismiss it unless it had an `x` button. But of course I'm totally open to a new approach. I'd be interested to hear from others.

lib/community/pages/create_post_page.dart
hjiangsu: Hmm, if we are only using BlocListener, I don't know if we have access to the state from the child. However, if we were using a BlocBuilder or BlocConsumer, then I would agree with this.

I think its okay to set states in here given that we're not using BlocBuilder/BlocConsumer

lib/post/pages/create_comment_page.dart
hjiangsu: The refactor should now be merged in! I think we can just set a long duration on the snackbars for now - enough time so that the message can be read or a screenshot can be taken

hjiangsu (commented):
First of all, great job on this @coslu! This is a pretty big change so thanks for taking it on 🗡️ 

I just added some comments to some of the existing comments to provide my thoughts. Overall, I think this is pretty good, and should be good to go once the comments are addressed/conflicts are fixed!

micahmo (approved):
Since you addressed the comments and fixed the merge conflicts, I'm going to go ahead and merge this so people can start using it. Thanks again!

PR #645, opened 2023-08-16
Do not grey out read own/saved posts in profile

CTalvio:
## Pull Request Description

Very simple cosmetic change, read posts are no longer greyed out when viewed from your own profile, as they are otherwise all always read, unless manually set to unread.

## Checklist

NA


hjiangsu (approved):
LGTM!

PR #648, opened 2023-08-17
Feature/combine post fab

micahmo:
## Pull Request Description

This PR optionally (and by default) combines the post FAB with the comment navigation buttons. This can be turned off on its own or, of course, turning off comment navigation will also show the regular FAB.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/61d52605-e01f-4f1c-88a0-fb0d701bf782

## Checklist


machinaeZER0:
This is beautiful! I would be perfectly happy with this as-is, but throwing an idea out there - instead of containing the FAB icon within the pill shape, what if it was a themed (or just outlined) circle/squircle overlaid on the current pill shape (the latter would match the actual FAB shape so it might be preferable)?

The way I'm envisioning it would make the whole thing a bit wider, and the overlaid FAB would jut out at the top and bottom, so it would definitely obscure more screen real estate - but my thinking is that it would be more obvious that it inherits the properties of the FAB, in that it can be swiped or long-pressed for more options. 

Thoughts? Again, if it doesn't seem like a good idea for Thunder (or should be tackled separately from this PR) then no worries!

micahmo:
Hey @machinaeZER0, thanks for the feedback as always! I tried to capture your idea from your description. Is this what you pictured?

| image |
| - |

If not, maybe you can mock something up.

Either way, it might be nice to get this in as-is and then improve the styling!

machinaeZER0:
Aw, thanks so much for testing that! I was thinking something like that, but maybe I can make a mockup to better encapsulate what's in my brain. I'm away from my computer through the weekend so I'd say stick with what you had for this initial PR and I can keep brainstorming?

ajsosa (approved):

Issue #628, opened 2023-08-12
Instance status on Account Selection


hjiangsu:
**Is your feature request related to a problem? Please describe.**

Idea to add in a general connectivity status for a given instance when selecting an account from the profile selection modal. This would quickly give users that ability to determine if there downtime associated with a given instance (e.g., lemmy.world)

We can use the existing logic which fetches the instance's image so that we don't make too many API calls. If an instance does not respond in time, we can visually add an indicator that the instance may be experiencing downtime.

This would help users hopefully differentiate between problems related to Thunder vs the instance itself, leading to less unrelated bug reports or issues

**Describe the solution you'd like**

We can potentially use a status indicator icon located somewhere in the instance's image to represent the status (red for downtime, green for normal status)

**Describe alternatives you've considered**
N/a

**Additional context**

micahmo:
I love this idea! I've been going to the account selection page and checking the icon as a status indicator. Makes sense to have something more explicit.

shortwavesurfer2009:
> I love this idea! I've been going to the account selection page and checking the icon as a status indicator. Makes sense to have something more explicit.

Ha, i never even noticed this icon before and when this issue mentioned it iwas like "what icon? Oh, that one." A status indicator is a great idea i think.

PR #655, opened 2023-08-17
Feature/instance liveness indicators

micahmo:
## Pull Request Description

This PR has a few changes related to instances on the profile page.

1. Instance icons now dissolve in without any vertical movement.
2. Latency/ping time is shown next to the instance name.
3. A liveness indicator is shown next to the instance icon. (This is purely based on whether we get a valid response for the icon.)

@hjiangsu For the ping to work on iOS, I found that we need to add an extra package and register on startup. I can't test that, so let me know if you can. Also a general note, the ping package doesn't work on Android emulators (not a big deal).

> ~Future enhancement: I would like to make a similar ping check on startup, in case an instance is down, so the user isn't watching the spinner forever and can change instances if they want.~ **Done in this PR.**

## Issue Being Fixed

Issue Number: #628

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/f94c0d26-0ae8-456f-89d6-399341f86cc8

## Checklist


micahmo:
Added a border around the indicator.

image

hjiangsu:
That looks really good! I'm not at my computer to test out iOS but I like the latency indicator :D

micahmo:
One more change: I added special handling for failure to load posts. It mentions the possibility that the user may be offline or the instance may be down. It also offers a direct link to the settings where the user can add/switch accounts/instances. Hopefully this should alleviate issues where Thunder just spins forever on startup when an instance is down. Note that the snackbar will still display the "real" error message for troubleshooting.

https://github.com/thunder-app/thunder/assets/7417301/ef01753b-5e9b-4479-897b-14958a558bc4

hjiangsu:
@micahmo Tested this on iOS and it is also working as expected - I see both the status indicator and the measured ping times for each instance!

PR #666, opened 2023-08-20
Add default user avatar

coslu:
## Pull Request Description
Creates a `UserAvatar` widget similar to the existing `CommunityIcon` and refactors user avatars to use that widget. This makes it so that users without an avatar show a default avatar with the first letter of the display name (if applicable) or the user name.

Also uses `CommunityIcon` inside `CommunityHeader` to ensure communities without an icon show the default icon instead of a blank circle. 

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings
**Before / After**

Screenshot_20230820_205652 Screenshot_20230820_233719

## Checklist


ajsosa (approved):
Nice change 👍

PR #667, opened 2023-08-20
Fixed infinite spinner for load more replies + toast

ajsosa:
## Pull Request Description

Removed an indefinite state change that would occur when clicking the loading more replies comment card.

## Issue Being Fixed

This fixes the infinite spinner that would occur when more replies would fail. This also fixes the issue of the error toast not popping up in that same situation.

## Checklist

- [X] Did you update CHANGELOG.md?


micahmo (approved):
Nice, I've been hitting this a lot. Thanks!!

PR #669, opened 2023-08-21
Feature/post user identifiers

micahmo:
## Pull Request Description

This PR adds the special user identifier chips to the usernames in the post view. To reuse code, some refactoring was done as well.

## Issue Being Fixed

Issue Number: N/A

### Normal User (OP is implied)

| image |
| - |

### Mod

| image |
| - |

### Admin + Mod

| image | 
| - |
## Checklist


CTalvio:
A nitpick, but same as with comments, the chip now offsets the location of the actual username. I notice this.

Should we have the same username deets display in the feed? I feel like it's all getting awfully crowded.

micahmo:
> the chip now offsets the location of the actual username

I think I would rather this than have the chip extend to the left of the margin. Here it wouldn't be a big deal but in comments it would be funny for any content to be to the left of where the comment text is.

> Should we have the same username deets display in the feed? I feel like it's all getting awfully crowded.

Yeah I agree on crowded. Maybe we can revisit. If anything, maybe we can just color the foreground rather than adding a whole chip. Just an idea.

ajsosa (approved):
👍

Issue #491, opened 2023-07-27
Incognito mode


thisisartman:
**Is your feature request related to a problem? Please describe.**
There's no way to browse privately. Sometimes, it's easier to not have the ability to upvote or comment, but having to log out for this every time is cumbersome.

**Describe the solution you'd like**
An additional option in the accounts list for anonymous/incognito mode.

**Describe alternatives you've considered**
The closest alternative to this is to log out of all accounts and then scroll. Obviously, this requires extra steps that a button would solve.

**Additional context**
![Screenshot_20230727-061202.png](https://github.com/thunder-app/thunder/assets/85949613/5b901c72-348f-4cad-a772-1a476353012c)

micahmo:
Related to #269.

Issue #269, opened 2023-07-07
Anonymous account


trymeouteh:
The Infinity app for Reddit is a multi reddit account Android client. You can sign to multiple reddit accounts and you can also use the apps anonymous account feature. The anonymous account is browsing reddit without being in any of your accounts and essentially just like browsing reddit in a inconigo tab and not being logged in.

CTalvio:
How would this work? You'd still have to use some instance or other?

trymeouteh:
> How would this work? You'd still have to use some instance or other?

Not only will you be able to add accounts, you can add an instance.

When browsing an instance, you are not using any account

CTalvio:
Ah. So you'd want to be able to add a entry to the "account" list, with no login? Just an instance. That could work well.

SergeantPanda:
I believe that is how Connect works. You are able to use a 'Guest' account for some instances that you do not want to log in to. It seems to work well.

trymeouteh:
Essentially there would be two kind of accounts. Instances and actual accounts. When you browsing an instance, your not logged into any account. When your browsing under an account, your logged into an account on a specific instance.

Issue #607, opened 2023-08-10
Replace top nav hamburger icon with user avatar when signed in


machinaeZER0:
I've always liked when apps show my avatar in some fashion when signed in, and often that will be done in a way where tapping the avatar will open a menu. I'd personally love to see the hamburger menu icon in top nav replaced with the user avatar when signed in (as a toggle, probably), which would continue to open the sidebar when tapped!

Not to go down a huge rabbit hole, but I could also see this paving the way toward having a little more personal/user-related info in said sidebar, which would then make this update a little more relevant (if eventually there's the ability to switch accounts from the sidebar, for example). Food for thought, but currently outside the scope of this feature request!

PR #677, opened 2023-08-23
Account improvements and anonymous instances

micahmo:
## Pull Request Description

Buckle up, this is a big one! 

This PR introduces a number of changes related to anonymous browsing and user management.

1. Display the current user (or anonymous), including avatar, at the top of the sidebar.
   - Tapping will navigate to the profile page.
   - Pressing the three dots will show the account switcher.
2. Anonymous instances are now supported.
   - Any number of anonymous instances may be added removed and switched to at any time.
3. The account switcher flow has been improved.
   - The current account/instance can be logged out directly in the switcher.
   - Logging out of the current account/instance will switch to (a) the last anonymous instance, if no more accounts, or (b) the last account, if no more instances. (This improves the current behavior where you are "stuck" on the instance of the last account when signing out.)

I tried to be very thorough with my testing, but this is a case where we need a QA team 😆 If anyone wants to grab this branch and provide additional testing, that would be appreciated! I'm especially thinking about potential regressions in anonymous subscriptions (@vbh). But it's a pretty wide-ranging change that may have other regressions as well.

> Note: There will be conflicts between this and #655, and I'd prefer that one to go first.

> Review without whitespace!

## Issue Being Fixed

Issue Number: #269, #491

#### New user section in the drawer

https://github.com/thunder-app/thunder/assets/7417301/e73e718d-738b-42b7-9234-cf2dbdb65eac

#### Add anonymous instances

https://github.com/thunder-app/thunder/assets/7417301/e486be1d-e94d-4c6d-8422-34e422c08200

#### Log out will revert to another account/instance if available

https://github.com/thunder-app/thunder/assets/7417301/2475ef33-5cda-4795-a9e5-a2a579fcc41e

#### Just showing that account/instance switching propagates everywhere properly

https://github.com/thunder-app/thunder/assets/7417301/1105ba6d-0ee6-4b79-8a9b-94985673c6ea

## Checklist


machinaeZER0:
LOVE this!! I'd mentioned in a feature request, but could the hamburger menu icon be replaced with the user avatar with this change? This would serve a dual benefit since you'd be able to see at a glance which account you're using, without even opening the sidebar (assuming the different accounts are using different avatars, at least)

machinaeZER0:
Issue #607 was the one I referenced above

micahmo:
> could the hamburger menu icon be replaced with the user avatar with this change

This is an interesting idea. My concern is that it might be slightly unusual for tapping on an avatar to open a drawer. For example in the Google apps, it opens a little modal which contains mostly account-related actions. The Google apps often also have a separate drawer. Or if you look at Sync, the profile icon is only shown when you disable drawer navigation (and pressing it brings up the account switcher modal), otherwise the hamburger icon is shown.

That said, all these concerns go away if it's an option, so I'm not opposed to the idea. 😉 Since this PR is already so huge, I'd probably want to do is separately.

micahmo:
Ah, nice counter example!! I'm definitely not opposed to this but again would prefer to do things in stages if that's ok. 😊

machinaeZER0:
Absolutely!

machinaeZER0:
Awww but it's so coooool 😄

micahmo:
> This does seem like a pretty big change, so I'll probably hold off on merging this in until after the next general release if thats all good!

I'm good either way. On the one hand, we're "early" in the release cycle (only just published the first prerelease) so we have plenty of time to get this tested. On the other hand, I'm guessing we're going to try to push a new GA faster than usual due being absent for a little while.

lib/thunder/bloc/thunder_bloc.dart
```diff
@@ -317,4 +339,31 @@ class ThunderBloc extends Bloc<ThunderEvent, ThunderState> {
   void _onFabSummonToggle(OnFabSummonToggle event, Emitter<ThunderState> emit) {
     emit(state.copyWith(isFabSummoned: !state.isFabSummoned));
   }
+
+  void _onAddAnonymousInstance(OnAddAnonymousInstance event, Emitter<ThunderState> emit) async {
+    final SharedPreferences prefs = (await UserPreferences.instance).sharedPreferences;
```
hjiangsu, line 344: Just curious, but should we include this when we export/import settings? My thought here is that maybe we should exclude these settings but let me know your thoughts!

hjiangsu (approved):
Oh man, this is a big one - I mainly did a skim through rather than a thorough review on this one, but it looks pretty good overall!

I just had a few points of discussion in some places, but I'm okay with handling those changes in a different PR if needed. Anyways, can you confirm if this is all ready to go to be merged in?

micahmo:
> I mainly did a skim through rather than a thorough review on this one

Haha fair enough! I'm hoping if we get it in early, we can get plenty of feedback.

> can you confirm if this is all ready to go to be merged in?

It should be all set! I recently retested after merging in the latest develop. Let me look at your comments though.

lib/thunder/pages/thunder_page.dart
```diff
@@ -219,6 +219,13 @@ class _ThunderState extends State<Thunder> {
                                         CommunityPage(
                                           scaffoldKey: _feedScaffoldKey,
                                           pageController: pageController,
+                                          navigateToAccount: () {
+                                            _feedScaffoldKey.currentState?.closeDrawer();
+                                            setState(() {
+                                              selectedPageIndex = 2;
+                                              pageController.animateToPage(2, duration: const Duration(milliseconds: 500), curve: Curves.ease);
```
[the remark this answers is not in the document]
micahmo: Good catch! Done.

lib/thunder/bloc/thunder_bloc.dart
micahmo: Yeah I agree. I added both anonymous instance settings to the import/export exclusion list.

PR #681, opened 2023-08-23
Fixed NPE that broke commenting

ajsosa:
## Pull Request Description

Fixed NPE that broke commenting

## Checklist


micahmo (approved):
Is it null when you're leaving a top-level comment and not null when you're replying?

ajsosa:
Yep, that's exactly what's happening.

PR #690, opened 2023-08-26
Fix/dismiss read

micahmo:
## Pull Request Description

This PR fixes an issue where the top unread post would get dismissed after dismissing read posts.

Note that the code I'm removing was deemed to be important based on the original discussion here.

https://github.com/thunder-app/thunder/pull/592#discussion_r1290415509

So I would definitely want @CTalvio to take a look and make sure this is safe. Everything _seems_ fine on my end, but I wasn't part of the original code. Maybe there's a particular test I should run to see if there's any regression.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/bc594e9a-7e98-48ba-9114-341f90069383

### After

https://github.com/thunder-app/thunder/assets/7417301/1e164ba4-4e86-4105-a051-e8d71061c797

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Fixed indefinite state change of `isFetchingMoreComments` when loading more replies. This was also suppressing the snackbar error toast when loading more replies failed - contribution from @ajsosa
 - Fix default community icons not showing in community headers - contribution from @coslu
 - Fixed null pointer exception that broke commenting on posts - contribution from @ajsosa
+- Fix issue with dismissing read posts - contribution from @micahmo
 
 ## 0.2.3+16 - 2023-08-15
 
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -431,11 +431,6 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
           await Future.delayed(const Duration(milliseconds: 60));
         }
       }
-      await Future.delayed(const Duration(milliseconds: 800));
-      setState(() {
-        widget.postViews!.removeWhere((e) => e.postView.read);
-        toRemoveSet.clear();
-      });
       // Load in more posts, if so many got dismissed that scrolling may not be possible
       WidgetsBinding.instance.addPostFrameCallback((_) {
         bool isScrollable = _scrollController.position.maxScrollExtent > _scrollController.position.viewportDimension;
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -431,6 +431,12 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
           await Future.delayed(const Duration(milliseconds: 60));
         }
       }
+      await Future.delayed(const Duration(milliseconds: 800));
+      setState(() {
+        for (var post in toRemoveSet) {
+          widget.postViews!.remove(post);
+        }
+      });
       // Load in more posts, if so many got dismissed that scrolling may not be possible
       WidgetsBinding.instance.addPostFrameCallback((_) {
         bool isScrollable = _scrollController.position.maxScrollExtent > _scrollController.position.viewportDimension;


CTalvio:
If I'm reading this correctly, this merely removes the code that actually removes the read posts from the array they are stored in.

In this state, the feature merely updates a growing list of posts that are not to be displayed, but which are still loaded and now each and every posts is being checked against that list on whether they should be displayed whenever they are cached.

Essentially, its now just an animation that shrinks posts into tiny invisible slivers, which are still there.

Does that list reset on refresh as before?

CTalvio:
The correct way to fix this, as far as I can tell, would be to not use `removeWhere` and instead replace it with a loop of our own, which removes posts based on whether they are in `toRemoveSet` instead of whether they are read.

Or, using `removeWhere` might work with a contains check with `toRemoveSet` instead of a read check.

micahmo:
> the feature updates a growing list of posts that are not to be displayed, but which are still loaded and now each and every post is being checked against that list on whether they should be displayed whenever they are on screen

That's a good point. While it may work on the surface, it'll consume unnecessary resources for all the read posts to continue existing. Let me take another crack at this.

> The correct way to fix this, as far as I can tell, would be to not use `removeWhere` and instead replace it with a loop of our own, which removes posts based on whether they are in `toRemoveSet` instead of whether they are read.
> 
> Or, using `removeWhere` might work with a contains check with `toRemoveSet` instead of a read check.

These both seem reasonable, but I don't think the read check is the problem (otherwise it would imply that `postView.read` is returned incorrectly). In fact, trying both of these suggestions causes all of the read posts to come back afterwards. 😖 I think there's some weird timing issue with building that is the ultimately culprit.

micahmo:
@CTalvio I pushed another variation. What do you think about taking out the call to `toRemoveSet.clear()`? I think the problem is that this list needs to persist between rebuilds so that the removal happens every time. And it seems to be naturally cleared during a refresh since we get a new instance of the widget entirely.

(I would post a demo, but it's the same as "After" above. 😉)

CTalvio:
The reason toRemoveSet should be cleared, is that if it is not, the contains check will still be iterating through its growing entirety for every post that is loaded.

CTalvio:
> These both seem reasonable, but I don't think the read check is the problem (otherwise it would imply that `postView.read` is returned incorrectly). In fact, trying both of these suggestions causes all of the read posts to come back afterwards. 😖 I think there's some weird timing issue with building that is the ultimately culprit.

That makes no sense. How can the posts come back if they are removed from the array? Where are they coming back from? They are re-appearing because they are still in the array when toRemoveSet is cleared, basically it simply reverses the dismiss animation.

And the first post disappearing when removeWhere is executed also makes no sense, the only way it could disappear is if it is indeed removed from the array. Which yes, implies exactly that removeWhere is misfiring, badly.

micahmo:
> The reason toRemoveSet should be cleared, is that if it is not, the contains check will still be iterating through its growing entirety for every post that is loaded.

I hear ya. Again, it's a performance (memory) concern, if nothing else.

> How can the posts come back if they are removed from the array?

I believe it's something along the lines of...
1. Remove the posts
2. Clear the removal list
3. A rebuild is called (not sure why)
4. Everything comes back because the removal list is empty.

The first post coming back I can't explain! But I can say it's still there after the loop, so it's really not a `removeWhere` problem.

CTalvio:
`toRemoveSet` is basically just a way to control the dismiss animation. It does not modify data in any way. Though we do need to use it as reference when we do do that.

It should to be expanded so that only posts in it are actually removed from the post list. Otherwise it is possible to mark a post as read during the animation, and then have it disappear on you. (because the actual removal of posts in logic happens a bit after the merely visual animation of them disappearing)

Only posts that are read in the instant you press dismiss, should actually disappear. toRemoveSet is a list of these posts, and is hence perfect for this.

micahmo:
> It should to be expanded so that only posts in it are actually removed from the post list.

I did push that change as well. But there's still the very odd behavior that if we clear `toRemoveSet`, then either the top post will be removed, or all the read posts will come back. 😕 I will keep working on this unless you have any other ideas!

CTalvio:
> 4\. Everything comes back because the removal list is empty.

The removal list only causes the posts to be hidden, it basically just triggers the animation, which actually merely shrinks the posts. If it is keeping the posts hidden, they are actually still there.

The rebuild doesn't actually _reload_ anything. After the posts are removed from the array, the rebuild is performed to update the view from the now truncated array, which no longer contains the posts that were just removed.

The toRemoveSet is no longer needed to keep them hidden, because they are gone from the **actual** array.

Here is how it should work.

1. Record read posts when dismiss read is pressed
2. Animate the dismissal of those read posts (view _looks_ like array minus read posts)
3. Remove the recorded read posts from the array storing them
4. Rebuild view, transitioning from the appearance of no read posts, to them actually being gone

CTalvio:
That is how it works right now, afaik, except that for some reason, the first item is affected by the changes to the array/list storing the posts.

And it only happens the first time you use the feature. It is extremely weird.

CTalvio:
Ok, I did some testing and reading.

This might have something to do with how flutter handles data and widgets. Due to how it uses the locations of things in trees to match stuff, at times, it is mixing up things. To force it to properly match the things it needs to match, we need to give post items keys, to identify them.

Basically, this shit: https://youtu.be/kn0EOS-ZiIc?si=PSSL1nXdec0xUDc-&t=163

micahmo:
This issue has been fixed by #697, so I'm closing this.

PR #697, opened 2023-08-30
Fix top post missing and dismiss read top post disappearing

CTalvio:
## Pull Request Description

Taglines no longer take the place of the top post in the feed. Conincidentally this also fixes the issue of the top item in the feed "disappearing" when using dismiss read.

## Issue Being Fixed

Supercedes #690

## Checklist
NA

### DIFF
--- a/lib/community/widgets/post_card_list.dart
+++ b/lib/community/widgets/post_card_list.dart
@@ -194,7 +194,7 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
                       },
                       child: CommunityHeader(communityInfo: widget.communityInfo),
                     );
-                  } else if (widget.tagline?.isNotEmpty == true) {
+                  } else if (widget.tagline!.isNotEmpty) {
                     final bool taglineIsLong = widget.tagline!.length > 200;
 
                     return Padding(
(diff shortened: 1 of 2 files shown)


micahmo (approved):
Very nice find! Now I can use dismiss read again! 😆 Thank you!

lib/community/widgets/post_card_list.dart
```diff
@@ -194,7 +194,7 @@ class _PostCardListState extends State<PostCardList> with TickerProviderStateMix
                       },
                       child: CommunityHeader(communityInfo: widget.communityInfo),
                     );
-                  } else if (widget.tagline?.isNotEmpty == true) {
+                  } else if (widget.tagline!.isNotEmpty) {
```
micahmo, line 197: It looks like due to [this line](https://github.com/thunder-app/thunder/blob/736d741129ce2c0d33046acd66efbd32cd75efe5/lib/community/bloc/community_bloc.dart#L266), `tagline` can never be `null`. So we should probably go through and make it not nullable. Otherwise, this looks like a code smell (null assertion operator without null check). But fine for now.

CTalvio: Ah, I tested that but did not leave it in... I think I committed in-between leaving it `tagline!`

CTalvio: Restored, I could go and do that while I'm at it...

micahmo: Hey, sorry, I think I was unclear! At first I thought something was fishy because you used `!` without checking for `null`. But after I investigated where `tagline` is set ([here](https://github.com/thunder-app/thunder/blob/736d741129ce2c0d33046acd66efbd32cd75efe5/lib/community/bloc/community_bloc.dart#L266)), I realized it's actually never `null`, so your null assertion would be safe. I only said it's a code smell because other people might have a similar thought (how do we know it's safe to use null assertion without checking). So I think the better solution would be to make the type non-null entirely (`String` instead of `String?`). Then we wouldn't need either the null assertion or null conditional operator here.

Let me know if that makes sense and if you want to make that change. If not, I'll go ahead and merge this and do it later. Your code is totally fine either way, it just got me thinking. 😆

micahmo:
Perfect, thanks for the updates and thanks again for finding and fixing this!

PR #700, opened 2023-09-01
Add drafts for posts and comments

micahmo:
## Pull Request Description

This PR introduces drafts for posts and comments. Drafts are saved upon exiting the edit screen (without posting) and on a timer (in case the app is quit/crashes). They are saved per community or post/comment.

At this time, only title/url/text are saved.

I also cleaned up some spots that were still using the modal for comments, was currently broken.

I also found that replying to comments with no post context (e.g., inbox replies, user profile comments) is currently broken (to be addressed later).

> P.S. I would like to do undo/redo in the future as well.

## Issue Being Fixed

Inability to save drafts.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/14217e84-6531-4b7d-a750-e4710ecd685c

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -22,8 +22,16 @@ import 'package:thunder/utils/instance.dart';
 class CreatePostPage extends StatefulWidget {
   final int communityId;
   final FullCommunityView? communityInfo;
+  final void Function(DraftPost? draftPost)? updateDraft;
+  final DraftPost? previousDraftPost;
 
-  const CreatePostPage({super.key, required this.communityId, this.communityInfo});
+  const CreatePostPage({
+    super.key,
+    required this.communityId,
+    this.communityInfo,
+    this.previousDraftPost,
+    this.updateDraft,
+  });
 
   @override
   State<CreatePostPage> createState() => _CreatePostPageState();
@@ -36,6 +44,7 @@ class _CreatePostPageState extends State<CreatePostPage> {
   bool imageUploading = false;
   bool postImageUploading = false;
   String url = "";
+  DraftPost newDraftPost = DraftPost();
 
   final TextEditingController _bodyTextController = TextEditingController();
   final TextEditingController _titleTextController = TextEditingController();
@@ -50,12 +59,31 @@ class _CreatePostPageState extends State<CreatePostPage> {
     _titleTextController.addListener(() {
       if (_titleTextController.text.isEmpty && !isSubmitButtonDisabled) setState(() => isSubmitButtonDisabled = true);
       if (_titleTextController.text.isNotEmpty && isSubmitButtonDisabled) setState(() => isSubmitButtonDisabled = false);
+
+      widget.updateDraft?.call(newDraftPost..title = _titleTextController.text);
     });
 
     _urlTextController.addListener(() {
       url = _urlTextController.text;
       debounce(const Duration(milliseconds: 1000), _updatePreview, [url]);
+
+      widget.updateDraft?.call(newDraftPost..url = _urlTextController.text);
     });
+
+    _bodyTextController.addListener(() {
+      widget.updateDraft?.call(newDraftPost..text = _bodyTextController.text);
+    });
+
+    if (widget.previousDraftPost != null) {
+      _titleTextController.text = widget.previousDraftPost!.title ?? '';
+      _urlTextController.text = widget.previousDraftPost!.url ?? '';
+      _bodyTextController.text = widget.previousDraftPost!.text ?? '';
+
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) async {
+        await Future.delayed(const Duration(milliseconds: 300));
+        showSnackbar(context, AppLocalizations.of(context)!.restoredPostFromDraft);
+      });
+    }
   }
 
   @override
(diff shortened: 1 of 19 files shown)


CTalvio:
# YYYYEEEEEEEESSSSSSSSS!!!!!!!!!

lib/community/pages/create_post_page.dart
```diff
@@ -22,8 +22,16 @@ import 'package:thunder/utils/instance.dart';
 class CreatePostPage extends StatefulWidget {
   final int communityId;
   final FullCommunityView? communityInfo;
+  final void Function(DraftPost? draftPost)? updateDraft;
```
hjiangsu, line 25: For the naming of this, I think I prefer `onUpdateDraft` to match similar naming conventions for callback functions!

lib/community/pages/create_post_page.dart
```diff
@@ -50,12 +59,31 @@ class _CreatePostPageState extends State<CreatePostPage> {
     _titleTextController.addListener(() {
       if (_titleTextController.text.isEmpty && !isSubmitButtonDisabled) setState(() => isSubmitButtonDisabled = true);
       if (_titleTextController.text.isNotEmpty && isSubmitButtonDisabled) setState(() => isSubmitButtonDisabled = false);
+
+      widget.updateDraft?.call(newDraftPost..title = _titleTextController.text);
```
hjiangsu, line 63: Do you think we should add a debouncer for these calls to update drafts? One downside is that it might not save the exact draft since it would be on a debouncer, but the upside is that we potentially reduce the number of calls to `updateDraft` for all cases (title, url, body, etc)

lib/community/widgets/community_sidebar.dart
```diff
@@ -129,11 +136,26 @@ class _CommunitySidebarState extends State<CommunitySidebar> with TickerProvider
...
                                   child: ElevatedButton(
                                     onPressed: isUserLoggedIn
-                                        ? () {
+                                        ? () async {
                                             HapticFeedback.mediumImpact();
                                             CommunityBloc communityBloc = context.read<CommunityBloc>();
                                             AccountBloc accountBloc = context.read<AccountBloc>();
                                             ThunderBloc thunderBloc = context.read<ThunderBloc>();
+
+                                            SharedPreferences prefs = (await UserPreferences.instance).sharedPreferences;
```
hjiangsu, line 145: It seems like this logic is mostly the same for a lot of the changes - could we abstract this a bit more to reduce code duplication? It looks like the main difference is the `draftId` but let me know if this is not possible!

lib/core/enums/local_settings.dart
```diff
@@ -94,6 +94,8 @@ enum LocalSettings {
   postFabLongPressAction(name: 'settings_post_fab_long_press_action', label: ''),
   enableCommentNavigation(name: 'setting_enable_comment_navigation', label: 'Enable Comment Navigation Buttons'),
   combineNavAndFab(name: 'setting_combine_nav_and_fab', label: 'Combine FAB and Navigation Buttons'),
+
+  draftsCache(name: 'drafts_cache', label: ''),
```
hjiangsu, line 98: One thing to note about here is that we probably don't want to save the drafts that get created when we export/import the user settings to JSON!

hjiangsu (commented):
First of all, thanks for taking this on! This is going to be a really nice QoL feature to have :D

I just left a few comments on some parts to consider, but other than those, I think it was done pretty well! One test I would also try out is import/exporting settings and see if the drafts also come with it. I think the appropriate behaviour there is to ensure that we ignore the drafts when we export/import settings

lib/community/pages/create_post_page.dart
micahmo: Done.

lib/community/pages/create_post_page.dart
micahmo: I wasn't thinking so in this case because the callback is not actually persisting the draft, but rather just updating the in-memory representation at a higher level. Take a look at the implementation of the callback.

https://github.com/thunder-app/thunder/blob/2a51731b7eb176aa83e73537d2248847b878eb34/lib/community/widgets/community_sidebar.dart#L172

The actual persisting happens either when closing the editor without publishing or on a 10-second timer, so the impact there should be pretty minimal.

https://github.com/thunder-app/thunder/blob/2a51731b7eb176aa83e73537d2248847b878eb34/lib/community/widgets/community_sidebar.dart#L153-L157

What do you think?

lib/community/widgets/community_sidebar.dart
micahmo: Yeah you're right there already was and now is even more duplication among all the different places where we can create posts/comments. I totally agree about trying to reduce that, although I think it might be tricky in this case. Everything defined here is needed later. For example, the draft objects are updated by the callback, the timer is canceled after the navigation push, etc. I could try to refactor the whole section (a common way to create posts/comments) but different areas use different Blocs, etc. I would be more inclined to take this as a separate task where we can re-test everything. But that's up to you!

lib/core/enums/local_settings.dart
micahmo: Good point! I added a mechanism to exclude settings from import/export, and I excluded drafts.

lib/community/pages/create_post_page.dart
hjiangsu: I see! That makes sense then - I was mostly worried about performance impacts on updating the shared/local preferences too many times. Just updating the memory should be okay though!

lib/community/widgets/community_sidebar.dart
hjiangsu: Sounds good - if the logic is tied to other parts, then that makes sense. We can think about a refactor of the common aspects of posts/comments in another PR

hjiangsu (approved):

PR #710, opened 2023-09-06
Improve user colors

micahmo:
## Pull Request Description

This PR adjusts the special user chip colors. Currently many of the colors looks very similar (e.g., OP, self, and mod are all shades of ~blue~ the current theme color), so this should provide some distinction while still following the current theme.

## Issue Being Fixed

Issue Number: N/A

### Before

| pvmY5Iz7Go |
| - |

#### OP / Self / Mod in Different Themes

| Header | Header | Header |
|--------|--------|--------|
| iKSfKyFNHd | XzQKcWFaXM | L7CdMahq42 |
| 4XhZ6JVo2P | 4Q3e6ugoU5 | 8AfsNndL8s | 

#### Bot

| Header |
|--------|
| t3fP4zbWKs |
| CovW8LG1SX | 

#### Admin

| Header |
|--------|
| image |
| hQtoz96rZu | 

## Checklist


hjiangsu:
I think it looks good! Maybe others have suggestions or feedback on the general colours?

@machinaeZER0 @tom-james-watson @CTalvio @ajsosa

machinaeZER0:
I wouldn't be opposed to having these match what's used for lemmy on the web side, and/or staying consistent across theme choice so users are on the same page regarding identification by color at a glance.

I do also like the proposed implementation for this use case though, and I think it does probably make sense more than slightly different shades of the theme color.

micahmo:
> I wouldn't be opposed to having these match what's used for lemmy on the web side

As far as I can tell, we are following the web to an extent.
* bot: seems to be a shade of gray, so I took some liberty here with the purple image
* mod: shade of orange/brown image
* admin: shade of red image
* self: none?
* OP: shade of blue image


> staying consistent across theme choice so users are on the same page regarding identification by color at a glance.

I think if you look at the side-by-side images, the variance across themes is pretty minor. It's enough to make the app feel fresh if you change theme, but not so much that the identification is lost (i.e., it's clearly still green/red/blue/etc.). We followed the same approach for the post type badges.

CTalvio:
The only thing I might add, is that I've been thinking it might be fun to use a gradient of the multiple colors, when someone has more than one thing.

micahmo:
> The only thing I might think to try, is that I've been thinking it might be fun to use a gradient of the multiple colors, when someone has more than one thing.

That's an interesting idea! I agree that having only one color is a bit of a limitation, although (a) the priority of colors has been intentionally selected and (b) the icons and tooltip still provide all the information.

hjiangsu:
It seems like everyone is in agreement so I'll go ahead and merge this in!

> The only thing I might think to try, is that I've been thinking it might be fun to use a gradient of the multiple colors, when someone has more than one thing.

This is an interesting idea, and we can revisit this at a later time!

hjiangsu (approved):

PR #712, opened 2023-09-06
Added dimension information to thumbnails to reduce layout jank when using full screen images option

hjiangsu:
## Pull Request Description

When there is a thumbnail image from the API, retrieve the image's width and height dimensions. This reduces a lot (but not all) layout jank that occurs when scrolling through the feed.

To compare the differences, run the latest release build of Thunder and turn on "View Full Height Images" option in the settings. Go back to feed, refresh if needed, and scroll through the feed (both up and down the feed)

With the latest version, you should be seeing a lot of layout jank. Using this fix should drastically reduce that layout jank. This is only tested on iOS, so some tests should be done on Android as well!

Please note that this is only really visible when you are using a release build, not debug builds.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Unable to upload screen recording on device :(

## Checklist

### DIFF
--- a/lib/utils/post.dart
+++ b/lib/utils/post.dart
@@ -131,7 +131,15 @@ Future<PostViewMedia> parsePostView(PostView postView, bool fetchImageDimensions
   } else if (url != null) {
     if (fetchImageDimensions) {
       if (postView.post.thumbnailUrl?.isNotEmpty == true) {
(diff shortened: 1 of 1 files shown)


micahmo (approved):
LGTM! And seems to work well on Android.

PR #714, opened 2023-09-07
System back goes to history on profile

micahmo:
## Pull Request Description

This PR taps into the system back button to navigate from "Saved" to "History" on the profile page. This is an Android-specific thing, but I definitely agree that it makes sense for Back to undo the most recent navigation action, even within a page.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/98d91968-e854-4c96-b5a9-6396918afe74

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Star indicator for saved posts now prefixes the post title so that it's consistent with the indicators for locked posts and featured community posts - contribution from @ajsosa
 - Improved ability to refresh posts - contribution from @micahmo
 - Improve the option selector dialog to show the currently selected item - contribution from @micahmo
+- Use Android system back button to navigate from Saved to History on profile page - contribution from @micahmo
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


hjiangsu:
If this is an Android-specific thing, then I would wrap the `setState` logic with the Android platform!

e.g., 
```dart
if (Platform.isAndroid) {
  setState(() => { ... });
}
```

micahmo:
> If this is an Android-specific thing, then I would wrap the `setState` logic with the Android platform!

Well, maybe I shouldn't say it's Android-specific. Any OS that has a back button as recognized by the `BackButtonInterceptor` package should use this functionality. 😊 Realistically it's probably just Android, but it's nice to let the package handle that detection for us. I think the less custom code the better. Also note that we use `BackButtonInterceptor` several other places already and don't have explicit OS checks.

https://github.com/thunder-app/thunder/blob/4d91b899f2de88b146898323bef37bd853694db9/lib/community/pages/community_page.dart#L70

https://github.com/thunder-app/thunder/blob/4d91b899f2de88b146898323bef37bd853694db9/lib/community/pages/community_page.dart#L494-L499

hjiangsu:
Ahh okay, that makes a bit more sense!

hjiangsu (approved):

Issue #19, opened 2023-06-19
Downvote button is displayed on instances where downvotes are disabled 


jordan1776:
**Describe the bug**
Downvote button is displayed on instances where downvotes are disabled (beehaw.org for example). 

**To Reproduce**
1. Downvote a post on beehaw.org
2. DioException [Bad Response] The request returned an invalid status code of 400.

**Expected behavior**
Either not display the button or have it clearly greyed out.

**Smartphone (please complete the following information):**
 - Device: Xperia 1 II
 - OS: Android 11

PR #717, opened 2023-09-07
Disable or hide downvote actions when instance does not have downvotes enabled

hjiangsu:
## Pull Request Description

This PR hides downvote buttons and disables downvote actions when an instance does not have downvotes enabled. The following changes have been added/tested:
- Swipe gestures for downvotes for both posts and comments will throw an error indicating the instance has disabled downvotes
- In the comfortable view, the downvote icon is hidden when downvotes are disabled
- The comment button action downvote button is hidden when downvotes are disabled
- The post page downvote action is disabled

## Issue Being Fixed

Issue Number: #19 

## Screenshots / Recordings
| Screenshots  |
| ------------- |
| simulator_screenshot_D8D6A0D9-DAE1-43EF-8233-E9BC41D3D106  |
| simulator_screenshot_6D2771A9-121D-40A6-B32B-663859AC48B9  |
| simulator_screenshot_4C823B94-3D33-4A4C-B0E4-DE4C722AD2B4 |
| simulator_screenshot_4EC1D842-1372-4D8D-A720-B455C5E7BE0C |

## Checklist

- [ ] Did you add `semanticLabel`s where applicable for accessibility? (N/A)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,32 @@ jobs:
 
       - name: Tests
         run: flutter test
+
+      # Read the version number
+      - name: Read pubspec.yaml version
+        id: output-pubspec-version
+        uses: NiklasLehnfeld/flutter-version-number-action@main
+        with:
+          file-path: pubspec.yaml
+
+      # Read the globals.dart file
+      - name: Read globals.dart version
+        id: output-globals-version
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: lib/globals.dart
+
+      # Get just the first line of the globals.dart file
+      - name: Get first line of globals.dart file
+        uses: jungwinter/split@v2
+        id: split-globals-output
+        with:
+          msg: ${{ steps.output-globals-version.outputs.content }}
+          separator: '\n'
+
+      # Assert that the version has been updated in globals
+      - name: Assert globals.dart has correct version
+        uses: nick-fields/assert-action@v1
+        with:
+          expected: const String currentVersion = '${{ steps.output-pubspec-version.outputs.version-number }}';
+          actual: ${{ steps.split-globals-output.outputs._0 }}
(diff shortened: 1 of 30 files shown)


micahmo (approved):
The code looks good to me!

The only question I'd have is whether there's an inconsistency between hiding the button in some places and disabling it in others. I don't have a strong preference, but just wanted to mention it as a point of discussion. Otherwise, LGTM!

hjiangsu:
Just noticed that I was missing logic for checking the downvote setting when we switch accounts, so I'll add that in first before merging it in

> The only question I'd have is whether there's an inconsistency between hiding the button in some places and disabling it in others

I thought a bit about this too - I think hiding the buttons themselves make sense since that isn't an action that should be performed (in hindsight, maybe I can also just hide the downvote icon/value on the post too). However, for things like swipe gestures, I decided to keep the downvote action in and just show an error because the swipe gesture setting is somewhat of a "global" setting.

It might potentially cause some confusion if the user has accounts on different instances where the downvote is enabled/disabled, and the swipe gesture settings change when they switch between instances

micahmo:
> I think hiding the buttons themselves make sense since that isn't an action that should be performed (in hindsight, maybe I can also just hide the downvote icon/value on the post too).

Sounds good, that's what I was getting at. If we're going with hiding, we should be able to hid on the post too. I think it's a flex layout so hopefully everything should still look fine.

> However, for things like swipe gestures, I decided to keep the downvote action in and just show an error because the swipe gesture setting is somewhat of a "global" setting.

Yep, totally agreed on the decision here here. My comment was less about swipe gestures.

hjiangsu:
Alright, updated some logic to handle checking instance settings when adding/switching accounts, and also removed downvote on post page @micahmo 

simulator_screenshot_09C19558-D060-4F22-AB7A-ACA0A7529143

micahmo (approved):
LGTM!

Just a mental note that I will have to test this functionality with anonymous instances in #677.

**EDIT:** Well, duh, you can't vote at all if you're not logged in! Confirmed, these changes work fine with and don't conflict at all with #677.

PR #720, opened 2023-09-07
Fix double tap zoom bug

CTalvio:
## Issue Being Fixed

When attempting to pan immediately after using double tap zoom, the double tap zoom is triggered instead. This PR fixes that, so that is possible to immediately pan after ending a zoom.

## Checklist


micahmo (approved):
Ahh very nice! This one's been bugging me a bit. 😆 Thanks!

PR #724, opened 2023-09-09
Hide community name and show username in community view

### LINKED ISSUE #303: [Request] hide community name on posts when viewing community?

micahmo:
## Pull Request Description

This PR fixes #303 by hiding the community name in the post metadata when we're viewing a specific community. Also as suggested by [this comment](https://github.com/thunder-app/thunder/issues/303#issuecomment-1627792641), in order to fill the empty space, the username is shown instead.

## Issue Being Fixed

Issue Number: #303

## Screenshots / Recordings

| image |
| - |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Improve contrast and distinction of special user identifiers - contribution from @micahmo
 - Show swatches and live previews for accent color selection - contribution from @micahmo
 - Use Android system back button to navigate from Saved to History on profile page - contribution from @micahmo
+- Hide community name and show usernames when viewing a community - contribution from @micahmo
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


hjiangsu (approved):
LGTM!

PR #723, opened 2023-09-08
Refactor navigation to community/user pages, and fixes to iOS swipe back not working properly

hjiangsu:
## Pull Request Description

- Refactors some code which handles navigation to CommunityPage/UserPage
  - `navigateToCommunityPage` and `navigateToUserPage` replaces the previous functions for navigating to community/user pages by name. It is now more generic and will accept the name or the id of the community/user
- Increased width for swipeable page route detection
- Moved swipeable page route offset back to default for iOS, but kept it the same for Android
- CreateCommentPage and CreatePostPage are no longer full screen swipe to go back. This fixes an issue where it is really difficult to horizontally scroll on the markdown options at the bottom of the screen 

Test out the following if possible!
- Navigating to community/user pages from all the possible places (from long press actions, tapping on usernames/communities, user and community sidebars, etc)
- Test out comment swipe detection to see if that's been improved (hopefully should trigger accidental comment swipes less often)

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #704, opened 2023-09-05
Improve block management

micahmo:
## Pull Request Description

This PR adds a few improvements around block management.

1. Localize account settings page.
2. Show community and user icons in block settings.
3. Show display names.
4. Make users/communities tappable to visit.
5. New ability to add blocks directly from the settings page.
   - Including a new user/community picker dialog with support for autocomplete/search or direct name (via all of the syntax that we support in link detection).

> P.S. I hope the username autocomplete can be used in the post/comment editor at some point (i.e., #256).

## Issue Being Fixed

Issue Number: N/A

### Before

| image |
| - |

### After

https://github.com/thunder-app/thunder/assets/7417301/d3b7cbf6-d8d1-4e48-aa7e-861f89e83a93

## Checklist

- [ ] Did you update CHANGELOG.md?
- [x] Did you use localized strings where applicable?
- [x] Did you add `semanticLabel`s where applicable for accessibility?

PR #726, opened 2023-09-10
Fix breaking change from conflicting merges

micahmo:
## Pull Request Description

PRs #704 and #723 had some logical conflicts, so this PR resolves that and gets `develop` building again.


hjiangsu (approved):
Oops, I did it through GitHub so it didn't do the CI checks until after the two were merged! Thanks for fixing it!

PR #691, opened 2023-08-26
Remove thumbnail progress indicators

micahmo:
## Pull Request Description

This PR removes some usage of loading indicators (`CircularProgressIndicator`).

Currently, there are loading indicators for images and link previews. On Android, these indicators are incosistent in size and thickness, and on iOS, they're totally different styles. They give the general impression that things are constantly loading. In other apps, there are no such indicators; images just fade in when available. In my opinion (again, this is just a matter of preference) that looks much cleaner.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/d54b1fd0-0912-4068-bfce-740f52f0be85

### After

https://github.com/thunder-app/thunder/assets/7417301/84388673-0a2c-4b8b-83ca-9a9f5d234880

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #727, opened 2023-09-11
Restore the share/downloading indicator

micahmo:
## Pull Request Description

I was a little too aggressive in #691 and accidentally removed the `CircularProgressIndicator` that shows when you share an image and it has to download.

| image |
| - |


hjiangsu (approved):
LGTM!

Issue #163, opened 2023-07-01
For the overscroll function, please follow the system and don't override


Coffee-Viking:
**Is your feature request related to a problem? Please describe.**

Pulling down when reaching the top of the screen makes the text and all menus stretch, this is a function Android refers to as "overscroll".

People with vision impairment however can get dizzy, so it can be disabled:

https://support.google.com/pixelphone/thread/132936922/how-to-turn-of-android-12-overscroll?hl=en

Thunder however overrides the system settings (regardless whether it's disabled via developer options or accessibility settings) and overscrolls regardless.

**Describe the solution you'd like**

Please follow the system default, or implement a manual switch to disable it.

**Describe alternatives you've considered**

None possible.

hjiangsu:
I'll have to see what is causing this to happen - I'm not too sure how Flutter handles that setting, so this might take a bit more time to implement and fix

Coffee-Viking:
Got it, appreciate the quick response!

hjiangsu:
Sorry that this has taken so long to get to, but this is now in-progress! The solution that I'm working towards is to have a setting to reduce the animations used by Thunder.

Thunder itself (as far as I'm concerned), is not able to read the system settings, so this is my plan: introduce a new setting which reduces animations. When enabled, certain animations will be reduced/removed:
- Overscrolling effects from lists (feed, posts, etc)
- Potentially completely removing page transition animation when switching between tabs (in-discussion on Matrix)

Please let me know if there are any other potential animations that you would like to see removed or reduced when this option is on!

cc: @micahmo, @ajsosa, @CTalvio

micahmo:
I like that plan! Here are a few more potential animations to eliminate in that mode. Not sure these are all equally important, they just came to mind.

* Navigation change (e.g., loading post)
* Comment expand/collapse
* Bottom modal

hjiangsu:
Do you think we should completely remove the navigation change animation, or just drastically reduce the time it takes? I did a quick test and reducing it to zero makes it feel a bit jittery.

micahmo:
Hard to say (even whether we should touch that one at all). Android has an animation speed setting which I know many people like to lower to make the phone feel more snappy. On the other hand, iOS has a reduce motion setting which turns many animations into dissolves. If this is about accessibility, then maybe the latter makes more sense. Then again, I don't know if we can turn SwipeablePageRoute into a dissolve.

hjiangsu:
[...]

PR #730, opened 2023-09-12
Add Thunder setting to reduce animations throughout the app

hjiangsu:
## Pull Request Description

This PR introduces a new settings section "Accessibility" which holds a single option (at this time) which reduces the animations used in Thunder. Currently, the following things have changed:
- Bottom navigation bar animation is removed when the setting is toggled on
- Navigation change (e.g., loading post) animation has been reduced drastically
- Overscroll stretching effects have been removed from the feed and post pages (replaced with the iOS equivalent)

## Issue Being Fixed

Issue Number: #163 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Add option to disabling graying out read posts - contribution from @micahmo
 - Show sort type icon - contribution from @micahmo
 - Downvote actions will be disabled when instances have downvotes disabled
+- Added accessibility settings to reduce animations/motion
 - Automatically save drafts for posts and comments - contribution from @micahmo
 
 ### Changed
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -8,6 +8,8 @@ PODS:
     - Flutter
   - flutter_icmp_ping (0.0.1):
     - Flutter
+  - flutter_keyboard_visibility (0.0.1):
+    - Flutter
   - flutter_native_splash (0.0.1):
     - Flutter
   - FMDB (2.7.5):
@@ -43,6 +45,7 @@ DEPENDENCIES:
   - flutter_custom_tabs (from `.symlinks/plugins/flutter_custom_tabs/ios`)
   - flutter_file_dialog (from `.symlinks/plugins/flutter_file_dialog/ios`)
   - flutter_icmp_ping (from `.symlinks/plugins/flutter_icmp_ping/ios`)
+  - flutter_keyboard_visibility (from `.symlinks/plugins/flutter_keyboard_visibility/ios`)
   - flutter_native_splash (from `.symlinks/plugins/flutter_native_splash/ios`)
   - gallery_saver (from `.symlinks/plugins/gallery_saver/ios`)
   - image_picker_ios (from `.symlinks/plugins/image_picker_ios/ios`)
@@ -70,6 +73,8 @@ EXTERNAL SOURCES:
     :path: ".symlinks/plugins/flutter_file_dialog/ios"
   flutter_icmp_ping:
     :path: ".symlinks/plugins/flutter_icmp_ping/ios"
+  flutter_keyboard_visibility:
+    :path: ".symlinks/plugins/flutter_keyboard_visibility/ios"
   flutter_native_splash:
     :path: ".symlinks/plugins/flutter_native_splash/ios"
   gallery_saver:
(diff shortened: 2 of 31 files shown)


micahmo:
Ooh very nice! I actually quite like the faster navigation animation.

Unfortunately, the overscroll is still there on Android. 😞

https://github.com/thunder-app/thunder/assets/7417301/61c39ec5-53fb-48ff-88a7-8c5ad7b3fe64

hjiangsu:
Dang, does turning on the setting reduce the overscroll effect then? Or is it the same as if the setting was off?

I thought this would help, but maybe its specific to iOS: https://stackoverflow.com/questions/58645048/how-to-remove-overscroll-on-ios

hjiangsu:
Looking into this a bit deeper, it seems like this is what is being applied on Android to show the stretching from the overscrolling effects: https://api.flutter.dev/flutter/widgets/StretchingOverscrollIndicator-class.html

For us to potentially remove this effect, we would need to switch over to using https://api.flutter.dev/flutter/widgets/GlowingOverscrollIndicator-class.html when the reduce animation setting is toggle on. This is only an Android specific issue, as iOS does not have the overscroll effects when scrolling so we need to keep that in mind! 

I haven't been able to implement the change yet but it seems like this should do the trick

hjiangsu:
Just tested it out with the `GlowingOverscrollIndicator` and it did not work as I expected it to work. Instead, I changed the physics to use `BouncingScrollPhysics` when reduced animations are on. What this does is that it allows you to scroll past the content of the page, without activating the overscroll stretch effect. This is how iOS seems to handle it by default!

@micahmo could you test this out and see if this makes sense and removes the overscroll effects?

micahmo:
Nice, seems to work fine for me on Android!

As for whether it's better, I guess that's a question for @Coffee-Viking or someone else who has an idea of animations as it relates to accessibility. On the one hand, this change introduces much more motion on the screen, but on the other hand, things aren't stretching. So I can't say which is better.

hjiangsu:
I think the main thing here is to remove that stretching effect that is caused when overscrolling because it causes some users to become nauseous when seeing that.

If everything is uniform and all the content moves uniformly across the screen, I think that would reduce that motion effect and hopefully reduce the nauseating feeling! My guess (but feel free to correct me here) is that reduce motion should reduce any "non-user" generated motion such as the stretch motion and navigation motion!

micahmo:
That all makes sense to me!

hjiangsu:
@micahmo This should be ready for review now! Could I also get you to test something out when you have the chance? 

Could you try to change your system animation/motion settings on Android and see if that affects the general animation speed for parts of the app? This includes the native animations for widgets such as expanding/collapsing sidebar, opening up bottom modals, etc (so nothing where we set the duration of a animation or transition)

Thanks! :D

micahmo (approved):
> This should be ready for review now!

The code LGTM! Very nice!

> Could you try to change your system animation/motion settings on Android and see if that affects the general animation speed for parts of the app?

You bet. I actually run with my animations at .5x on my physical device and haven't noticed that Thunder follows this. Just tested again and can confirm nothing changes with relation to the animation multiplier in Android (either in `develop` or in this branch with animations disabled).

micahmo:
Actually, found one more place you might want to change when animations are disabled. This is when pressing the back button on Android, we scroll through the pages to the home feed.

https://github.com/thunder-app/thunder/blob/adde240d9bddb59362eae31d8d3f6c5daca79c50/lib/thunder/pages/thunder_page.dart#L116

hjiangsu:
> You bet. I actually run with my animations at .5x on my physical device and haven't noticed that Thunder follows this. 

Yeah, it seems like developer settings to set animation speed is not available in Flutter: https://github.com/flutter/flutter/issues/10101#issuecomment-398940831

Does Android have any accessibility settings (outside of the developer settings) which reduces motion in any way? For example, iOS has Reduce Motion setting in Accessibility. If so, does setting any of those settings affect the animations within Thunder?

> Actually, found one more place you might want to change when animations are disabled. 

Nice catch! I'll add some logic in there as well to handle this

micahmo:
> Does Android have any accessibility settings (outside of the developer settings) which reduces motion in any way?

Yes, I just found a setting called "Remove animations". The funny thing is, when I first turned it on, Thunder did respect it (e.g., the bottom modal popped up immediately). But then after restarting Thunder, it went back to animations again. So I'm not sure what to make of that. 🤷

hjiangsu:
> Actually, found one more place you might want to change when animations are disabled.

This should be good now! If you could test this out yourself, that would be helpful (I tested on emulator and it seemed to be okay)

> But then after restarting Thunder, it went back to animations again.

Hmm, might just be a Flutter thing then. I think we can leave that be for now. Thanks for testing it by the way!

micahmo:
Looks good to go!

PR #733, opened 2023-09-13
Show all releases via in-app update

micahmo:
## Pull Request Description

This fixes the issue where pressing the in-app update prompt would only show the latest stable release instead of all releases.

Note that this technically opens a page with _all_ of the releases, but the latest will be at the top. (There doesn't seem to be a way to directly link to the latest prerelease.)

## Issue Being Fixed

In-all link was going to the latest stable release.

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/models/version.dart
+++ b/lib/core/models/version.dart
@@ -2,6 +2,7 @@ class Version {
   final String version;
   final bool hasUpdate;
   final String? latestVersion;
+  final String? latestVersionUrl;
 
-  Version({required this.version, this.hasUpdate = false, this.latestVersion});
+  Version({required this.version, this.hasUpdate = false, this.latestVersion, this.latestVersionUrl});
 }
--- a/lib/core/update/check_github_update.dart
+++ b/lib/core/update/check_github_update.dart
@@ -30,13 +30,14 @@ Future<Version> fetchVersion() async {
     if (response.statusCode == 200) {
       final release = json.decode(response.body);
       String latestVersion = release[0]['tag_name'];
+      String latestVersionUrl = release[0]['html_url'];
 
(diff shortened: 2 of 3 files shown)


hjiangsu:
Hmm, I was thinking maybe we could obtain the url for the latest version that we found using GitHub's API.

In `fetchVersion()`, we use GitHub's API to fetch the list of all releases. It seems like we might be able to get the html URL for that given release as well!

e.g, https://api.github.com/repos/thunder-app/thunder/releases seems to contain `html_url` attribute in its response that we could potentially use!

micahmo:
Good point! Updated to use the API and get the specific latest release URL.

hjiangsu:
Just one suggestion here!

Since `showUpdateNotification(BuildContext context, Version? version)` gets passed in a `Version` object, I wonder if it might be better if we add an attribute to `Version` so that we can use that rather than to have to perform a separate query to fetch the URL

This way, we can both do the checks for updates, and also place the URL when we run `fetchVersion()`. I hope this makes sense (if not, just let me know and I can clarify)

For reference:
- https://github.com/thunder-app/thunder/blob/develop/lib/core/models/version.dart
- https://github.com/thunder-app/thunder/blob/7af18c882675f5fed0b35dfde0b1689ab02cf44d/lib/core/update/check_github_update.dart#L21
- https://github.com/thunder-app/thunder/blob/7af18c882675f5fed0b35dfde0b1689ab02cf44d/lib/thunder/pages/thunder_page.dart#L319

micahmo:
Alrighty, let me know what you think now.

hjiangsu (approved):
LGTM!

PR #737, opened 2023-09-14
Fix swipe drafts overwrite

micahmo:
## Pull Request Description

This is a tiny fix to the new drafts feature which prevents a problem related to creating comments from the swipe gesture where every comment would essentially have the same global draft. The problem was that `selectedCommentId` was null, so we needed to get the real id in order for the drafts to be unique.


hjiangsu (approved):
LGTM!

Issue #487, opened 2023-07-26
Jittery upward scrolling with full height images enabled


murtaza64:
**Description**  
After enabling the "full height images" feature, scrolling upwards in the feed results in noticable jitter. 

**How to Reproduce**  
1. Enable full height images
2. Scroll upward in the main feed

**Device & App Version:**  
 - Device: Samsung Galaxy s22 ultra
 - OS: android
 - Version: 13

CTalvio:
Duplicate #395  #404

K4LCIFER:
I appear to have the same issue:

https://github.com/thunder-app/thunder/assets/7829231/d0239014-04c4-4bae-bd45-541f0352d409

hjiangsu:
@K4LCIFER  I'm unable to reproduce the exact jitteriness that is shown in the video. Could you try the latest pre-release version and let me know if this issue is still occurring?

As for general jank (random jumps when scrolling), this is most likely caused by the link previews in the feed. Currently, we don't have checks to determine the height of the images present within links. Because of this, when a link image is loaded, it can push down content when it renders on the screen causing the jank.

One way we can reduce this is to set a static height for all link preview images. This way, the feed can pre-allocate the correct amount of space needed and thus reduce the jank present. Alternatively, we would need to figure out a way to calculate the proper heights of each of those images and pre-allocate the space

K4LCIFER:
> Could you try the latest pre-release version

Which exact version are you referring to?

> As for general jank (random jumps when scrolling), this is most likely caused by the link previews in the feed. 

I don't think that what I am experiencing is due to links loading in their thumnails thereby changing the length of the feed -- if that is what you mean. The reason that I say this is because if the feed is left to load, the issue remains. So even when all the thumbnails are loaded, the issue will still be there.

hjiangsu:
@K4LCIFER This version: https://github.com/thunder-app/thunder/releases/tag/v0.2.1-alpha.2%2B14

K4LCIFER:
If it's alright, I will wait until the next full release at which point I will report back to this thread.

Fmstrat:
New version fixes it for me. Though I have learned the in-app notification doesn't do pre-releases. Will open a separate issue for that.

K4LCIFER:
@Fmstrat was your issue the exact same as mine?

Fmstrat:
Yes.

Fmstrat:
[...]

PR #745, opened 2023-09-15
Reintroduce caching with a limit

micahmo:
## Pull Request Description

PR #463 introduced caching for `ExtendedImage`. PR #576 reverted that change. While we have done several things to reduce the jankiness of reloading images (e.g., removing Hero, removing spinners, calculating height), there are still some times when the image reloads unexpectedly and it can be a bit jarring/annoying. One such example is when marking a post as unread, of all things. There's also still the "issue" where if you scroll out of frame and scroll back, all the images have to reload again (discussed [here](https://github.com/thunder-app/thunder/pull/576#issuecomment-1668829168)). Finally, there is some discussion in #487 that caching may help to prevent that issue as well.

In this PR, I have first of all set `clearMemoryCacheWhenDispose` back to `false`. But more importantly, I have also set `cacheMaxAge` to 1 minute. This should give us enough time that common operations (scrolling up/down, marking unread) will show the cached image, but the cache will also expire relatively quickly so that prolonged usages of the app will not maintain more than a minute's worth of image cache. (And of course, we can discuss this value.)

Let me know your thoughts on this approach! And maybe we can do some profiling too.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/1097cd05-0c71-4264-8a26-ff0650d0c6a2

### After

https://github.com/thunder-app/thunder/assets/7417301/7a260ebf-1268-46c8-8e8b-64a8b203534f

## Checklist

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -114,7 +114,8 @@ class _ImagePreviewState extends State<ImagePreview> {
             width: widget.width ?? MediaQuery.of(context).size.width - 24,
             fit: BoxFit.cover,
             cache: true,
-            clearMemoryCacheWhenDispose: true,
+            clearMemoryCacheWhenDispose: false,
+            cacheMaxAge: const Duration(minutes: 1),
             cacheWidth: ((MediaQuery.of(context).size.width - 24) * View.of(context).devicePixelRatio.ceil()).toInt(),
             loadStateChanged: (state) {
               if (state.extendedImageLoadState == LoadState.loading) {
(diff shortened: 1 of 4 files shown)


hjiangsu:
It looks good! I'm still a bit wary about re-introducing the `clearMemoryCacheWhenDispose` change because it had issues in the past where memory would not be cleared and it would result in a memory leak.

However, we can go ahead and profile this again and check how it performs. I'll keep this on hold until the general release so that we don't introduce this right away if thats good with you!

hjiangsu (approved):
I'll go ahead and merge this sooner so that we can get feedback from the nightly builds!

micahmo:
Reporting back after daily driving this for a while... Unfortunately setting `cacheMaxAge` seems to have an unintended side effect which is that the cache is invalidated after the given time period even if the widget is not disposed! That means, for example, if you're reading comments for a while and then navigate back to the main page, all the images reload. 😞

I would like to investigate this some more, because I would like the cache to be maintained as long as the widgets are not disposed, but also keep the cache around for a little if they are.

That said, if there isn't a nice solution by the time the next release comes around, I'd rather roll this back than have the current behavior lol.

PR #763, opened 2023-09-23
Fix the lack of a loading indicator on instances with taglines

micahmo:
## Pull Request Description

This PR fixes an issue similar to #697 where the bottom loading indicator would not be shown in the feed if the instance had taglines. Essentially we need to treat the tagline like a community header when calculating the items length and indices.

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu (approved):
LGTM! This might not be too necessary soon with the refactor of the feed page (which won't use PostCardList) but it's still good to keep it in just in case I can't finish up the refactor or if there are issues with it!

PR #765, opened 2023-09-24
Fix issue after deleting non-active account

micahmo:
## Pull Request Description

This PR fixes an issue with the new account switcher where deleting a non-active accont could cause the current active account to deactivate, and cause us to revert to an unlisted anonymous instance.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/811a18be-8825-442a-822f-7faaf696ef50

### After

https://github.com/thunder-app/thunder/assets/7417301/bf3dcf01-1a40-4aad-bafa-33700b15ca13

## Checklist

- [ ] Did you update CHANGELOG.md? -- N/A bug against new feature

### DIFF
--- a/lib/account/widgets/profile_modal_body.dart
+++ b/lib/account/widgets/profile_modal_body.dart
@@ -9,6 +9,7 @@ import 'package:swipeable_page_route/swipeable_page_route.dart';
 import 'package:thunder/account/models/account.dart';
 import 'package:thunder/account/pages/login_page.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
+import 'package:thunder/core/theme/bloc/theme_bloc.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
 import 'package:thunder/user/utils/logout_dialog.dart';
 import 'package:thunder/utils/instance.dart';
@@ -80,6 +81,11 @@ class _ProfileSelectState extends State<ProfileSelect> {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
+    final bool darkTheme = context.read<ThemeBloc>().state.useDarkTheme;
+    Color selectedColor = theme.colorScheme.primaryContainer;
+    if (!darkTheme) {
+      selectedColor = HSLColor.fromColor(theme.colorScheme.primaryContainer).withLightness(0.95).toColor();
+    }
     String? currentAccountId = context.watch<AuthBloc>().state.account?.id;
     String? currentAnonymousInstance = context.watch<ThunderBloc>().state.currentAnonymousInstance;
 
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
Looks good, just one question:

`await Future.delayed(const Duration(milliseconds: 1500)`

Is there a reason for the delay? I'm assuming its to wait for the accounts to switch/fetch information but maybe there's a better way to handle this aside from a static value? If we are keeping the delay, then it would be good to add some feedback to the user to let them know something is happening/has happened!

Also, on an unrelated note, I noticed that the selected account colour doesn't show up well when its in dark mode, but this can be for another PR!

micahmo:
> Is there a reason for the delay?

Good question! First I'll just mention that this file already had a similar delay added in #677, but I know that was such a big PR it was probably overlooked.

https://github.com/thunder-app/thunder/blob/3721092472e793d0a78c09b7c7e0e066ff76f89d/lib/account/widgets/profile_modal_body.dart#L256

But even more importantly, I'm really just copying some code that you added a while back.

https://github.com/thunder-app/thunder/blob/274f6fd1d01f443c2c14024312ba73b1d3dbed69/lib/thunder/pages/thunder_page.dart#L195-L196

While I never really looked into this, based on the comment it seems, as you alluded to, that there is some fetching/loading that we have to wait for before proceeding. Without this, calling `SwitchAccount` directly after `RemoveAccount` doesn't work. I didn't bother looking into what `RemoveAccount` might be doing that causes a delay, since you had already added a similar wait. Hopefully it's ok, and short enough not to be a bother. I can even try shortening it a bit.

> it would be good to add some feedback to the user to let them know something is happening/has happened!

Good point! In the profile switcher modal, can you think of a good place to add a spinner?

> on an unrelated note, I noticed that the selected account colour doesn't show up well when its in dark mode

I think someone mentioned this in Matrix and said that some themes look better than others. I think I probably just need to increase the opacity.

hjiangsu:
> While I never really looked into this, based on the comment it seems, as you alluded to, that there is some fetching/loading that we have to wait for before proceeding.

Yeah, I think maybe we can shorten the delay a bit more so that it feels "quicker". I think shortening the duration and adding the loading indicator should be good enough for now!

> Good point! In the profile switcher modal, can you think of a good place to add a spinner?

I think we can add a spinner in place of the trash icon when we remove an account!

> I think someone mentioned this in Matrix and said that some themes look better than others. I think I probably just need to increase the opacity.

It might just not be applying the proper colour? At least when I switch between light and dark, the selected colour is the same. I'll post a screenshot to show you what I mean!

hjiangsu:
image

image

micahmo:
> I think maybe we can shorten the delay a bit more so that it feels "quicker".

Done! Shortened as much as I could while keeping it working.

> I think we can add a spinner in place of the trash icon when we remove an account!

Done! Works for log out and delete. There's still a slight delay after the account is gone and before the previous/new one is selected, but I think it's much better now.

https://github.com/thunder-app/thunder/assets/7417301/36defddf-05de-4f42-95da-9becaff3f520

> It might just not be applying the proper colour? At least when I switch between light and dark, the selected colour is the same. I'll post a screenshot to show you what I mean!

Oh, thank you, that's not what I was expecting lol (should've just tried it myself!). Fixed.

| image |
| - |


Should be good to go now!

hjiangsu:
Nice, it looks good now! Thanks :D

PR #766, opened 2023-09-25
Remove password length limit

micahmo:
## Pull Request Description

Removes the limit for password length when logging in.

## Issue Being Fixed

Issue Number: https://programming.dev/post/3448277

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 - Handle issue where failing to retrieve image dimensions blocks post loading - contribution from @Fmstrat
 - Show additional posts loading indicator on instances with taglines - contribution from @micahmo
+- Remove password length limit - contribution from @micahmo
 
 ## 0.2.4 - 2023-09-20
 ### Added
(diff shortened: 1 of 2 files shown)


hjiangsu:
Hmm, there's some caveats when it comes to removing the password limit, especially with password managers.

When filling in the password using a password manager for a password greater than 60 characters, it throws an error. When this happens, it just throws a `password_incorrect` error which is pretty generic. This can cause confusion with users and they may think there's an issue on Thunder which prevents them from logging in.

For more context, this is the given issue that was opened a while ago for this specific issue: https://github.com/thunder-app/thunder/issues/52#issuecomment-1605622523

If we go forward with this route, we need to think of ways to handle this situation better so that this issue doesn't pop up again!

micahmo:
[I commented on Lemmy](https://programming.dev/comment/3193433), and I'm hoping we can hear back from snowe about the questions I asked. 😊 For now, I'll leave this as a draft.

micahmo:
Since we never heard back from [snowe](https://programming.dev/u/snowe) and we have a suspicion that his login would have worked fine due to the known truncation behavior, I'm going to close this PR. I've leave the branch around just in case (although it would be easy enough to recreate if needed).

PR #769, opened 2023-09-26
Fix handling of instance/network down

micahmo:
## Pull Request Description

This PR fixes an issue raised in Matrix chat which is that, if the user's network or instance is currently down, Thunder will spin forever (and also be frozen, such that the user cannot switch accounts). This is because, as of #717, we now do a downvote check during account loading (which precedes the network handling introduced in #655).

The fix is simply to wrap this initial check in a try/catch and add a timeout so that it can't kill us. Then we'll proceed with the feed loading, which is where the network handling exists. I suppose there's a tiny edge case where (a) we're connecting to an instance with downvotes disabled, (b) the downvote check fails (i.e., temporary outage), and (c) the feed loading succeeds. In this case, the downvote buttons will appear but not work. I hope this is such a narrow case and negilble impact that it will be acceptable.

I also updated some timeouts throughout the app to be more consistent.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
This can be closed when #770 is merged.

hjiangsu:
Closing as #770 is merged!

Issue #778, opened 2023-09-27
"Accessibility mode" toggle switch


shortwavesurfer2009:
Now that Thunder has an "Accessibility" menu in settings i suggest adding an "Accessibility Mode" toggle that would change all the following settings

1. General > Compact List View > set to "on"
2. General > Tappable Authors & Communities > set to "off"
3. General > Show Comment Button Actions > set to "off"
4. General > Enable Comment Navigation Buttons > set to "off"
5. Gestures > Navbar Swipe Gestures > set to "off"
6. Gestures > Navbar Double-Tap Gesture > set to "off"
7. Gestures > Post Swipe Actions > set to "off"
8. Gestures > Comment Swipe Actions > set to "off"
9. ~~Accessibility > Reduce Animations > set to "on"~~

Below the accessibility mode switch would be a text block explaining what settings will be impacted. Very similar to the text in the Floating Action Button area.

micahmo:
I definitely like the idea of making it easier for people to enable "accessibility" as a whole in the app. My concern is that some of these things are just personal preferences, and I'm curious whether we'd want to introduce a setting that makes the app perfect for only one person (as much as we love you @shortwavesurfer2009 😊). Remember that you can always import/export settings to make configuring your own perfect experience easier.

"Accessibility" is a broad term and covers a range of preferences, disabilities, etc., and I'm afraid there's not a one-size-fits-all. I am guessing that most of the things you've selected here make the app work better with a screen reader. But what if, for example, someone simply wanted motion effects to be reduced.

I wonder if we could arrive at a compromise. Maybe a few meta-settings would help (like one that that disables all swipe-related gestures (post/comment/navbar)). Or maybe there's some other way we can identify settings that could potentially be related to accessibility, so they're easier to find.

Anyway, this is just my ignorant 2₵. I would love to hear from @hjiangsu and others members of the community, especially those whom this would personally effect!

shortwavesurfer2009:
@micahmo no, i absolutely agree. I change more than these 9 but attempted to boil them down to what were not personal preferences. I tried to select only the ones that would reduce elements, gestures, etc. But i did come at this from a screen reader user POV.

micahmo:
Thanks for the thoughtful response, as always, and for categorizing the list and the motivation behind each option. Let's definitely keep moving forward together on this path to make accessibility easier for everyone!

shortwavesurfer2009:
[...]

PR #785, opened 2023-09-29
Add support for accessibility profiles

micahmo:
## Pull Request Description

This PR adds support for accessibility profiles. An accessibility profile is a list of settings and suggested values. Applying the profile will toggle the settings to the suggested values. Note that applying a profile doesn't lock you in to those settings; you may still manually set anything. Also the profile can be applied any time or any number of times. It is essentially just a "baseline".

In the UI, each profile displays a description (why one might want to use this profile) as well as a list of the settings and values.

Once the framework is in place, it will be very easy to add new profiles (you essentially just create a widget, and pass the list of settings and desired values), and we can solicit these from the community. To start, this PR intoduces one initial profile for Screen Readers with the settings suggested in #778.

> Technical note: Currently only boolean-based settings are supported as part of a profile, but I don't think it would be hard to add others. Just something to keep in mind.

@shortwavesurfer2009 Can you take a look and let us know if this is what you had in mind based on our discussion?

## Issue Being Fixed

Issue Number: #778

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/4629c8b2-1773-4c70-a4fb-16de325f2193

## Checklist


shortwavesurfer2009:
Looks good from what i can see

lib/settings/widgets/accessibility_profile.dart
```diff
@@ -0,0 +1,85 @@
+import 'package:dynamic_color/dynamic_color.dart';
+import 'package:flutter/material.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+import 'package:thunder/core/enums/local_settings.dart';
+import 'package:thunder/core/singletons/preferences.dart';
+import 'package:thunder/settings/widgets/expandable_option.dart';
+import 'package:thunder/shared/snackbar.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
+class AccessibilityProfile extends StatelessWidget {
```
hjiangsu, line 10: Thoughts on changing this from `AccessibilityProfile` to `SettingProfile` to make it more generic? It looks like this could be used across different profiles (including any potential custom user profiles if we support that in the future)

lib/settings/widgets/accessibility_profile.dart
```diff
@@ -0,0 +1,85 @@
...
+                  if (context.mounted) {
+                    showSnackbar(context, AppLocalizations.of(context)!.settingTypeNotSupported(entry.value.runtimeType));
+                  }
+                }
+              }
+              if (context.mounted && success) {
+                showSnackbar(context, AppLocalizations.of(context)!.profileAppliedSuccessfully(name));
+              }
+            },
+            child: Text(AppLocalizations.of(context)!.apply),
```
hjiangsu, line 71: Just a nice to have, but would it make sense to temporarily set the button to be disabled (and additionally change the button name from "Apply" to "Applied" once this is successful? The button can be reset when they exit and enter the page again. This gives the user a bit more feedback on their action in addition to the snackbar that shows up!

Thoughts on this?

hjiangsu (approved):
Looks good to me, just some discussion points that I wanted to bring up!

lib/settings/widgets/accessibility_profile.dart
micahmo: Sure! Assuming you're just talking about the name of this widget and not the whole feature / setting. Done!

lib/settings/widgets/accessibility_profile.dart
micahmo: Just pushed a change that does this. If it looks good to you, I think we're ready to go.

https://github.com/thunder-app/thunder/assets/7417301/829fcd27-9ce2-4ef9-bedf-77a369ae3e8c

hjiangsu: I think that works, thanks for the changes!

lib/settings/widgets/accessibility_profile.dart
```diff
@@ -0,0 +1,97 @@
...
+                          success = false;
+                          if (context.mounted) {
+                            showSnackbar(context, AppLocalizations.of(context)!.settingTypeNotSupported(entry.value.runtimeType));
+                          }
+                        }
+                      }
+                      if (context.mounted && success) {
+                        showSnackbar(context, AppLocalizations.of(context)!.profileAppliedSuccessfully(name));
+                        setState(() => recentSuccess = true);
+                        Timer(const Duration(seconds: 5), () async {
```
hjiangsu, line 77: Small suggestion here: I think it might be better to use `Future.delayed` rather than `Timer` here as `Timer` implies that this is a periodic function (even though `Future.delayed` technically uses `Timer` under the hood)!

https://api.flutter.dev/flutter/dart-async/Future/Future.delayed.html

micahmo: Done!

Issue #508, opened 2023-07-31
Integrate Lemmy Explorer for Community Search


Fmstrat:
**Is your feature request related to a problem? Please describe.**
Related: https://github.com/tgxn/lemmy-explorer/issues/137

LE is a web crawler for communities that works very well. Hopefully the above is accepted as a path forward and we can directly leverage the great work they are doing via an API. If not, and if there is interest here, I can set up a containerized API service that pulls the Redis data dump nightly to provide access to the app for Thunder users.

**Describe the solution you'd like**
A search that finds communities no matter where they are, directly built into Thunder.

**Describe alternatives you've considered**
Using LE outside of Thunder.

Fmstrat:
There is also the Data library file: https://data.lemmyverse.net/data/community.full.json

Which contains ~18M of data on all communities. Two options present themselves outside of direct integration:

1) Nightly pulls of the JSON, with an API that searches
2) If a user chooses "search all" in communities, it displays a modal letting them know it's pre-downloading community data for a faster experience, and just download the JSON locally for searching.

hjiangsu:
This is a duplicate of #14 - I'll close this to keep things clean but feel free to let me know if you think they are separate issues @Fmstrat!

tgxn:
> There is also the Data library file: https://data.lemmyverse.net/data/community.full.json
> 
> Which contains ~18M of data on all communities. Two options present themselves outside of direct integration:
> 
>     1. Nightly pulls of the JSON, with an API that searches
> 
>     2. If a user chooses "search all" in communities, it displays a modal letting them know it's pre-downloading community data for a faster experience, and just download the JSON locally for searching.

I would love for you to use the data dumps, I wanted to pre-load the processing for my frontend, so I wouldn't need to have any backend server infrastructure (other than the the crawler). That's why they are so big.
I don't really plan on hosting a dedicated search api, but I could probably whip something simple up that people could host in docker or something...

For my frontend, I split the data into chunks of ~150 items, and then load them all in parallel, with TanStack.

Undocumented, but I have the chunked data on the data page too:
`https://data.lemmyverse.net/data/community.json` will give you the count of chunks, and then the chunks are `https://data.lemmyverse.net/data/community/0.json`.

Fmstrat:
@hjiangsu Since we have the Lemmyverse Dev participating in this issue (in response to our discussion), should we reopen this and close #14 instead?

[...]

PR #799, opened 2023-10-03
Add instance pipeline

micahmo:
## Pull Request Description

This PR introduces a new pipeline which fetches a list of popular instances from [fediverse.observer](https://fediverse.observer/). The list is placed into a text file and also a code file so that it can be consumed programmatically by the Thunder app. Of course, we can also generate other kinds of files in the future (e.g., Android manifest files for handling links). The workflow then opens a PR in the repo so that we can review, approve, and merge the changes. At this time, the list of instances is generated by combining what's already in the file with the results of the API call, which queries for instances with >50 subscribers ([as Jerboa does](https://github.com/dessalines/jerboa/blob/99faf3fad3e7f21d65bc8c2b881c0748f15b55a5/app/update_instances.gradle.kts#L17-L20)). One nice thing is that it doesn't remove any instances, so we can manually add ones that aren't detected automatically, if needed.

Initially, I would like to use this list for a couple of smaller things.
* Pre-populating a list of suggested instances to log into.
* Faster checking of whether a link references a Lemmy instance (rather than our [current API check](https://github.com/thunder-app/thunder/blob/800ea093ed11bef3573f07f07bcdd663fc640b7c/lib/utils/instance.dart#L108-L125), which can be slow).

In the future, I would also like to use this list for in-app post/comment navigation, as well as the big one, handling external links.

For reference...
* [An example run of this workflow in my fork](https://github.com/micahmo/thunder/actions/runs/6388024379/job/17337202103)
* [An example of the kind of PR it will open](https://github.com/micahmo/thunder/pull/6/files)

> Note: This PR does introduce an API dependency on fediverse.observer. @hjiangsu should probably review [their terms](https://fediverse.observer/terms), and make sure we're ok with it, but everything seems fine. It's also free at this time, but of course we could remove this if that changes. For reference, Jerboa (developed by a Lemmy dev) also queries fediverse.observer in their pipeline.

> Note: I had to change a few settings related to action permissions in my repo. I can share those details if needed.

## Issue Being Fixed

Not directly related to any issue; this is more of an infrastructure thing. But should pave the way for some nice features.

## Checklist


hjiangsu:
This is pretty cool so thanks for working on this! Everything seems pretty good with me with regards to this new pipeline.

A couple of questions:
- How does this differ from something like lemmyverse?
- How should we handle the (inevitable) cases where instances are no longer active or available? Since we don't override the file, I'm assuming we would be able to do a diff between the raw pipeline output and our existing entries to remove the non-active instances (but maybe you had some other ideas on this)

Anyways, this has my approval! It would be good if you could share the changes for action permissions so that I can do that as well on here 😁

hjiangsu (approved):

micahmo:
> How does this differ from something like lemmyverse?

As far as I know, lemmyverse does not have an official API for their data. The author commented [here](https://github.com/thunder-app/thunder/issues/508#issuecomment-1667485426) and said that he doesn't "really plan on hosting a dedicated search api". They do have a data dump (https://data.lemmyverse.net/data/community.full.json), and I suppose we could download and parse that. But the fediverse API, which uses GraphQL, is very nice and allows to pass a where clause and request only specific fields, making it a very lightweight call.

> * How should we handle the (inevitable) cases where instances are no longer active or available?

That's a good question. The same could be asked for cases where we might want to manually exclude an instance for other reasons (although I doubt we want to take any kind of stance like that, as a client app). I guess I was thinking that there's no harm is keeping those domains around. If they go down, then they'd be inaccessible either way, so we're no worse off assuming that they're Lemmy instances. However, if you'd prefer, I can create an exclusion mechanism (i.e., another text file, maintained manually) that ensure select domains don't end up in the output. Let me know!

> It would be good if you could share the changes for action permissions

Of course! I had to change the following.

* Settings > Actions > General > Workflow permissions > Read and write permissions
* Settings > Actions > General > Allow GitHub Actions to create and approve pull requests

Issue #780, opened 2023-09-27
Add haptic feedback when one uses the Floating Action Button


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
The floating action button has no feedback when you use it.

**Describe the solution you'd like**
It would be great to add haptic feedback for when the floating action button activates, and you use one of the actions.

**Describe alternatives you've considered**
None.

**Additional context**
None.

micahmo:
Hey @K4LCIFER, which scenarios should have feedback, and should they all be the same intensity?

- Single-press
- Single-press to open
- Long-press
- Long-press to open
- Press close
- Swipe close
- Single-press extended action

Also this should probably wait until #752 is done. 😊

K4LCIFER:
> Hey @K4LCIFER, which scenarios should have feedback, and should they all be the same intensity?
> 
>     * Single-press
> 
>     * Single-press to open
> 
>     * Long-press
> 
>     * Long-press to open
> 
>     * Press close
> 
>     * Swipe close
> 
>     * Press extended action
> 
>     * Press comment navigation buttons
> 
> 
> Also this should probably wait until #752 is done. 😊

I'm not sure about all of those, but the one that certainly should have it is a long press. When you trigger the start of a long press it should provide feedback, then when it activates, more feedback.

PR #800, opened 2023-10-03
Fix FAB hatics

micahmo:
## Pull Request Description

This PR tweaks the haptic feedback that occurs when long-pressing the FAB. Instead of only activating when the long-press threshold has been met, there is now feedback as soon as the button is pressed _and_ when the long-press kicks in. 

I believe all other interactions with the FAB already have feedback.

## Issue Being Fixed

Issue Number: #780

## Checklist


hjiangsu (approved):
LGTM!

PR #808, opened 2023-10-05
Added support for read all inbox

ggichure:
## Issue Being Fixed

Adds  functionality to read all inbox items, by tapping the checklist icon

Issue Number:  N/A

## Screenshots / Recordings

[read_all_inbox.webm](https://github.com/thunder-app/thunder/assets/31619962/b41a914a-ec4a-4596-80e7-4952f35095b6)
## Checklist


lib/inbox/pages/inbox_page.dart
```diff
@@ -81,6 +81,15 @@ class _InboxPageState extends State<InboxPage> {
         centerTitle: false,
         title: AutoSizeText('Inbox', style: theme.textTheme.titleLarge),
         actions: [
+          IconButton(
+            icon: const Icon(
+              Icons.checklist,
+              semanticLabel: 'Read All',
```
micahmo, line 87: It looks like nothing else on this page is localized, so I wouldn't be a stickler about it, but it is nice to try to localize any new strings that we're adding. 😊

lib/inbox/bloc/inbox_bloc.dart
```diff
@@ -255,6 +259,27 @@ class InboxBloc extends Bloc<InboxEvent, InboxState> {
...
+      if (account?.jwt == null) {
+        return emit(state.copyWith(status: InboxStatus.success));
+      }
+      await lemmy.run(MarkAllAsRead(
+        auth: account!.jwt!,
+      ));
+
+      add(GetInboxEvent(reset: true, showAll: !state.showUnreadOnly));
+    } catch (e) {
+      emit(state.copyWith(status: InboxStatus.failure, errorMessage: e.toString(), totalUnreadCount: 0));
```
micahmo, line 279: Just curious about setting the unread count to 0 in case of a failure. Maybe we should just leave it as is? Not sure.

micahmo (approved):
LGTM

lib/inbox/bloc/inbox_bloc.dart
ggichure: Fixed.

lib/inbox/pages/inbox_page.dart
ggichure: Fixing this. Not clear on how you generate the localization files. I added `build_runner: ^2.4.6`

micahmo: Sorry, we need better documentation on localization!

1. Add your string to all `.arb` files.
2. Run `flutter gen-l10n`.
3. Replace the string with `AppLocalizations.of(context)!.yourResourceName`.

Let me know if that's enough detail! We should probably throw it in a README.

What does `build_runner` do?

ggichure: > What does build_runner do?


> provides general-purpose commands for generating files, including testing the generated files or serving both source and generated files. [Build runner](https://dart.dev/tools/build_runner)

E.g If you have localization files, generated models, it will gen all this in one go.

hjiangsu (approved):
LGTM!

PR #812, opened 2023-10-08
Improve URL handling in post creator

micahmo:
## Pull Request Description

This PR fixes an issue where a poorly formatted URL could either...
 - Cause the create post screen to become unusable (because the parsing would throw an exception and the same contents would always be loaded from drafts)
 - Or cause the post to be lost when attempting submission because Lemmy would reject it

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/ea72ce23-394c-4368-ac79-41980e59c145

### After

https://github.com/thunder-app/thunder/assets/7417301/1580536a-b752-420d-b1fd-2ff809528d86

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -45,6 +45,7 @@ class _CreatePostPageState extends State<CreatePostPage> {
   bool imageUploading = false;
   bool postImageUploading = false;
   String url = "";
+  String? urlError;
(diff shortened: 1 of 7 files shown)


ggichure (approved):
LGTM

hjiangsu (approved):
LGTM!

Issue #252, opened 2023-07-06
Ability to report a comment (long-press menu / swipe gesture)


krestenlaust:
**Describe the solution you'd like**
Long-pressing a comment should allow you to report it.

**Describe alternatives you've considered**
Add an option to let the user have a swipe gesture make a report (I'm assuming it'll open a modal).

**Additional context**
#189 (original issue)

PR #816, opened 2023-10-10
Ability to report comments.

ggichure:
## Pull Request Description

This PR Adds ability to report comments.

## Issue Being Fixed

Issue Number: #252

## Screenshots / Recordings
[report_comments.webm](https://github.com/thunder-app/thunder/assets/31619962/1b226c70-ed3c-41b5-95d6-67c7fb581de4)

### For Errors

Screenshot 2023-10-12 at 19 29 59

## Checklist


lib/post/widgets/report_comment_dialog.dart
```diff
@@ -0,0 +1,130 @@
...
+                      switch (state.status) {
+                        case PostStatus.success:
+                          showSnackbar(
+                            context,
+                            AppLocalizations.of(context)!.commentReported,
+                          );
+                          Navigator.of(context).pop();
+                          break;
+                        case PostStatus.failure:
+                          showSnackbar(context, state.errorMessage ?? AppLocalizations.of(context)!.unexpectedError);
```
ggichure, line 90: The `snackbar` is shown behind the bottom sheet. Trying a few things. 
Easy non intuitive way is to make it a full page.
I could also use toasts.

ggichure: see https://github.com/thunder-app/thunder/pull/816/commits/006942becb925a4b1fb08eef4503c2a7b4e469a0

lib/post/bloc/post_bloc.dart
```diff
@@ -554,6 +560,32 @@ class PostBloc extends Bloc<PostEvent, PostState> {
...
+  Future<void> _reportCommentEvent(ReportCommentEvent event, Emitter<PostState> emit) async {
+    try {
+      emit(state.copyWith(status: PostStatus.refreshing, moddingCommentId: event.commentId, selectedCommentId: state.selectedCommentId, selectedCommentPath: state.selectedCommentPath));
+
+      Account? account = await fetchActiveProfileAccount();
+      LemmyApiV3 lemmy = LemmyClient.instance.lemmyApiV3;
+
+      if (account?.jwt == null) {
+        return emit(state.copyWith(
+            status: PostStatus.failure, errorMessage: 'You are not logged in. Cannot report a comment.', selectedCommentId: state.selectedCommentId, selectedCommentPath: state.selectedCommentPath));
```
micahmo, line 572: Can you localize this? For context, you can use `GlobalContext.context`.

lib/post/bloc/post_bloc.dart
```diff
@@ -554,6 +560,32 @@ class PostBloc extends Bloc<PostEvent, PostState> {
...
+      }
+      await lemmy.run(CreateCommentReport(commentId: event.commentId, reason: event.message, auth: account!.jwt!));
+
+      return emit(
+          state.copyWith(status: PostStatus.success, comments: state.comments, moddingCommentId: -1, selectedCommentId: state.selectedCommentId, selectedCommentPath: state.selectedCommentPath));
+    } on LemmyApiException catch (e) {
```
micahmo, line 578: If you use `getExceptionErrorMessage`, you shouldn't need to catch `LemmyApiException` in particular since it will return `e.toString()` if it's not a `LemmyApiException`.

lib/post/utils/comment_action_helpers.dart
```diff
@@ -55,6 +58,11 @@ final List<ExtendedCommentCardActions> commentCardDefaultActionItems = [
     icon: Icons.share_rounded,
     label: AppLocalizations.of(GlobalContext.context)!.shareLink,
   ),
+  ExtendedCommentCardActions(
+    commentCardAction: CommentCardAction.report,
+    icon: Icons.report_outlined,
+    label: '${AppLocalizations.of(GlobalContext.context)!.report} ${AppLocalizations.of(GlobalContext.context)!.comment}',
```
micahmo, line 64: Just curious about the use of concatenating resource strings like this. It works in English, but maybe in other languages with different word order, it might not make sense to translate them separately and then combine. I'd recommend a single resource for this.

lib/post/utils/comment_action_helpers.dart
```diff
@@ -210,3 +248,22 @@ List<ExtendedCommentCardActions> _updateDefaultCommentActionItems(BuildContext c
...
+  BuildContext context, {
+  required int commentId,
+}) {
+  showModalBottomSheet(
+    context: context,
+    showDragHandle: true,
+    isScrollControlled: true,
+    builder: (_) => BlocProvider.value(
+      value: context.read<PostBloc>(),
+      child: StatefulBuilder(
```
micahmo, line 262: Any reason for using a `StatefulBuilder`?  It doesn't look like you're using `setState`.

lib/post/widgets/comment_card.dart
```diff
@@ -22,6 +22,7 @@ class CommentCard extends StatefulWidget {
   final Function(int, bool) onCollapseCommentChange;
   final Function(int, bool) onDeleteAction;
   final Function(CommentView, bool) onReplyEditAction;
+  final Function(int, bool) onReportAction;
```
micahmo, line 25: Any reason for the `bool` param? It looks like it's ignored.

lib/user/pages/user_page_success.dart
```diff
@@ -118,410 +104,433 @@ class _UserPageSuccessState extends State<UserPageSuccess> with TickerProviderSt
...
-              GestureDetector(
-                onTap: () {
-                  setState(() {
-                    _displaySidebar = !_displaySidebar;
-                  });
-                },
-                onHorizontalDragUpdate: (details) {
-                  if (details.delta.dx < -3) {
+    return BlocProvider<post_bloc.PostBloc>(
+      create: (context) => post_bloc.PostBloc(),
```
micahmo, line 108: Does this work? I know we have issues with post actions on the user page because of the different bloc types.

micahmo (commented):
Nice work and nice feature!! And thanks for the additional localizations that you added!

I left some comments for discussion.

I also had one visual suggestion. When you press report, the button disappears and progress indicator appears. The progress indicator is both taller and narrower than the button, which means the UI awkwardly shifts. What do you think about integrating it like this?

Nice work and nice feature!! And thanks for the additional localizations that you added!

I left some comments for discussion.

I also had one visual suggestion. When you press report, the button disappears and progress indicator appears. The progress indicator is both taller and narrower than the button, which means the UI awkwardly shifts. What do you think about integrating it like this?

https://github.com/thunder-app/thunder/assets/7417301/bac17d4f-0d99-4249-86ca-23776ea9e9fa

ggichure:
> Nice work and nice feature!! And thanks for the additional localizations that you added!
> 
> I left some comments for discussion.
> 
> I also had one visual suggestion. When you press report, the button disappears and a progress indicator appears. The progress indicator is both taller and narrower than the button, which means the UI awkwardly shifts. What do you think about integrating it like this?
> qemu-system-x86_64_xI0WUOq9WO.mp4

Nice. I'll work on the progress indicator.

lib/post/utils/comment_action_helpers.dart
ggichure: I do have `setState` in the [dialog](https://github.com/thunder-app/thunder/blob/fb6cfc165b89c013717e76952b5ede3bd4af384b/lib/post/widgets/report_comment_dialog.dart#L105).

lib/user/pages/user_page_success.dart
ggichure: Yes it does, We do however need to re-work where and when they are created and used.

lib/post/widgets/comment_card.dart
ggichure: Yes it is. Had some concerns  on 

1. If we should hide the reported comment.
2. What happens on reload.

From what I have seen Lemmy marks it as reported and no action in the UI.

lib/post/bloc/post_bloc.dart
ggichure: Sure

lib/post/utils/comment_action_helpers.dart
micahmo: The `ReportCommentDialog` is a `StatefulWidget`, so it has its own notion of `setState`. I'm not sure if you need the `StatefulBuilder` here. Could be wrong though. 😊

lib/post/widgets/comment_card.dart
micahmo: That's a fair question. Are you suggesting that we might want to internally track user-reported comments and not show them? What happens in the web UI?

I'm still confused about the usage of this param. The `onReportAction` just invokes the new dialog. Unlike something like upvote or save, there's no opposite action to perform; it's always just reporting, so I'm still not grasping why we need the bool. I'm sure I'm missing something though. 😊

lib/user/pages/user_page_success.dart
hjiangsu: Yeah, this should work since its essentially creating a new PostBloc instance here rather than fetching an existing PostBloc from the widget tree. I do agree with @ggichure that we need to refactor how they are created and used.

lib/post/utils/comment_action_helpers.dart
hjiangsu: I'm not too sure here either but I _think_ what @micahmo mentioned is correct since the child itself is a stateful widget?

lib/post/widgets/comment_card.dart
ggichure: Removed the bool. 
 Assuming we were hiding the comment on report , we could use this to remove the comment.

lib/post/utils/comment_action_helpers.dart
ggichure: @micahmo  You're correct. Thanks .

lib/post/bloc/post_bloc.dart
ggichure: Done

lib/post/utils/comment_action_helpers.dart
ggichure: Resolved.

hjiangsu:
Just double checking that this is ready for review?

ggichure:
> Just double checking that this is ready for review?

Yes it is.

hjiangsu (approved):
LGTM!

Issue #811, opened 2023-10-08
Various issues with feed refactor


micahmo:
I've experienced a few more issues after the feed refactor, so I'm opening a new issue to track them. @hjiangsu Let me know if this is ok!

 - [x] Navigated feeds do not have FAB.

https://github.com/thunder-app/thunder/assets/7417301/1fcd93a7-ca7b-4cf1-ab2b-d44a4bf97103

 - [x] Using the back gesture on a sidebar feed doesn't go back to the default post listing (this was introduced in [#402](https://github.com/thunder-app/thunder/pull/402), for reference)

https://github.com/thunder-app/thunder/assets/7417301/cda0cd32-6ef1-4912-8f4b-b3e28f28c435

 - [x] Feeds navigated via the sidebar do not respect the default sort and in fact override the main feed sort.

https://github.com/thunder-app/thunder/assets/7417301/449b06d8-6e2e-4710-84d9-57f93f2691f7

Let me know if you need a hand with any of these (especially the Android back gesture one).

**EDIT:** Adding one more.

 - [x] Feeds extend below the OS navigation area while comments do not.

| Header | Header |
|--------|--------|
| ![image](https://github.com/thunder-app/thunder/assets/7417301/be1b7ac3-e535-417f-8357-ccae7a7ac86f) | ![image](https://github.com/thunder-app/thunder/assets/7417301/fe8eb89d-2915-4335-834b-f6a01a9b9b6e) |

hjiangsu:
Thanks for filling out more of the issues! 

>  Navigated feeds do not have FAB.

Was this the case before the refactor? I can't remember if navigating to communities was present when tapping on a community

> Using the back gesture on a sidebar feed doesn't go back to the default post listing (this was introduced in https://github.com/thunder-app/thunder/pull/402, for reference)

Yeah, I might need a bit of help on this one - I imagine I just forgot to migrate that back handler logic during the refactor

> Feeds navigated via the sidebar do not respect the default sort and in fact override the main feed sort.

I think I know the issue on this one so I can do a quick fix for it!

> Feeds extend below the OS navigation area while comments do not.

This one is most likely just a SafeArea issue that should be fixable pretty quickly

shortwavesurfer2009:
Various buttons have lost their accessibility labels. The hamburger menu, refresh, 3 dot button next to username instance in the hamburger menu, the action buttons on a reply long press (upvote, downvote, save, reply, edit)

shortwavesurfer2009:
[...]

Issue #789, opened 2023-10-01
Unresponsiveness while loading communities


kovalensky:
**Description**  

While choosing communities from the sidebar it's not possible to re-choose community if one is already loading.
I think requests should be async.

Also sometimes there's an undebuggable bug happening. Sometimes while Thunder is trying to load the home page it stucks and keeps loading endlessly and due to previous bug I can't change the community to reinitialize a new request, then I have to close and re-open the app.

**Screenshots**  

As you can see without matter how many times I click another community it will load the first chosen one:

https://github.com/thunder-app/thunder/assets/45331093/c1d27cd8-8394-4b2b-aed4-b7f11b2e604a

**Device & App Version:**  
 - OS: Android 10, MIUI 12.5
 - Version: 0.2.4

kovalensky:
@hjiangsu Any opinions my asian bro? :D

hjiangsu:
> While choosing communities from the sidebar it's not possible to re-choose community if one is already loading.
I think requests should be async.

This seems to be due to the fact that we debounce/throttle requests on our end to limit the number of requests that gets sent to an instance in cases of spam. This means that performing multiple repeated actions in quick succession may sometimes not work because Thunder throttles it. We can play around with this value and find a more reasonable compromise, but this is what I assume is happening.

> Also sometimes there's an undebuggable bug happening. Sometimes while Thunder is trying to load the home page it stucks and keeps loading endlessly and due to previous bug I can't change the community to reinitialize a new request, then I have to close and re-open the app.

This sounds like a network issue, but I'm not too sure as there's not much info here. If you encounter it again, try to notice the conditions that it happens in (cellular service vs WiFi, network related issues, switch in connection, etc.) and open a new issue with those details if possible.

On a different note, I would also like to give a quick reminder that discussions on here should abide by our [code of conduct](https://github.com/thunder-app/thunder/blob/develop/CODE_OF_CONDUCT.md) to allow for a safe and inclusive space for everyone involved.

kovalensky:
> This sounds like a network issue, but I'm not too sure as there's not much info here.

I have a stable uplink (27mbit), this happens randomly, could be an instance slowing down others, or something in that matter, I think there should be some kind of connection repeating mechanism, since it seems Thunder is not doing ot,  I'm not able to debug it.

I'm familiar with the code of conduct, it's also important to have a good mood in conversations without having any agression towards other person, jokes could be harmless and funny ;)

kovalensky:
[...]

PR #820, opened 2023-10-11
Make feed actions more responsive (refreshing, sorting, switching communities/feeds)

hjiangsu:
## Pull Request Description

This PR aims to improve the responsiveness when refreshing, sorting, and switching communities in the drawer. I have made the corresponding feed events to be restartable, which allows them to be interrupted and restarted whenever a new event is dispatched.

As such, this essentially removes the throttling of events from before. Instead, I've added a few more checks in areas where API calls were being triggered more frequently to hopefully prevent the spam of API calls.

Since this removes the throttling of some feed events, the user can now potentially reach the instance's rate limits which might cause the instance to not accept any more API requests for a given duration. @micahmo, @ggichure feel free to test this out and see if this implementation works for both of you!

## Issue Being Fixed

Issue Number: #811, #789

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Collapsed comments are easier to expand - contribution from @micahmo
 - Show up to 99 before adding + in the unread count - contribution from @micahmo
 - Migrate from old BottomNavigationBar to NavigationBar - contribution from @ggichure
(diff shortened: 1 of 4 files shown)


micahmo (approved):
Seems reasonable to me!

ggichure:
The feed types are working as expected.

Issue #819, opened 2023-10-11
Disabled FAB still impedes scroll swipe (again!)


everdred:
**Description**  
The circumstances of this bug are almost identical to #651, which was fixed. But in short: disable the floating action button and attempt to scroll using the spot where the button used to be, and the scroll gesture is caught by the invisible button.

The differences this time:
* Attempting to scroll from that spot results in the content area becoming dark, with only bottom bar visible. Tap elsewhere to dismiss. (On #651 there was no visual feedback.)
* Appears to only affect post feed, not when viewing an individual post where the button is also disabled.

**Screenshots**  
[Screencast of bug](https://github.com/thunder-app/thunder/assets/5367819/8147d557-6270-474e-b53f-41cc885d63d9)

**Device & App Version:**  
 - OS: Android 13
 - Version v.0.2.5-1

hjiangsu:
Thanks for reporting this @everdred! This was most likely a result from the underlying changes that were made to the feed page. I'll link this with #811 since its somewhat related to that

micahmo:
For reference, original fix in #653.

hjiangsu:
As this was introduced in a nightly build but was fixed previously, I'll mark it as complete!

PR #822, opened 2023-10-11
Fix issue where disabled feed fab still contained gesture detector to summon fab

hjiangsu:
## Pull Request Description

This PR fixes the issue where the gesture detector was still present to summon the feed fab when the feed fab is disabled.  There is also a small tweak to disable hiding feed fab when we open up a community in a new page (through long-press menu actions). 

## Issue Being Fixed

Issue Number: #819

## Checklist

### DIFF
--- a/lib/feed/widgets/feed_fab.dart
+++ b/lib/feed/widgets/feed_fab.dart
@@ -51,9 +51,6 @@ class FeedFAB extends StatelessWidget {
                 semanticLabel: singlePressAction.title,
                 size: 35,
               ),
-              onSlideDown: () {
-                context.read<ThunderBloc>().add(const OnFabSummonToggle(false));
-              },
               onPressed: () {
                 HapticFeedback.lightImpact();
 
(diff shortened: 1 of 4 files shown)


lib/feed/widgets/feed_fab.dart
```diff
@@ -51,9 +51,6 @@ class FeedFAB extends StatelessWidget {
                 semanticLabel: singlePressAction.title,
                 size: 35,
               ),
-              onSlideDown: () {
```
hjiangsu, line 54: I removed this because this does not do anything. The slide down logic is located in the `GestureFab`.

micahmo (approved):
LGTM!

Out of curiosity, why do we disable hiding the FAB on navigated feeds?

hjiangsu:
> Out of curiosity, why do we disable hiding the FAB on navigated feeds?

There's an issue where if we hide the fab from the navigated feeds, the gesture detector area ends up moving from the bottom right corner to the top left corner.

Since the navigated feeds don't have the bottom navigation bar, the fab ends up being less intrusive which makes hiding the fab less necessary. I figured this is a good compromise for now. If there are requests to add back this specific feature, then I can revisit the issue with the gesture detector area!

PR #825, opened 2023-10-12
Fixes to various feed FAB issues

hjiangsu:
## Pull Request Description

This PR fixes a few issues with the feed fab. The list of issues that were found and fixed with this PR is as follows:
- [x] Fixed issue where unsupported single/long press actions showed up on the feed fab. This fix performs a check for unsupported actions, and defaults to the "open fab" action when it detects an unsupported action.
- [x] Fixed an issue where Thunder would freeze due to conflicting hero tags for fabs. This happens because there are multiple fabs used for the feed page. There is a main fab which is used for general feed/subscriptions, and there is another fab used solely for navigated community feeds. There's a situation where if you navigate to a community feed, and go back, the fab cannot perform the proper animation due to conflicting hero tags. The fix here is to pass a hero tag to the feed that is unique and non-conflicting.
- [x] Partial fix for an issue where navigating back from a community would not work as expected. 

> Partial fix for an issue where navigating back from a community would not work as expected. 

Some more context on this. There is an issue that happens with navigating back from a navigated community. To reproduce this issue:
- Navigate to a community from the drawer
- In that feed, navigate to the same community using the long-press menu
- Open up and close the fab in the navigated community
- Attempt to navigate back to the main feed

When doing this action, Thunder was freezing because there was no more pages to pop off the navigation stack. The partial fix here is to use `maybePop` rather than `pop` to prevent any freezing from occurring. This only partially fixes it because the app bar still shows a "back" icon rather than a "drawer" icon.

## Issue Being Fixed

Issue Number: #811 

## Checklist

### DIFF
--- a/lib/feed/view/feed_page.dart
+++ b/lib/feed/view/feed_page.dart
@@ -412,7 +412,7 @@ class _FeedViewState extends State<FeedView> {
                     curve: Curves.easeIn,
                     child: Container(
                       margin: const EdgeInsets.all(16),
-                      child: const FeedFAB(),
+                      child: FeedFAB(heroTag: state.communityName),
                     ),
                   ),
               ],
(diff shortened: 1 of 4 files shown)


micahmo (approved):
LGTM!

Issue #762, opened 2023-09-22
☂️ Support for lemmy 0.19.x


hjiangsu:
**Description**
This is an umbrella issue to track overall changes to support lemmy 0.19.x versions and potential features. As we get more details on the changes in 0.19.x, this issue should be updated to reflect those changes.

Related Posts:
- https://lemmy.ml/post/5328302

**Tracking**
- [x] Ability to block instances. Related issue(s): #701 
- [x] Perform checks to see if 2FA two-step process requires changes
- [x] Breaking change: add ability to authenticate API calls using headers/cookies
- [x] Ability to import and export community follows, blocklists, and profile settings
  - PR: #1638
- [x] New Scaled/Controversial post sort type and Controversial comment sort type
- [x] 0.19.4 Features: #1447

**Discussion Points**
- There needs to be a way to ensure that we continue support for 0.18.x instances without any breaking changes

micahmo:
Exciting! Here are a couple initial musings I just wanted to jot down.

* There will also certainly be some work needed in https://github.com/thunder-app/lemmy_api_client, and I wonder if we'll be able to pick up any upstream changes from https://github.com/liftoff-app/lemmy_api_client.
* We might want to add a way to check the server version, which will allow us to (a) refuse connections to versions we don't support and (b) once we support multiple version, split API calls as needed.
* If/when there is a beta build of 0.19.x, it would be nice if we could find an instance running it for testing.
* It looks like [LemmyNet/lemmy#3965](https://github.com/LemmyNet/lemmy/pull/3965) will be a very nice change for us, as the lack of child comments loading has been a source of consternation over here. 😆

SleeplessOne1917:
I saw this linked in LemmyNet/lemmy#3965. I'm still not entirely sure if I should always be returning deleted/removed comments, or just change how they get filtered on the server so that they get returned in more situations than they are currently. Given this is something that is mostly going to affect maintainers for various different lemmy frontends, the input of the contributors here would be quite useful.

micahmo:
Hey @SleeplessOne1917, thanks for checking in with us!

I believe we would prefer that you always return deleted comments. I think dessalines hits the nail on the head in [this comment](https://github.com/LemmyNet/lemmy/pull/3965#discussion_r1331583068). If the tree is manipulated in any way on the back end, that will make it much harder for us to render on the front end.

[...]

PR #829, opened 2023-10-13
Support new sort types

micahmo:
## Pull Request Description

This PR adds support for the new post and comment sort types in Lemmy 0.19.x.

It also introduces a more general way to check whether the current instance supports a given feature.

See thunder-app/lemmy_api_client#6 for the API changes.

## Issue Being Fixed

Related to #762.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/da06f046-d422-448a-b978-0cea348a383e

## Checklist


lib/core/singletons/lemmy_client.dart
```diff
@@ -7,9 +10,57 @@ class LemmyClient {
...
+      };
+    }
+
+    return false;
+  }
+
+  static final Map<String, FullSiteView> _lemmySites = <String, FullSiteView>{};
+}
+
+enum LemmyFeature {
```
hjiangsu, line 56: Just throwing an idea out there to see what you think: would it be better to make this a more advanced enum (similar to [FontScale](https://github.com/thunder-app/thunder/blob/develop/lib/core/enums/font_scale.dart)) where we define the version that is supported by the given feature

e.g.,

```dart
enum LemmyFeature {
  sortTypeControversial,
  sortTypeScaled,
  commentSortTypeControversial,
}

extension LemmyFeatureExtension on LemmyFeature {
  Version get minSupportedVersion {
    switch (this) {
      case LemmyFeature.sortTypeControversial:
        return Version(0, 19, 0, preRelease: ["rc", "1"]);
      case LemmyFeature.sortTypeScaled:
        return Version(0, 19, 0, preRelease: ["rc", "1"]);
      case LemmyFeature.commentSortTypeControversial:
        return Version(0, 19, 0, preRelease: ["rc", "1"]);
    }
  }
}
```

then, in the `supportsFeature` function, we can just grab the version of the instance, and compare it to the enum value `minSupportedVersion`

```dart
bool supportsFeature(LemmyFeature feature) {
  // previous logic
 
  return version >= feature.minSupportedVersion;
}
```

lib/settings/pages/general_settings_page.dart
```diff
@@ -405,11 +406,12 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
                         ListOption(
                           description: LocalSettings.defaultFeedSortType.label,
                           value: ListPickerItem(label: defaultSortType.value, icon: Icons.local_fire_department_rounded, payload: defaultSortType),
-                          options: allSortTypeItems,
+                          options: [...SortPicker.getDefaultSortTypeItems(includeVersionSpecificFeature: IncludeVersionSpecificFeature.never), ...topSortTypeItems],
                           icon: Icons.sort_rounded,
                           onChanged: (_) {},
                           isBottomModalScrollControlled: true,
                           customListPicker: SortPicker(
+                            includeVersionSpecificFeature: IncludeVersionSpecificFeature.never,
```
hjiangsu, line 414: Just to clarify, the default feed/comment sort options in the settings page won't allow the user to select the new sort types right?

If thats the case, then I think its okay to leave it as is for now. Eventually, we would want to allow them the option to select those (and if the lemmy instance doesn't support the selected sort, then we can default to one of available sort types)

hjiangsu (commented):
Just a few discussion points! I don't have a particular stance on those discussion points so I'm fine with merging it in as it is (with the exceptions regarding the .arb files)

lib/l10n/app_es.arb
```diff
@@ -175,6 +175,8 @@
   "shareLink": "Share Link",
   "upvote": "Upvote",
   "downvote": "Downvote",
+  "scaled": "Scaled",
```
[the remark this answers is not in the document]
micahmo: Done!

lib/l10n/app_fi.arb
```diff
@@ -175,6 +175,8 @@
   "shareLink": "Share Link",
   "upvote": "Upvote",
   "downvote": "Downvote",
+  "scaled": "Scaled",
```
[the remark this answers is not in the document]
micahmo: Done!

lib/settings/pages/general_settings_page.dart
micahmo: Yes, exactly. I figured for now, most instances won't have these, so users may be confused if they select one of the new ones by default and don't see it anywhere. Once we get a feeling that many instances are upgraded, then we can add them (with a fallback, as you said).

hjiangsu: Sounds good!

lib/core/singletons/lemmy_client.dart
micahmo: I like this idea, and I think it can be even simpler. Instead of an extension, we can have a property on the enum and set it in the constructor. I added this in ba005579a29879271021ea9aed75e313c8a906a5, let me know what you think.

micahmo:
BTW, it looks like 0.19 is broken due to rc2 on voyager.lemmy.ml. I'll look into it!

lib/core/singletons/lemmy_client.dart
hjiangsu: Just took a look at it and it looks good to me!

hjiangsu (approved):

PR #830, opened 2023-10-13
Migration of textScaleFactor to textScaler

hjiangsu:
## Pull Request Description

This PR is mainly under-the-hood changes to migrate the deprecated `textScaleFactor` to use `textScaler`. For more information about the deprecation, see here: https://docs.flutter.dev/release/breaking-changes/deprecate-textscalefactor

To accomplish this, I've created a new widget `ScalableText` which works similar to the regular `Text` widget, except it accepts a `fontScale` parameter. The `fontScale` parameter is defined [here](https://github.com/thunder-app/thunder/blob/develop/lib/core/enums/font_scale.dart). The main reason for creating this new widget is to hopefully simplify the logic for calculating the final `fontSize` attribute required.

Previously, to apply a `textScaleFactor`, all we had to do was add the parameter to any `Text` widget

``` dart
Text('example', textScaleFactor: MediaQuery.of(context).textScaleFactor * fontScale.textScaleFactor)
```

To have the equivalent functionality as before using `TextScaler`, the following changes had to be made. Note that the scaling is dependent on the `textTheme` we choose (e.g., `bodyMedium`, `titleSmall`, etc.)

```dart
Text('example', style: theme.textTheme.bodyMedium.copyWith({
    fontSize: MediaQuery.textScalerOf(context).scale(theme.textTheme.bodyMedium.fontSize * fontScale.textScaleFactor),
})
```

By using the new `ScalableText` widget, that migration code changes to the following

``` dart
ScalableText('example', fontScale: fontScale)

ScalableText(
    'example 2',
    style: theme.textTheme.titleSmall.copyWith({ color: Colors. black }),
    fontScale: fontScale
)
```

The `ScalableText` takes any existing styling that was initially applied, and appends the appropriate `fontSize` based on the `fontScale` that is passed in. This should help simplify things in the future when we want to add more resizable text elements. This approach also considers the system's font size like the previous implementation.

In addition, I've fixed up some linting warnings throughout the whole codebase to clean up warnings and errors. Since some of the code was adjusted for this migration, we should double check and see if the text scaling is roughly the same as it was before, and adjust accordingly if not.

@micahmo could you run through this and see if you notice any changes to the text size from the current `develop` build vs this change? I'll keep it as a draft until more checks are done to confirm that the general text sizes have not changed much!

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu:
One thing to note is that we need to use `flutter channel beta`. This change has not been merged upstream to the `stable` channel yet (this is why the linter is failing its checks)

lib/community/widgets/post_card_view_comfortable.dart
```diff
@@ -221,7 +220,7 @@ class PostCardViewComfortable extends StatelessWidget {
                 textContent,
                 maxLines: 4,
                 overflow: TextOverflow.ellipsis,
-                textScaleFactor: MediaQuery.of(context).textScaleFactor * state.contentFontSizeScale.textScaleFactor,
+                textScaler: TextScaler.linear(state.contentFontSizeScale.textScaleFactor),
```
micahmo, line 223: Any reason for not using `ScalableText` here?

micahmo (approved):
Very nice, thanks for taking this on! And nice cleanup too. The `ScalableText` definitely makes this transition easier.

Things look good on Android. The base size is unchanged, which is good. 😊

lib/community/widgets/post_card_view_comfortable.dart
hjiangsu: Oops, I forgot to change this instance!

hjiangsu:
I'll keep this as a draft and wait until this deprecation change reaches the upstream `stable` channel (likely with the release of Flutter 3.14 to the stable channel)

hjiangsu:
The deprecation change has made its way to the stable branch, so I'm opening this up for review

Issue #815, opened 2023-10-10
Add toggle to disable swipe to close comments


nimxaa:
**Is your feature request related to a problem? Please describe.**

I sometimes trigger it by accident and then lose the comment that I was reading.

**Describe the solution you'd like**

Option to disable it.

**Describe alternatives you've considered**

\-

**Additional context**

I'm talking about this:

https://github.com/thunder-app/thunder/assets/145933422/694392f7-c484-409b-9e49-8014d1407f32

Amunak:
Realistically it should be enough if the gesture was better tuned - expect that the user wants to scroll, and the back gesture would trigger only if you move in the X direction way more than currently.

nimxaa:
> Maybe an option here then is to have an option in the settings which disables the full screen swipe to go back feature?

That would be nice, because you can still use the back button on your phone to go back.

PR #831, opened 2023-10-13
Setting to enable/disable full screen navigation swipe gesture

hjiangsu:
## Pull Request Description

Adds a setting to enable/disable the full screen navigation swipe gesture. When this setting is disabled, you can still swipe to go back by swiping from the edge of the screen. This setting has been placed under Settings -> Gestures.

Additionally, I've added all remaining strings to be localized for the Gesture settings page. There were a few duplicate localization strings, so I've removed them here as well.

## Issue Being Fixed

Issue Number: #815

## Screenshots / Recordings

simulator_screenshot_948C3427-A5DD-4495-ACC9-04E0D7B4A33B

## Checklist


lib/l10n/app_en.arb
```diff
@@ -196,6 +194,14 @@
...
   "emptyInbox":"Empty Inbox",
   "showAll":"Show All",
+  "gestures": "Gestures",
+  "navigation": "Navigation",
+  "sidebar": "Sidebar",
+  "fullScreenNavigationSwipeDescription": "Swipe anywhere to go back when left-to-right gestures are disabled",
+  "sidebarBottomNavSwipeDescription": "Swipe bottom nav to open sidebar",
+  "sidebarBottomNavDoubleTapDescription": "Double-tap bottom nav to open sidebar",
+  "commentSwipeGesturesHint": "Looking to use buttons instead? Enable them in the comments section in general settings.",
+  "postSwipeGesturesHint": "Looking to use buttons instead? Change what buttons appear on post cards in general settings.",
```
micahmo, line 204: Do these strings need to be added to the other .arb files?

micahmo (approved):
LGTM!

lib/l10n/app_en.arb
hjiangsu: I did a test and it seems like its not needed. I'm also setting up Weblate for translations so it might be better to keep the other files empty! https://hosted.weblate.org/projects/thunder/thunder/

micahmo: Very nice!! Should we go through and remove any English strings in non-English files then?

hjiangsu: Yup, I think we can do that. We can also go ahead and remove any `.arb` files that have not yet been translated. This will just help the Weblate statuses better reflect the overall translation progress.

I've set up hooks so that Weblate automatically pushes new translations to the `translations` branch if you're interested. It seems like we already have a new language translated!

I'm holding off from making an announcement just yet until Weblate approves my application for libre (free) status on the project

PR #833, opened 2023-10-17
Fix color of FAB close button

micahmo:
## Pull Request Description

This is a super tiny color fix which updates the color of the FAB close button (in traditional or combined mode) to match the foreground of the rest of the FAB icons/text.


hjiangsu (approved):
LGTM!

Issue #639, opened 2023-08-15
Share links to Thunder to create new post


Grafcube:
**Is your feature request related to a problem? Please describe.**
Sometimes when I have a link to a lemmy instance, it would be convenient if I could open it in the app using either Android's supported URLs feature or by sharing the link.

**Describe the solution you'd like**
I would like a way to use Thunder as the default app to open lemmy/kbin links. In case the instance is not in the list of supported URLs (IIRC only a fixed list of supported URLs can be added), I could share the URL to the app and then choose between creating a new post with that URL or opening it in the app with my preferred account.

**Describe alternatives you've considered**
I don't know if any other method would work.

**Additional context**
None

realmain:
Duplicate of #60

micahmo:
I'm going to close this as a dup in favor of #60, only because that one's older and has more discussion and votes. However, @Grafcube feel free to reopen this, or add any other specific discussion points as needed!

micahmo:
I realized there's a suggestion in this issue that isn't captured elsewhere, so I'm going to rename and reopen to track this idea in particular:

> I could share the URL to the app and then choose between creating a new post with that URL

PR #836, opened 2023-10-17
Ability to create post without being in community feed

micahmo:
## Pull Request Description

This PR adds support for initiating a new post without an existing community context (i.e., from home feed) and allowing the user to select any community to post to.

> Note: This could lead into a feature where we can share any link to Thunder and create a new post from it, as suggested in #639.

> Note: Technically, this change allows you to initiate a new post within a given community and then change to a different target community. I don't think this is necessarily a bad thing though.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/35c6e313-3c00-4cc6-9d0f-49442f95d698

## Checklist


hjiangsu (approved):
LGTM!

One feature that we could add (doesn't have to be in this PR of course) would be to have a pre-generated list of suggestions based on the communities that the user is subscribed to. This way, the user can easily select one of their subscribed communities to post to rather than having to type them in!

We do keep a list of the user's subscriptions in AccountState so this could be used in this case to reduce further API calls.

micahmo:
> One feature that we could add (doesn't have to be in this PR of course) would be to have a pre-generated list of suggestions based on the communities that the user is subscribed to.

Good idea!! Done in e659566f56998c63ed4b7d148ba6cfb8ae65866b.

hjiangsu:
Looks good! I just realized I made a typo with `subscriptions` (should be subscriptions instead of subsciptions)

PR #687, opened 2023-08-25
Change instance/username entry keyboard type

micahmo:
## Pull Request Description

This is a small PR which changes the keyboard type for the instance and username input fields. On my keyboard, at least, when I press period, it thinks I'm done with a sentence so it adds a space, which is obviously very frustrating when typing a URL! By changing the input type, the keyboard is requested not to do this.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/926d2696-cf72-4677-91be-cb0276cbd714

### After

https://github.com/thunder-app/thunder/assets/7417301/79f6cd0f-44e7-42da-bad9-3b08d5d41dec

## Checklist

- [x] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable? -- N/A
- [ ] Did you add `semanticLabel`s where applicable for accessibility? -- N/A

PR #837, opened 2023-10-18
Fix the keyboard type for the search input

micahmo:
## Pull Request Description

This is a super small fix akin to #687 which fixes the keyboard type for the search field input to prevent some issues with some keyboards.

### DIFF
--- a/lib/search/pages/search_page.dart
+++ b/lib/search/pages/search_page.dart
@@ -153,6 +153,7 @@ class _SearchPageState extends State<SearchPage> with AutomaticKeepAliveClientMi
                   child: Stack(
                     children: [
                       TextField(
+                        keyboardType: TextInputType.url,
                         focusNode: searchTextFieldFocus,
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM!

hjiangsu:
Just wanted to mention something I noticed - maybe we can change the keyboard type to only be url when we're searching for Posts -> Search by URL?

Otherwise there's no way to add a space when searching for communities or posts!
IMG_1619

micahmo:
Could that be an iOS problem? I can still add a space on Android!

The main reason for this change was that my Android keyboard would add a space after a period (thinking you wanted to start a new sentence), which doesn't work well for URLs. If the iOS keyboard doesn't do that, then maybe this can be an Android/iOS difference rather than a search type difference.

https://github.com/thunder-app/thunder/assets/7417301/1a13321f-5bec-4ec3-aaef-c654b912548a

hjiangsu:
Ahh yeah, it might be platform differences. I'll make a PR to see if I can exclude this behaviour on iOS!

Issue #758, opened 2023-09-21
Remove overscroll effect when "remove animations" is checked


Jackthelad93:
The one animation that persists even when "remove animations" is checked is the overscrolling effect, which is quite jarring if you're sensitive to motion effects. The ability to disable it alongside the other animations would be welcome.

micahmo:
@Jackthelad93 Does the overscroll animation change at all? I believe in #163, @hjiangsu was able to change it from "stretching" to "bouncing" and there was some discussion about whether the latter is better. Maybe you can chime in on what you're seeing.

Jackthelad93:
@micahmo Huh. Well, I just tried putting animations back on, and there's barely any overscroll effect at all. More like that "bouncing" effect you described, which is tolerable unlike the stretching effect.

It seems like the toggle is reversed when it comes to the overscroll. So I'm assuming that it's a bug.

micahmo:
Thanks for the response! For clarify, I think we're using our terms in opposite ways. When I say "stretch" I mean that the top of the content is not moving, but rather the entirety of the page is essentially being elongated in the scroll direction. When I say "bounce" I mean that the page is being pulled past the top so that it is offset in the scroll direction. Maybe we need some videos. 😆

As for which is considered "reduced", bounce was chosen intentionally (it's not backwards). Take a look at the comments on #730 starting [here](https://github.com/thunder-app/thunder/pull/730#issuecomment-1716045209), and specifically the following.

> I think the main thing here is to remove that stretching effect that is caused when overscrolling because it causes some users to become nauseous when seeing that.
> 
> If everything is uniform and all the content moves uniformly across the screen, I think that would reduce that motion effect and hopefully reduce the nauseating feeling! My guess (but feel free to correct me here) is that reduce motion should reduce any "non-user" generated motion such as the stretch motion and navigation motion!

We had a lively debate about this and decided that the bounce was better than the stretch. However, neither of us being experts in accessibility issues. Or original goal was to remove any animation upon completion of scrolling, but that was proving difficult. In your opinion, is the stretch better than the bounce?

The original poster of #163 was unhappy with the stretch, which is what prompted our original change, but maybe we made it worse with the bounce.

hjiangsu:
Hey @Jackthelad93, thanks for creating this issue! I think maybe we need to look at this a bit further then and adjust some of the expectations on how this should work. It would be great if someone with more accessibility knowledge could chime in as @micahmo mentioned 😄

[...]

PR #839, opened 2023-10-19
Disable overscroll effect on Android when reduce animations setting is enabled

hjiangsu:
## Pull Request Description

This PR disables overscroll effects on Android whenever the reduce animation setting is on. Previously, we used to set the physics for scrollable lists to use BouncingScrollPhysics to match what happens on iOS. However, as mentioned in https://github.com/thunder-app/thunder/issues/758, that was not an optimal solution to this issue.

As such, I've figured out an alternate method of completely disabling overscroll effects by overriding the default scrollBehaviour when the app initializes.

Note that I had to add in `reduceAnimations` to `ThemeState` as `main.dart` does not have access to `ThunderBloc`. This does duplicate the code a bit since this setting is also present in `ThunderState`. To clean this up, we can remove `reduceAnimations` in `ThunderState` and only use the value present in `ThemeState` (I'll leave this as a TODO for the future PR)

@micahmo I've tested this on an emulator, but do try it out on a physical device and see if it works appropriately!

## Issue Being Fixed

Issue Number: #758 

## Screenshots / Recordings
[overscroll.webm](https://github.com/thunder-app/thunder/assets/30667958/0fa04b0e-58a8-432f-b4ee-438adb5f7b1a)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Show up to 99 before adding + in the unread count - contribution from @micahmo
 - Migrate from old BottomNavigationBar to NavigationBar - contribution from @ggichure
 - Adjusted logic to allow for instant switching of sort types, refreshing feed, and switching communities/pages in drawer
+- Removed overscroll effect on Android when reduce animation setting is enabled
 
 ### Fixed
 - Handle issue where failing to retrieve image dimensions blocks post loading - contribution from @Fmstrat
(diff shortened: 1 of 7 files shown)


micahmo (approved):
Works perfect on a physical device!!

> Note that I had to add in `reduceAnimations` to `ThemeState` as `main.dart` does not have access to `ThunderBloc`.

FWIW, @ggichure proposed moving `ThunderBloc` to `main.dart` in #832. If you go with that approach, that would solve this issue too.

hjiangsu:
> FWIW, @ggichure proposed moving ThunderBloc to main.dart in https://github.com/thunder-app/thunder/pull/832. If you go with that approach, that would solve this issue too.

That could also be a possibility!

Issue #760, opened 2023-09-22
Add an option/toggle to automatically open links in Reader View


throwawaytime:
This would be helpful as dark mode easily gets destroyed by links to bright white backgrounded pages; reader view addresses that.  As a bonus it also eliminates ads and can circumvent paywalls.

hjiangsu:
Hey @throwawaytime, thanks for the suggestion! After briefly looking at this, it seems like we may be able to get reader view on iOS devices. I'm not sure if Android has a similar implementation of this.

If you know of anything with regards to the Android side, that would be helpful! I'm not seeing this option with the package we use to handle custom tabs. Maybe @micahmo can chime in on this and see if theres a viable solution for Android

References:
- Safari allows for reader view via `enterReaderIfAvailable`: [API reference](https://pub.dev/documentation/flutter_custom_tabs/latest/flutter_custom_tabs/SafariViewControllerOption/entersReaderIfAvailable.html)

micahmo:
I'm not familiar with this as a standard Android feature. Some browsers and apps have a reader view feature, but I don't think it's something we could easily invoke. That said, I'm fine with this being an iOS-only feature until/unless we have a solution for Android.

throwawaytime:
> Hey @throwawaytime, thanks for the suggestion! After briefly looking at this, it seems like we may be able to get reader view on iOS devices. I'm not sure if Android has a similar implementation of this.
> 
> If you know of anything with regards to the Android side, that would be helpful! I'm not seeing this option with the package we use to handle custom tabs. Maybe @micahmo can chime in on this and see if theres a viable solution for Android
> 
> References:
> 
> * Safari allows for reader view via `enterReaderIfAvailable`: [API reference](https://pub.dev/documentation/flutter_custom_tabs/latest/flutter_custom_tabs/SafariViewControllerOption/entersReaderIfAvailable.html)

I am only familiar with it on iOS.  This would be a great enhancement even if it has to be iOS-only.

throwawaytime:
> I'm not familiar with this as a standard Android feature. Some browsers and apps have a reader view feature, but I don't think it's something we could easily invoke. That said, I'm fine with this being an iOS-only feature until/unless we have a solution for Android.

I am only familiar with it on iOS.  This would be a great enhancement even if it has to be iOS-only.

Issue #729, opened 2023-09-11
Ability to swipe back from built in browser 


EonAndahalf:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

micahmo:
Hi @EonAndahalf, thanks for the suggestion! Do you mean a full-screen left-to-right swipe that would dismiss the browser, like you can do on some other pages?

This is a cool idea, and something Relay supported. What I will say is that, using Custom Tabs means that the default browser takes over the whole screen and we don't have a lot of control to insert gesture handling. Another pitfall is that you may often need to swipe left-to-right on the web page you're viewing (I had to disable the gesture in Relay because it too often conflicted with actual browsing).

The last thing I'll mention is that if you're using Android 10+ with gesture navigation enabled, you can swipe from from the edge of the screen to go back. It's not the same, but at least it's a swiping gesture which will navigate back from the browser. 😊

Anyway, sorry for rambling and thanks for the suggestion!

PR #840, opened 2023-10-19
Added option to toggle reader mode for links on iOS

hjiangsu:
## Pull Request Description

This PR adds a way to toggle reader mode on iOS when tapping on an external link. This only works when `openInExternalBrowser` is false as it relies on `flutter_custom_tabs` parameter.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/760

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/30ef29dc-9818-48e6-ac46-02fe95c0ae69

## Checklist

### DIFF
--- a/lib/settings/pages/general_settings_page.dart
+++ b/lib/settings/pages/general_settings_page.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:dynamic_color/dynamic_color.dart';
 import 'package:flutter/material.dart';
 
@@ -39,6 +41,7 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
   // General Settings
   bool scrapeMissingPreviews = false;
   bool openInExternalBrowser = false;
+  bool openInReaderMode = false;
   bool useDisplayNames = true;
   bool markPostReadOnMediaView = false;
   bool showInAppUpdateNotification = false;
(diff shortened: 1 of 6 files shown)


lib/settings/pages/general_settings_page.dart
```diff
@@ -720,6 +728,14 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
                           iconDisabled: Icons.link_rounded,
                           onToggle: (bool value) => setPreferences(LocalSettings.openLinksInExternalBrowser, value),
                         ),
+                        if (Platform.isIOS)
+                          ToggleOption(
+                            description: LocalSettings.openLinksInReaderMode.label,
+                            value: openInReaderMode,
+                            iconEnabled: Icons.menu_book_rounded,
+                            iconDisabled: Icons.menu_book_rounded,
```
micahmo, line 736: Looks like there's no "off" variant of this icon. I know we had been pretty good about finding one of each for most/all of our settings so far. Oh well!

micahmo (approved):
LGTM!

BTW, I had an idea to bring back the in-app browser you developed with three options (in-app browser, custom tabs, external browser). I thought some people might like the more app-native experience better, and it might allow us to integrate some third-party reader mode on Android (if such a thing exists).

hjiangsu:
> BTW, I had an idea to bring back the in-app browser you developed with three options (in-app browser, custom tabs, external browser)

Hmm, I wonder if there are any users who prefer the in-app browser more than the custom tabs. I haven't heard anything myself but if this is the case, then we can definitely do this!

micahmo:
> I wonder if there are any users who prefer the in-app browser more than the custom tabs.

Not that they prefer it, per se, but that they want features which would required it. One example would be #729. We can't handle swipe gestures on the custom tabs, but we probably could in app. Another would be, for example, a comment button that jumps straight from a web page to the post comments (like Relay has).

hjiangsu:
Ahh I see, that makes sense!

PR #843, opened 2023-10-19
Receive share intents android

ggichure:
## Pull Request Description

This PR  adds support to handle incoming intents to create posts. 
supporting images, text and URI.

~URIs are handled as plain text and as such will be in the body of the text. ~
Ability to handle URIs and derive title from links. 

## Issue Being Fixed 

Issue Number: #639  #850

## Screenshots / Recordings

<table>
  <tr>
    <td align="center">
      Screenshot_1697987135
    </td>
    <td align="center">
      Screenshot_1697987122
    </td>
  </tr>
  <tr>
    <td align="center">
      Screenshot_1697987163
    </td>
    <td align="center">
      Screenshot_1697987170
    </td>
  </tr>
</table>

grab_link_title

## Checklist


ggichure:
@micahmo  `CreatePostPage` requires `communityId`, I reckon this should be optional given that on `onCommunitySelected` we can mutate this. Or is there something I'm missing?

micahmo:
@ggichure Thank you so much for taking this on! 

> `CreatePostPage` requires `communityId`

#836 made that optional.

https://github.com/thunder-app/thunder/pull/836/files#diff-1dfd79b4b214fd9ed9d53e1f2ac1f8ba62ee8a7efb0b81119ee00d7c478ae308R25-R26

lib/thunder/pages/thunder_page.dart
```diff
@@ -78,9 +91,83 @@ class _ThunderState extends State<Thunder> {
...
+    textIntentDataStreamSubscription = ReceiveSharingIntent.getTextStream().listen((
+      String? value,
+    ) {
+      if (value?.isNotEmpty ?? false) {
+        _navigateToCreatePostPage(text: value);
+      }
+    });
+  }
+
+  Future<void> _navigateToCreatePostPage({String? text, File? image}) async {
```
micahmo, line 140: Could this be moved to a central place (maybe `lib\feed\utils\post.dart`??) as I was thinking I could reuse it for crossposts.

macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -19,7 +19,7 @@ func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
   DeviceInfoPlusMacosPlugin.register(with: registry.registrar(forPlugin: "DeviceInfoPlusMacosPlugin"))
   DynamicColorPlugin.register(with: registry.registrar(forPlugin: "DynamicColorPlugin"))
   FileSelectorPlugin.register(with: registry.registrar(forPlugin: "FileSelectorPlugin"))
-  FLTPackageInfoPlusPlugin.register(with: registry.registrar(forPlugin: "FLTPackageInfoPlusPlugin"))
+  FPPPackageInfoPlusPlugin.register(with: registry.registrar(forPlugin: "FPPPackageInfoPlusPlugin"))
```
micahmo, line 22: Was this intentional?

micahmo (approved):
Hey, this LGTM! This is something I've been wanting for a while, so thanks for taking it on!!

I left some comments on the code.

I also tested this feature and had the following comments.

 - Is it possible to add a "subtitle" to the share intent? For example, Boost says "Create post" so it's obvious what's going to happen when you share.
   image
 - Is it possible to detect when a URL is shared and place it in the URL field instead of the body of the post? I would guess (but may be wrong) that the most common use case will be sharing news articles, which we would want to post as links.
 - This is just a general note, but it would be super cool if we could grab the article title from link and use that as suggested post title, like the Lemmy web UI does. Something for the future. 😊

BTW, I have a coding style note which is totally just a personal preference but I'd like to mention and maybe get some feedback. When checking the value of a nullable bool, I prefer to use an equality test against the value I want, rather than a fallback to the value I don't. For example, I find...
``` dart
bool? myVal;
if (myVal == true) {}
```

much more clear than 

``` dart
bool? myVal;
if (myVal ?? false) {}
```

However again, that's just a preference on my part. 😊

macos/Flutter/GeneratedPluginRegistrant.swift
ggichure: Did a `flutter pub upgrade`  and it generated this.

lib/thunder/pages/thunder_page.dart
ggichure: Done.

lib/community/pages/create_post_page.dart
```diff
@@ -84,15 +95,29 @@ class _CreatePostPageState extends State<CreatePostPage> {
...
+          uploadImage(
+            context,
+            imageBloc,
+            postImage: true,
+            imagePath: widget.image?.path,
+          );
+        });
+      }
+    } else {
+      if (widget.previousDraftPost != null) {
```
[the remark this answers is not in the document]
ggichure: Done

ggichure:
> This is just a general note, but it would be super cool if we could grab the article title from link and use that as suggested post title, like the Lemmy web UI does. Something for the future.

This would be awesome. I'll add it as a feature request.

ggichure:
> Is it possible to add a "subtitle" to the share intent? For example, Boost says "Create post" so it's obvious what's going to happen when you share.

Done 
Screenshot 2023-10-23 at 10 39 44

lib/community/pages/create_post_page.dart
```diff
@@ -27,12 +29,21 @@ class CreatePostPage extends StatefulWidget {
...
+
   const CreatePostPage({
     super.key,
     required this.communityId,
     this.communityView,
     this.previousDraftPost,
     this.onUpdateDraft,
+    this.image,
+    this.text,
+    this.creatingFromIntent = false,
```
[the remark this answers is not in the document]
ggichure: Done

ggichure:
> Is it possible to detect when a URL is shared and place it in the URL field instead of the body of the post? I would guess (but may be wrong) that the most common use case will be sharing news articles, which we would want to post as links.

Done. 

```dart
final uri = Uri.tryParse(initialText!);
      if (uri?.isAbsolute == true) {
```

ggichure:
> > This is just a general note, but it would be super cool if we could grab the article title from link and use that as suggested post title, like the Lemmy web UI does. Something for the future.
> 
> This would be awesome. I'll add it as a feature request.

This has been done. Although currently only works for urls coming from intents. 
I'll need some guidance on handling this for manually added links.

android/app/src/main/kotlin/com/example/thunder/MainActivity.kt
```diff
@@ -1,6 +1,23 @@
...
+import android.os.Bundle
 
 class MainActivity: FlutterActivity() {
-}
+    override fun configureFlutterEngine(@NonNull flutterEngine: FlutterEngine) {
+        GeneratedPluginRegistrant.registerWith(flutterEngine);
+        java.lang.Thread.sleep(1000);
+    }
+    override fun onCreate(savedInstanceState: Bundle?) {
+        if (intent.getIntExtra("org.chromium.chrome.extra.TASK_ID", -1) == this.taskId) {
```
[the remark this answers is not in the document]
ggichure: Done

micahmo (approved):
Thank you so much for responding to all the feedback! I tested again and it works great. I'd say it's good to go! 😊

hjiangsu (approved):
LGTM! Thanks again for doing this. I think this package also supports iOS so I can work on getting the iOS to handle this properly

micahmo:
It looks like all link handling (i.e., the "open by default" feature) is now triggering the create post page. 😞 I'll see if I can look into this but might need your help @ggichure.

ggichure:
> It looks like all link handling (i.e., the "open by default" feature) is now triggering the create post page. 😞 I'll see if I can look into this but might need your help @ggichure.
> 
> **EDIT**: Looks like it's because `navigateToPost` is looking for `InstanceBloc` which isn't defined when triggering from share. Might need your help too @hjiangsu. 😆
> qemu-system-x86_64_hxPuk0utVy.mp4

Interesting. I'll aslo look into it.

Issue #273, opened 2023-07-07
More search options


trymeouteh:
Please add more search options

### Type
 - [ ] All
 - [x] Comment
 - [x] Post
 - [x] Communities
 - [x] Users
 - [x] URL

### Filter
 - [x] Subscribed
 - [x] Local
 - [x] All

### Sort
 - [x] New
 - [x] Old
 - [x] Top Hour
 - [x] Top Day
 - [x] Top Week
 - [x] Top Month
 - [x] Top Year
 - [x] Top All Time

And add a search bar within each community to fold posts and comments within that specific community.

micahmo:
Just an FYI, I edited the original post to have checkmarks for each item we need to implement. I'm checking them when the related code gets merged into this repo (not necessarily when the feature is release).

trymeouteh:
Your unable to filter a user search to local instance and global. And instead of the filter being called All Posts and Local Posts it should be called All Users and Local Users

micahmo:
> Your unable to filter a user search to local instance and global.

I agree. However, the web UI still shows those options, although they don't seem to have any affect, so I was just following that pattern. Do you think we should hide that filter in Thunder when `Users` is selected?

| ![image](https://github.com/thunder-app/thunder/assets/7417301/9285510a-7fb9-4740-90c6-9b4a5cfa6ee3) |

> instead of the filter being called All Posts and Local Posts it should be called All Users and Local Users

How about just "Subscribed", "Local", "All", like the web?

micahmo:
Hey @trymeouteh (and anyone else that's interested), do you think this issue is fully resolved as of [0.2.7-1](https://github.com/thunder-app/thunder/releases/tag/0.2.7-1)? I guess the only thing we don't technically have is the "All" search type, but every searchable type is available (we just don't have a UI to combine results like that). What do we think?

hjiangsu:
I would consider this to be fully resolved! I'll mark it as such and if there are any further requests, a new issue can be made

PR #845, opened 2023-10-21
Add more search options

micahmo:
## Pull Request Description

This PR enhances the search page.

New chips have been added under the text box. (See Gmail for the UI inspiration here.)
1. The search type chip will let you pick what type of objects to search (currently only Communities and Users are supported, but now that this framework is place, others should be easy to add).
2. The sort option has been taken off the text box and placed into a chip here as well.
3. A feed type selector has also been added as the third chip.

In the future, filters by Community and Creator (like the web UI has) can be added as additional chips.

## Issue Being Fixed

Issue Number: Related to #273

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/daf4876a-bc58-4d6d-8259-9f6e9909a5a7

## Checklist


micahmo:
Updated this to follow the new refactor. Should be good to go!

CTalvio:
Damn buddy, you're _still_ on fire! I'm catching up on what ya'll been doing.

I think I'd prefer the search options appear in a grid instead of being horizontally scrollable, and then having that grid be collapsed by default behind an "advanced search" button. I think there was some talk of this in matrix, as something like this will be necessary eventually due to the large number of possible options.

Can be another commit though, when I get back into swing I'll likely make a lot commits that are simple cosmetic/UX polish.

hjiangsu (approved):
Looks good to me. Something to maybe consider in the future is splitting up the logic into smaller widgets (e.g., splitting up the search sort bar) but it's not necessary!

PR #846, opened 2023-10-21
Do not throttle mark as read

micahmo:
## Pull Request Description

I was having some issues marking a bunch of messages as read. Every once in a while they would just spin forever. Turns out this particular call was throttled. I think this should be safe to do because it's always a user-generated action (as opposed to the issues we were having in the past where the app itself was generating too many requests).


hjiangsu (approved):
LGTM!

PR #847, opened 2023-10-21
Update instance workflow to mention manifest

micahmo:
## Pull Request Description

This is just a tiny fix to the instance workflow which mentions the new file that it is now updating.


hjiangsu (approved):
LGTM!

PR #849, opened 2023-10-22
Adjust prompt queries

micahmo:
## Pull Request Description

This is a super small change which adjusts the parameters used when performing queries to build autocomplete suggestions when prompting for user input related to finding users or communities. By increasing the limit and tweaking the sort type, we're more likely to get good suggestions. (Of course, the user can always type the exact name of a community for an exact match.)


hjiangsu (approved):
LGTM!

PR #854, opened 2023-10-25
Refactor to address breaking Lemmy API changes

hjiangsu:
## Pull Request Description

This is a draft PR which fixes the breaking changes from the lemmy api as described here: https://github.com/thunder-app/lemmy_api_client/pull/10

This is in draft stages right now as more testing is needed to ensure there's no breaking functionality still present. To briefly summarize the changes:
- VoteType is no longer present in lemmy API and has been replaced by an int indicating the vote (-1, 0, 1)
- Adjusted the classes/models to correspond with the new classes/models from the lemmy API refactor

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -34,6 +34,8 @@ PODS:
   - sqflite (0.0.3):
     - Flutter
     - FMDB (>= 2.7.5)
+  - uni_links (0.0.1):
+    - Flutter
   - url_launcher_ios (0.0.1):
     - Flutter
   - webview_flutter_wkwebview (0.0.1):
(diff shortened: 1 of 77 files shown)


hjiangsu:
@micahmo I ran this and it seems like most things are functional at this stage (I just did a very brief look over things), so feel free to run this build and let me know if you spot anything!

Similar to the feed refactor, it'll be impractical to check for every single thing, so it might make some time for us to spot all the issues that were introduced as a result of this PR

micahmo:
From a quick test, things look good to me too! Agreed that the most practical approach is to get it in and start using it.

micahmo:
You bet!

micahmo:
@hjiangsu I just noticed one small "regression" after running this branch (maybe it was an intentional change?). The user profile area in the drawer no longer shows the instance name.

| Before | After |
|--------|--------|
| image | image |

hjiangsu:
Good catch! I think I may have commented  that section out because the field no longer existed after the refactor. I'll take a look at it and find an alternative if possible!

micahmo:
Thanks! We still display the instance name in the account switcher, so I'm sure there's a way. 😊

hjiangsu:
@micahmo I've added that logic back in #857!

micahmo:
Awesome, let me know when that one's ready to test/review!

Issue #850, opened 2023-10-23
Grab the article title from link and use that as suggested post title


ggichure:
**Is your feature request related to a problem? Please describe.**

Currently when we are creating new posts from share intents, one has to type the title.

**Describe the solution you'd like**

Grab the article title from link and use that as suggested post title.

**Describe alternatives you've considered**

**Additional context**

micahmo:
This is done for URLs shared through intents. Still needs to be done for links typed/pasted into the new post screen.

PR #855, opened 2023-10-26
Support suggested post titles

micahmo:
## Pull Request Description

Following up on #843, this PR adds a feature of the web UI where, after inserting a link into the post, you may auto-populate the title field with a suggested title which is scraped from the URL.

## Issue Being Fixed

Issue Number: #850

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/fe786f7f-8d55-4a3e-9778-90d01955c698

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:flutter_typeahead/flutter_typeahead.dart';
 
 import 'package:lemmy_api_client/v3.dart';
 import 'package:link_preview_generator/link_preview_generator.dart';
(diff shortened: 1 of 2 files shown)


micahmo:
This should still be good to go after the refactor!

ggichure (approved):
LGTM

lib/community/pages/create_post_page.dart
```diff
@@ -136,15 +137,25 @@ class _CreatePostPageState extends State<CreatePostPage> {
...
-  Future<void> _getDataFromLink() async {
-    if (widget.url?.isNotEmpty == true) {
-      final WebInfo info = await LinkPreview.scrapeFromURL(widget.url!);
-      _titleTextController.text = info.title;
+  Future<String?> _getDataFromLink({String? link, bool updateTitleField = true}) async {
+    link ??= widget.url;
+    if (link?.isNotEmpty == true) {
+      try {
+        final WebInfo info = await LinkPreview.scrapeFromURL(link!);
+        if (updateTitleField) {
```
hjiangsu, line 145: Is this parameter being used anywhere? Is it necessary to have it here if it's not being used?

hjiangsu (commented):
Looks good, just one comment!

lib/community/pages/create_post_page.dart
micahmo: Yes. This method was originally called to set the title field. I repurposed it to also be able to return the title, so I wanted to make setting the field optional.

Here's where it's originally called. `updateTitleField` is `true` by default, so no need to pass it.

https://github.com/thunder-app/thunder/blob/774a22a54635d31c21c8669c9102d06398d96f1f/lib/community/pages/create_post_page.dart#L105

And here's where I call it to get the value. I pass `updateTitleField: false` because I don't want it to change anything.

https://github.com/thunder-app/thunder/blob/774a22a54635d31c21c8669c9102d06398d96f1f/lib/community/pages/create_post_page.dart#L273

Hope that makes sense!

hjiangsu: I see, I missed that. Thanks for the clarification!

Issue #701, opened 2023-09-02
Instance Filtering


1Phase1:
**Is your feature request related to a problem? Please describe.**
The ability to filter entire instances and all users and posts associated with those instances would be a very useful feature for dealing with certain problem instances. Like if I don't want to see LemmyNSFW posts on my c/All feed without needing to block every individual LemmyNSFW community or make every NSFW post invisible (since not all NSFW is automatically porn), I could just filter the entire instance.

**Describe the solution you'd like**
An instance filtering feature, like the ones featured in other Lemmy apps such as Sync and Connect, where you can open the actions menu on a post, then press the "filter instance" option to filter the instance, as well as the ability to go into the settings and view a list of filtered instances from there, and filter an instance directly from there by inputting the instance domain name.

**Describe alternatives you've considered**
As of now, there is no alternative for this feature, short of using a different Lemmy app (I'd prefer to stay on this one) or waiting for the feature to be implemented site-wide, and presently there's no indication of a time frame on when that will happen.

micahmo:
Thanks for the request @1Phase1! We've been tracking the progress on this over at Lemmy (https://github.com/LemmyNet/lemmy/pull/3869) and it looks like they're on the way to implementing it in the backend. Therefore, I believe we'd wait to implement this in Thunder until it's officially part of the API, otherwise we'd have to do it all over again when it is.

1Phase1:
@micahmo Ah, perfect. I was unaware there was a dedicated pull request for this feature. In that case, I'm satisfied with waiting for site-wide implementation. Thanks a lot for the answer. Thanks also for linking the pull request. I can now directly follow the PR's progress.

PR #857, opened 2023-10-26
Initial ability to block instances

hjiangsu:
## Pull Request Description

This PR introduces the ability to block instances when using an instance on 0.19.x (e.g., voyager.lemmy.ml). ~This is currently a WIP and is considered to be a draft at this time.~ This PR currently adds the following:

> Review with whitespace off

- Ability to block instances via long-press menu on posts in the feed page (hidden for unsupported instances < 0.19.x)
- Added instance blocking management to user settings page (hidden for unsupported instances < 0.19.x)
  - Additionally, I've refactored the code here to make things a bit more modular
- Added back instance to feed drawer that was lost during #854
- Added ability to block instances from the user settings page (similar to community/user), including a way to autocomplete instances
- Minor changes to user/community avatars to handle empty urls more appropriately

## Issue Being Fixed

Issue Number: #762, #701

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/7a16a004-43d8-4108-8347-dc912ae3993b

## Checklist


hjiangsu:
@micahmo I think this should be ready now!

lib/instance/bloc/instance_bloc.dart
```diff
@@ -0,0 +1,66 @@
...
+    emit(state.copyWith(status: InstanceStatus.success, message: null));
+  }
+
+  /// Handles instance related actions
+  Future<void> _onInstanceAction(InstanceActionEvent event, Emitter<InstanceState> emit) async {
+    emit(state.copyWith(status: InstanceStatus.fetching));
+
+    final l10n = AppLocalizations.of(GlobalContext.context)!;
+
+    switch (event.instanceAction) {
```
micahmo, line 52: I like this pattern of having one event and an enum of possible actions! Seems like it could simplify things.

lib/shared/community_icon.dart
```diff
@@ -26,8 +26,10 @@ class CommunityIcon extends StatelessWidget {
               )
             : null);
 
+    if (community?.icon == null || community!.icon!.isEmpty) return placeholderIcon;
```
micahmo, line 29: Could this be simplified as 

```
if (community?.icon?.isNotEmpty != true)
```
?

lib/shared/input_dialogs.dart
```diff
@@ -187,62 +259,65 @@ void showInputDialog<T>({
...
-                      error = null;
-                    }),
-                    autofocus: true,
-                    decoration: InputDecoration(
-                      isDense: true,
-                      border: const OutlineInputBorder(),
-                      labelText: inputLabel,
-                      errorText: error,
+            content: SizedBox(
+              width: min(MediaQuery.of(context).size.width, 700),
```
micahmo, line 263: Thank you so much for this change. The narrow width has been bothering me. 😊

lib/shared/user_avatar.dart
```diff
@@ -24,6 +24,8 @@ class UserAvatar extends StatelessWidget {
           ),
         ));
 
+    if (person?.avatar == null || person!.avatar!.isEmpty) return placeholderIcon;
```
micahmo, line 27: Same comment about simplifying this check.

lib/shared/input_dialogs.dart
```diff
@@ -167,6 +169,76 @@ Widget buildCommunitySuggestionWidget(payload, {void Function(CommunityView)? on
...
 }
 
+/// Shows a dialog which allows typing/search for an instance
+void showInstanceInputDialog(BuildContext context, {required String title, required void Function(Instance) onInstanceSelected, Iterable<Instance>? emptySuggestions}) async {
+  Future<String?> onSubmitted({Instance? payload, String? value}) async {
+    if (payload == null) return null;
+
+    onInstanceSelected(payload);
+    Navigator.of(context).pop();
+    return null;
```
micahmo, line 179: See my overall comments about this.

micahmo (approved):
The code looks great!! Nice work!

I also tested the feature a bit. I found one tiny thing that didn't work the way I expected. It looks like the input dialog does not support free text entry (only selection from the auto-complete list). If I type the name of an instance (whether valid or not) and press OK or enter on the keyboard, the OK button just becomes disabled with no indication of why.

I think you can check `value` (when `payload` is null) and see whether it corresponds to a real instance. If so, you can pass it to `onInstanceSelected`. If it's not valid, you can return an error message as the result of `onSubmitted`. That should only return `null` for a good instance. Take a look at `onSubmitted` in `showCommunityInputDialog` as an example. Also see this demo for an example of the different behavior.

https://github.com/thunder-app/thunder/assets/7417301/17c1f0df-7073-4b8f-a684-a21f2679ed69

The only other note I had is that we might want to keep both community actions and both instance actions next to each other in the post actions menu.

image

Otherwise, everything seems great!

Here are some future ideas:
 - Ability to block from instance page (with a button similar to the community/user sidebar)
 - Ability to block from comments (to be honest, comments are the thing that most often triggers a desire to block an instance 😆).

hjiangsu:
> I found one tiny thing that didn't work the way I expected. It looks like the input dialog does not support free text entry (only selection from the auto-complete list)

Ahh good point, I'll add that in!

> The only other note I had is that we might want to keep both community actions and both instance actions next to each other in the post actions menu.

Good idea, I'll also make these changes!

lib/instance/bloc/instance_bloc.dart
hjiangsu: This pattern was also used in the feed refactor so I just used the same pattern here!

lib/shared/community_icon.dart
hjiangsu: Addressed in 819d861d04d66f2f1699d94c1a905cddeb931c38

lib/shared/user_avatar.dart
hjiangsu: Addressed in 819d861d04d66f2f1699d94c1a905cddeb931c38

lib/shared/input_dialogs.dart
hjiangsu: Addressed in 88b50b7ea453b38fc548ed2fab1f440ee5c8417a

hjiangsu:
Should be good to go now! @micahmo

micahmo (approved):
Looks amazing!!

PR #861, opened 2023-10-29
Add comments as supported search type

micahmo:
## Pull Request Description

This PR adds support for searching for comments.

## Issue Being Fixed

Related to #273.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/06284a89-bc90-4a84-9906-3f499ceaa765

## Checklist


lib/search/bloc/search_bloc.dart
```diff
@@ -46,6 +52,14 @@ class SearchBloc extends Bloc<SearchEvent, SearchState> {
       _getTrendingCommunitiesEvent,
       transformer: throttleDroppable(throttleDuration),
     );
+    on<VoteCommentEvent>(
```
hjiangsu, line 55: Eventually, we might want to combine these general events into one single event (similar to how FeedBloc and InstanceBloc handles it) to make it a bit cleaner to work with!

hjiangsu (approved):
LGTM - Just had one comment but that doesn't need to be done at this point in time!

lib/search/bloc/search_bloc.dart
micahmo: Absolutely agreed, there's a lot of duplication around comments. Might make sense to have a `CommentBloc` at some point.

Issue #768, opened 2023-09-25
"Could not determine post to comment to." when replying from inbox


danielton1:
**Description**  
I get an error message saying it can't find the post to comment to when replying from my inbox

**How to Reproduce**  
1. Go to inbox
2. Select reply on a comment, either through the slide gesture or reply button
3. Compose comment and hit Send

**Expected Behavior**  
Reply should be posted to comment

**Screenshots**  

**Device & App Version:**  
 - Device: iPhone 13
 - OS: iOS 17.0.1
 - Version 0.2.4

**Additional Context**  
This problem exists on both the current app store and testflight builds.

micahmo:
Thanks for the report!

For the devs, should be the same underlying issue as #738.

PR #862, opened 2023-10-30
Fix issue with comment reply not working within Inbox page

hjiangsu:
## Pull Request Description

This PR fixes an issue where attempting to reply from the inbox to a comment does not work as expected. This only fixes the comment issue, it does not address not being able to vote from the reply. Note that this is likely a temporary fix until a better and more proper solution is implemented (refactoring of code to be less dependent on specific blocs)

## Issue Being Fixed

Issue Number: #768

## Checklist


micahmo (approved):
LGTM!

Yeah I've had to do some of these funny bloc checks too, which, like you said, is probably indicative that things need a refactor. But good enough for now. 😊

PR #866, opened 2023-10-30
Fix issues opening links in Thunder

micahmo:
## Pull Request Description

This PR fixes an issue ([described here](https://github.com/thunder-app/thunder/pull/843#issuecomment-1785186734)) where the "open by default" feature would trigger the create post page (as though the link was "shared" to the app) rather than navigating to the target of the link.

1. Updated `navigateToPost` to not rely on `InstanceBloc`. (@hjiangsu I'm not sure if this is the right fix, or if we actually need to make sure that the `InstanceBloc` exists.)
2. The second issue was more much ugly. Both `receive_sharing_intent` and `uni_links` were handling the `VIEW` intent, and I couldn't find a way to make the former not do that. My solution was to invoke a message from the native Kotlin code to our Dart code to set the current intent. Then in the handling, we can check which one we have. Specifically, we don't want the sharing code to handle the viewing intent. Again, not sure if this is the best approach, but it seems to work.

P.S. I'm not sure if this will have any impact on iOS. Since it doesn't have the ability to handle links or shares, I hope it won't matter now, but it might eventually.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/1c40618e-3760-422d-86b6-a31f2be99239

## Checklist


hjiangsu:
> Updated navigateToPost to not rely on InstanceBloc. (@hjiangsu I'm not sure if this is the right fix, or if we actually need to make sure that the InstanceBloc exists.)

I was just about to comment on the other issue about the `InstanceBloc`! I think we may need to create the InstanceBloc if it doesn't exist. With this current fix, are you able to navigate to a user/community page from within the post page? (e.g., Open the link in Thunder which navigates to the post page, and then from the post page, tap on the user/community on the post)

I had to pass in the InstanceBloc in order for that functionality to work yesterday (https://github.com/thunder-app/thunder/commit/9474c2a78e1a385f8f7a5ad9577efab5f71255a1) but didn't realize it affected opening links on Android 😅

micahmo:
> With this current fix, are you able to navigate to a user/community page from within the post page?

Ahh good catch, no that doesn't work. Although if I create a new `InstanceBloc`, then it still doesn't work due to needing `CommunityBloc`, so I guess this isn't a new problem.

micahmo:
Moving `AnonymousSubscriptionsBloc`, `CommunityBloc`, and `InstanceBloc` up to `main.dart` seems to make things happy. Again, not sure if that's the best approach. But it kinda makes sense, because we can't wait for `ThunderPage` while handling the intents.

hjiangsu:
I think everything looks good! I don't know too much about Kotlin in general so I can't comment on that (maybe @ggichure can pitch in any feedback on it)

I'll merge this in for now since it looks good on my end, but if there are any further changes that are needed, feel free to open up a new PR for it!

hjiangsu (approved):

Issue #304, opened 2023-07-09
support spoiler tags


phareous:
**Is your feature request related to a problem? Please describe.**

support spoiler tags in comments in the form of

::: spoiler label
Hidden text
:::

**Describe the solution you'd like**

Text hidden until you click on it

**Describe alternatives you've considered**

**Additional context**
![IMG_3549](https://github.com/hjiangsu/thunder/assets/1590943/55804e17-5e86-4228-8e98-6b07f9b75704)

gon555551:
I really hope this gets implemented!!!

hjiangsu:
For this, we would need to add in some custom logic to the Markdown parser to be able to detect spoiler tags as I think this is not something native to Markdown

Another point with this is that we should then support adding spoiler tags when creating or editing posts/comments

CTalvio:
I did some experimenting with this.

Markdown does "support" collapsible content, like this:
```
<details>
  <summary>Spoiler warning</summary>
  
  Spoiler text. Note that it's important to have a space after the summary tag. You should be able to write any markdown you want inside the `<details>` tag... just make sure you close `<details>` afterward.
  
  ```javascript
  console.log("I'm a code block!");
  ```
  
</details>
```

But this is really working using html. This example works when testing with this: https://dart-lang.github.io/markdown/

But for obvious reasons, doesn't when using a flutter markdown library.

And Lemmy has it's own syntax for spoiler with the triple colon.

micahmo:
I believe at some point we may need to fork `flutter_markdown` (which unfortunately is in a repo with all of the other first-party packages), since there are probably some other things we'd like to tweak as well (e.g., detecting `!` as part of a hyperlink for communities).

https://github.com/flutter/packages/tree/main/packages/flutter_markdown

hjiangsu:
We might not need to fork the markdown package. I think you can extend the package to add in custom syntax logic: https://pub.dev/documentation/markdown/latest/markdown/ExtensionSet-class.html

micahmo:
I started taking a look at this, and I think we may need to fork `flutter_markdown` after all.

For some background, I was able to add a custom `BlockSyntax` for parsing markdown spoilers. You can see it here.

https://gist.github.com/micahmo/9ed15cb884b79850a787816e1f9af497

Using just the `markdown` library, this syntax extension works great to generate HTML.

**Input**
``` dart
void main() {
  print(
    markdownToHtml(
      """### This is a title

This is a body
[...]

PR #867, opened 2023-10-30
Initial support for markdown spoiler tags

hjiangsu:
## Pull Request Description

This PR introduces some basic functionality for handling spoiler tags within Markdown. This works by creating a new `ExtensionSet` which parses spoiler tags, and generates a custom widget for tags defined as `spoiler` through the use of `MarkdownElementBuilder`

This implementation doesn't fully follow Lemmy's method of showing spoilers in that it doesn't show the title of the spoiler. However, most other things work as expected now. When there is a spoiler, it will hide it using a spoiler block. You can tap on the spoiler block to show or hide the spoiler.

Note: currently, the spoiler will take up roughly the same amount of space as the actual content of the spoiler. I don't know what the best way to do this is, so if there are any suggestions, that would be great!

When using Lemmy's spoiler tag definition, the spoiler will always be on a new line. This is because Lemmy's spoiler tag definition looks like:

```
::: spoiler title
body
:::
```

I've also added unofficial support for doing inline spoiler tags as defined below. Note that this is not an officially supported specification/syntax so it does not work across Lemmy apps. If there ever comes a point where there's a way to define an inline spoiler text, then we can modify the regex to use that spec.

```
::: spoiler text_of_spoiler :::
```

## Issue Being Fixed

Issue Number: #304

## Screenshots / Recordings

|  Before Tapping  |  After Tapping |
|---|---|
| simulator_screenshot_89403F89-D59F-4E37-BC74-37F942EAA5CD | simulator_screenshot_2DBA15BA-AF1D-445A-93BC-8F9448607FCD |
| simulator_screenshot_355FBBA0-165F-405C-9EEB-5A6E07ECCC26 | simulator_screenshot_BE82D676-73A0-443D-8056-ECE46375AB34|

## Checklist


lib/shared/common_markdown_body.dart
```diff
@@ -152,3 +162,122 @@ class CommonMarkdownBody extends StatelessWidget {
...
+      if (_spoilerBlockEnd.hasMatch(parser.current.content)) {
+        parser.advance();
+        break;
+      }
+      body.add(parser.current.content);
+      parser.advance();
+    }
+
+    // Create a custom Node which will be used to render the spoiler in [SpoilerElementBuilder]
+    final md.Node spoiler = md.Element('p', [
```
hjiangsu, line 210: @micahmo I encountered the same issue when I tried to add a custom tag in here (as you mentioned in https://github.com/thunder-app/thunder/issues/304#issuecomment-1660486274

The workaround here is to use a known element (e.g., `p` in this case), and then supply the custom element as the child of this (e.g., `spoiler`)

micahmo: Wow great find! I'm so glad you tried it out!

micahmo:
Hey I thought I had already commented on this PR, but now I can't find my comment. Fortunately I saved it. Here's what I meant to say.

---

Wow, nice work!! This already looks great!

Unfortunately, either I'm doing something wrong, or it doesn't work quite right on Android. Here's what I see on the post you linked. Some comments are empty, others don't respect the spoiler, etc. Can you try in your Android emulator just to see if I'm crazy?

| image |
| - |

hjiangsu:
Hmm, that's really interesting. I'll do a test with an Android emulator when I get the chance and see what the issue here is!

Vilianexe:
> Hmm, that's really interesting. I'll do a test with an Android emulator when I get the chance and see what the issue here is!

i have an android phone(samsung galaxy m21s), if you want me test it on hardware i'm happy to help :b

hjiangsu:
> Unfortunately, either I'm doing something wrong, or it doesn't work quite right on Android. Here's what I see on the post you linked. Some comments are empty, others don't respect the spoiler, etc. 

@micahmo I think I figured out the issue - could you try again and see if it works now?

> i have an android phone(samsung galaxy m21s), if you want me test it on hardware i'm happy to help :b

@Vilianexe Thanks for the interest! Do you happen to have some prior background in development or know how to set up your environment for development?

If you're curious on how to set up your computer for development or would like to learn how to, you can check out the [README](https://github.com/thunder-app/thunder#building-from-source). If not, you can wait until this is released as part of nightly build and provide any feedback then!

micahmo:
> I think I figured out the issue - could you try again and see if it works now?

Nice! Looking much better now! Let me know if this is what you're expecting.

https://github.com/thunder-app/thunder/assets/7417301/78aae73d-fff0-4363-9533-ba269c6279b2

A couple quick thoughts...
* Right now it's not obvious to me that something is a spoiler. I'm not sure if we should have the right arrow symbol like the how HTML renders `details` or if it should look like a link or something else. Maybe we can see how other apps do it.
* Also I was thinking we shouldn't have the inkwell effect on the comment (which maybe means the comment would also handle the tap and do a collapse?). Seems like the tap event is leaking through.
* I was wondering if the title should still be displayed even once the body is expanded.

Otherwise looking great!

hjiangsu:
> Nice! Looking much better now! Let me know if this is what you're expecting.

Yup, that's what I'm expecting! There's definitely some issues with this implementation at this time as I mentioned in the PR description.

> Right now it's not obvious to me that something is a spoiler. I'm not sure if we should have the right arrow symbol like the how HTML renders details or if it should look like a link or something else. Maybe we can see how other apps do it.

I agree, we can think of a better way to show that something is a spoiler. What is implemented at the moment is mainly a placeholder since I was more focused on the overall logic first. I can add in an arrow to make it a bit more obvious. If you have any screenshots of how other apps do it, it would be great!

> I was wondering if the title should still be displayed even once the body is expanded.

I can do that!

> Also I was thinking we shouldn't have the inkwell effect on the comment (which maybe means the comment would also handle the tap and do a collapse?). Seems like the tap event is leaking through.

Could you clarify a bit more what you mean here?

machinaeZER0:
Popping in here to say that I'm personally used to spoilers being a solid block until tapped (like Sync does, apparently). I do like the title idea though, that's pretty neat! Does "test spoiler title" mean the user is able to title their spoiler, or that's just a placeholder on our end?

After some thought, I think that the solid block approach feels "simplest"/most straightforward to me, and has the extra benefit of being less likely to be misinterpreted as an external link (if we went with the spoiler turning into a title). Thoughts on that reasoning?

micahmo:
> Does "test spoiler title" mean the user is able to title their spoiler

Yes exactly. Titles are part of the spoiler markdown specification. So in this example...

```
::: spoiler test spoiler title
spoiler body text
:::
```

`test spoiler title` is the title and is supposed to be shown. `spoiler body text` is the body which is supposed to be hidden. That's why I say Sync is "wrong" in not respecting the spec. Of course, having a title at all is optional. 😊

micahmo:
> Maybe the first step to allowing spoiler support is to implement the solid block for spoilers with the assumption that there won't be additional markdown formatting within the spoiler itself

Yeah, that would be a great first step! Just having spoilers at all, even if the contents are (rarely) incorrectly formatted, would be a huge bonus!

machinaeZER0:
That makes sense to me!

micahmo (approved):
Hey I grabbed the latest and checked out the original post you made, and things seem to be working as designed! Great job on this! I approved the PR.

However, the more I thought about this, the more I thought expanding/collapsing is the way to go (and how important showing the title is). My reasoning here is that the number 1 way I've seen spoilers used on Lemmy is not for actual spoilers, but to collapse long content.

For example, here is the programming.dev sidebar. You can see they use spoilers as more of an design/organization/layout tool than anything else. And because that's how they're rendered in the web UI and Jerboa, it seems to be a very common use case.

https://github.com/thunder-app/thunder/assets/7417301/1cc3bbc7-51bf-4828-8eeb-b4cda2ea823a

Now that sidebar looks pretty funny in Thunder, where we (a) lose the heading and (b) have huge chunks of blank space, which is exactly what the authors wanted to avoid.

https://github.com/thunder-app/thunder/assets/7417301/41d0defd-98b4-4ed7-b417-d1881cd06a8b

I'm sorry I didn't think of this use case before @hjiangsu put a ton of work into changing the implementation, but I'd be curious to hear what you and @machinaeZER0 think of this and whether it's worth doing something more like the original design.

machinaeZER0:
Oh weird, I would never have assumed those were handled as spoilers! Very odd implementation, there 🤔 because of this very prominent example I'm inclined to agree with going with the title/expansion route, if only for feature parity in this particular use case!

machinaeZER0:
(That said, one could argue that their sidebar is "misusing" spoiler tags and that catering to their implementation doesn't necessarily follow what one might consider "best practices?" So I guess I could really go either way, honestly.)

micahmo:
Thanks for the feedback @machinaeZER0. I totally agree with you that if were developing a true spoiler UI feature from scratch, it would be very different and almost certainly block-based. However, unfortunately it seems like people have latched onto the default UI implementation on the web and are using it much more as a content organization tool. And I'm afraid if we don't follow suit, we'll be the odd ones out. But hey, that's how these things work. People realize that can use something a certain way that works and looks cool, so they do! Also our implementation can be a little more elegant (maybe the title can look more like a dropdown button, maybe the content can expand with an animation, etc.).

Here are a few more examples of instances/communities using this formatting in their sidebars (seems very common in the top communities).

https://feddit.de/
https://feddit.cl/
https://lemdro.id/
https://dormi.zone/
https://lemmy.world/c/lemmyshitpost
https://lemmy.world/c/news
https://lemmy.world/c/asklemmy
https://lemmy.world/c/nostupidquestions
https://lemmy.world/c/mildlyinfuriating
https://lemmy.world/c/linuxmemes
https://lemmy.world/c/youshouldknow
https://lemmy.world/c/memes
https://lemmy.world/c/reddit

hjiangsu:
I can go ahead and tweak it a bit to make it similar to how the web UI does it. I wonder though how other apps are handling this

micahmo:
> I wonder though how other apps are handling this

Here's a rundown of apps from https://join-lemmy.org/apps that I could test.

* lemmy-ui (web): expander
* Photon (web): expander
* Eternity (Android): not supported
* Liftoff (Android): not supported
* Voyager (web/Android): not supported
* Combustible: (Android): expander
* Summit (Android): expander
* Jerboa (Android): expander
* Alexandrite (web): expander
* Boost (Android): not supported
* Sync (Android): not supported (I know they use bock syntax in posts/comments though)

Note that my test case was to look at community sidebars, so if some of these say not supported (like Sync), it may work in post/comments.

hjiangsu:
Just an update on this - I finally got around to changing the implementation to use expandable content rather than spoiler blocks like before.

Here are some previews on how it looks like now, but @micahmo feel free to run it and let me know if this aligns better to how it's being used!

| Collapsed | Expanded |
|-|-|
| simulator_screenshot_90FF4FE7-0AAF-41E8-89A9-61BA24A79FB2 | simulator_screenshot_A98915CA-E138-4675-A44B-87F78CFC5AC5 |
| simulator_screenshot_500852DF-2BD8-4D0F-AC3A-7B9751F3A492 |simulator_screenshot_B354BFFF-8C5B-464D-BC3D-5D41E9905FA2 |

micahmo:
> Here are some previews on how it looks like now

I think this looks amazing and is even an improvement on the default implementation! I'm glad that we can have our own flavor in Thunder that feels familiar but is unique!

P.S. I noticed that it doesn't seem to work on the `programming.dev` instance sidebar. I'm not sure if they're using the wrong syntax, or if it's because they have nested markdown, or something else. But either way it's an edge case. This works and looks great!

hjiangsu:
> I noticed that it doesn't seem to work on the programming.dev instance sidebar. 

I just tried it and it seems to work. Is there a particular instance you have that I can test this on?

| | |
|-|-|
| 4994AAC8-00D8-4259-B258-D7A2D9C11AE0_4_5005_c | 2FA98624-1BA4-4339-BD76-25A251AC1120_4_5005_c|

micahmo:
> I just tried it and it seems to work.

You're right, it does! Not sure what I was seeing. Maybe I was confused with a community sidebar. For example, I see this on programming@programming.dev.

| image |
| - |

But as you said...

> Regardless, I think we can merge this in first, and then fix the edge cases that pop up

Completely agreed!

hjiangsu:
> You're right, it does! Not sure what I was seeing. Maybe I was confused with a community sidebar. For example, I see this on [programming@programming.dev](mailto:programming@programming.dev).

Hmm, that may be a regex issue or possibly nested markdown as you mentioned. Regardless, I'll merge this in first and then look into this a bit more!

PR #868, opened 2023-10-31
Initial ability to share and open links in Thunder on iOS

hjiangsu:
## Pull Request Description

This PR introduces the ability to open and share links on iOS though the share extension.

This differs slightly to the Android implementation in that it opens up the given post/user/comment/instance if it detects the incoming link to be a Lemmy link. This is because there's no current way to open up Lemmy links on Thunder through iOS, and it seems like the most likely use-case for sharing a Lemmy link is to open the link within the app.

This is still in draft as more testing is required (for media, and other types of links) and I would like to implement the ability to open links for communities as well if possible.

## Issue Being Fixed

Issue Number: N/A

## Checklist


ios/Share Extension/ShareViewController.swift
```diff
@@ -0,0 +1,360 @@
+//
```
hjiangsu, line 1: TODO: This file still needs to be cleaned up as it is retrieved from the template of the package.

PR #870, opened 2023-11-02
User Page Refactor

hjiangsu:
## Pull Request Description

This draft PR refactors the existing User page to allow the User page to use the FeedBloc introduced previously in the Feed refactor. This is still a draft as there are certain features which are still missing/WIP. A few things to note:

- The sidebars (community/user) have been refactored to be more modular. The shared code between the two sidebars has been merged into a new `Sidebar` widget, and all other logic has been placed into their respective widget (UserSidebar/CommunitySidebar)
- `navigateToUserPage` has been removed and is replaced by `navigateToFeedPage`
- Cleaned up some files which were importing unused packages

Known issues that have not yet been fixed from the refactor:
- [ ] User sidebar actions are not implemented (blocking user)
- [ ] When on the user page -> comments tab, you cannot perform any comment actions
- [ ] It is currently difficult to distinguish separate comments for a given user in the user page (no dividers)
- [ ] The account page is semi-broken because it relied on the previous user page implementation

> More work is needed to clean up some of the changes from the refactor

@micahmo if you'd like, you can take this for a quick run and let me know if you spot any issues that have not yet been mentioned!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
This is awesome, nice work taking this on as always.

Here are some things I noticed after taking it for a super quick spin.

* With regards to the "current user" page, I do hope it can utilize the same new user page. I know you already identified this, but here are the things that I saw were broken.
  * It no longer has a header.
  * Without using the new feed/page, it won't take advantage of some of the stuff that assumes FeedBloc everywhere. It also doesn't get sorting for comments, etc.
  * It still uses toggle buttons while the new user profiles pages use tabs. Not sure if there's any plan to align these more, but I always liked that the current user page was pretty much the same as all the other user pages with the addition of the "Saved" section.
  * Tapping the username in the main nav drawer no longer navigates to the user tab.
* Both the community and user pages now have overflow issues, and it feels like there's a lot of empty whitespace.
   image
* Also in both cases the headers no longer scroll, the feed just goes on top of them. And especially in the user case, that looks a bit funny because the tab text goes on top.
  image
* On the new user page, when opening the sidebar, the shadow is in a bit of a funny place (above the tabs).
  image

Alright, that's all for now! 😊 Nice work again!

hjiangsu:
Thanks for giving it a quick run through! I do want to sync up the current user and the user pages so that they use the same or similar shared components, although some more thought will be required to figure out how to get the saved tabs and the toggled buttons

> Both the community and user pages now have overflow issues, and it feels like there's a lot of empty whitespace.

> Also in both cases the headers no longer scroll, the feed just goes on top of them. And especially in the user case, that looks a bit funny because the tab text goes on top.

> On the new user page, when opening the sidebar, the shadow is in a bit of a funny place (above the tabs).

Hmm, I wonder if this is again some differences between Android and iOS implementations. This is what I've been noticing on my end while testing on an iPhone. I can do some more tests in the Android emulator and see if I can figure out the issues here

https://github.com/thunder-app/thunder/assets/30667958/f392ce5e-348b-4655-b95a-254cd61894d5

micahmo:
> some more thought will be required to figure out how to get the saved tabs and the toggled buttons

Agreed. I do like the current implementation and animation! I don't suppose we could still use the toggle buttons for the new user pages?

> I wanted to keep the tab bar pinned to the top rather than disappear as the user scrolled

> An alternative to this is to keep the headers scrolling as part of the feed, but then the tab bar would be above the headers

Ah ok I see what you mean! I just tried a couple apps and I think boost does this in a nice way. The header is still on top, but the tabs get pinned once they're at the top. Not sure how hard this would be to implement.

https://github.com/thunder-app/thunder/assets/7417301/4f7580df-fdd7-4ed6-a050-604428e0f2bb

hjiangsu:
> Agreed. I do like the current implementation and animation! I don't suppose we could still use the toggle buttons for the new user pages?

It might be possible to use the current implementation - I'll just have to do some adaptations for it to work. I can do a quick test and see how that looks like!

> The header is still on top, but the tabs get pinned once they're at the top. Not sure how hard this would be to implement.

Yeah, this is similar to what I was trying to achieve. It involves using slivers which is still a bit new to me, so there's definitely some improvements that can be made on that front

machinaeZER0:
Heyyyy :D

All I have to go off of is the video posted above, but I would agree (if it hasn't been updated in the interim) that the content of the page feels like it naturally lives beneath/under the header/sidebars (which follows in the current implementation, as content scrolls up and into our headers). It does feel strange then in the video to see content scrolling over the headers, so I would also vote to have scrolling push the header up with it, either off the screen entirely or in a way that mashes/transforms it into a more compact version of itself (similar to the Boost example).

Boost's is still a little odd (feels strange that stuff like the hamburger menu persists in the transition), but otherwise the implementation feels natural, I think :)

PR #873, opened 2023-11-03
Ability to search for posts

micahmo:
## Pull Request Description

This PR adds support for searching for posts.

In making this change, I extracted the list of posts from the `FeedPage` and made it into a separate `FeedWidget`.

Also note that I'm using `FeedBloc` in a slightly new way. Previously we create the bloc and it would know how to fetch posts for different feed types. However, we already have `SearchBloc` which knows how to retrieve a list of posts as search results. Therefore, I tweaked `FeedBloc` to also be able to take a list of existing posts. I hope that makes sense, it seems to work fine. :blush:

I also fixed a bug where paged calls wouldn't respect all of the parameters.

> Review without whitespace

## Issue Being Fixed

Issue Number: #273

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/ea97ecb5-1919-4274-ab60-20237af639ff

## Checklist


lib/feed/view/feed_page.dart
```diff
@@ -301,54 +297,10 @@ class _FeedViewState extends State<FeedView> {
...
-                                            context.read<FeedBloc>().add(FeedItemActionedEvent(postId: postViewMedias[index].postView.post.id, postAction: PostAction.read, value: read));
-                                          },
-                                          listingType: state.postListingType,
-                                          indicateRead: true,
-                                        )
-                                      : null,
-                                );
-                              },
-                              childCount: postViewMedias.length,
+                            FeedWidget(
```
hjiangsu, line 300: I would rename this to `FeedPostWidget` or `FeedPostList` since the feed will eventually be able to hold comments as well (for user pages)!

hjiangsu (approved):
Just one comment, otherwise it looks good!

lib/feed/view/feed_page.dart
micahmo: Good point! I went with `FeedPostList` since we already have similar (`PostCardList`) and I don't generally like putting another type name in a class name (`Widget`). Should be all set now.

PR #875, opened 2023-11-03
Fix issues with authenticated actions after external link

micahmo:
## Pull Request Description

I was having an issue when opening external links in Thunder on my physical device. The post actions were disabled because it was detecting that I was not logged in. I think the reason this happened (and the reason I didn't see it on the emulator) is that on a physical device, the link handling happens very quickly, possibly before the active account is loaded.

The fix here is to watch the `AuthBloc` in case it changes after the post is loaded. If so, all the actions will become enabled shortly after the post loads, and everything works fine.

It's also unlikely to have any negative impact as there really isn't any other way for the `AuthBloc` once a post is opened.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/post/widgets/post_view.dart
+++ b/lib/post/widgets/post_view.dart
@@ -57,7 +57,7 @@ class PostSubview extends StatelessWidget {
     final PostView postView = postViewMedia.postView;
     final Post post = postView.post;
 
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM!

PR #880, opened 2023-11-06
Refactor link handling logic

micahmo:
## Pull Request Description

Currently we only support smart navigation from markdown bodies. That means, for example, if someone creates a post where the URL is a link to another post, we don't do in-app navigation.

This PR extracts the link handling logic to a common place, to be used anywhere we open a link. In fact, I made the original `openLink` private so we don't call it by accident. :blush:

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/ad38d86e-28c2-4507-a265-5a2edb8f4274

## Checklist


ggichure (approved):

hjiangsu (approved):
LGTM! Thanks for working on some refactoring, this will definitely help consolidate some of the logic 😄

PR #882, opened 2023-11-07
Support search by URL

micahmo:
## Pull Request Description

This PR addresses the URL search type.

image

This one's a strange beast because, while the others like Posts and Comments, denote the _type_ of object to search, URL denotes a _field_ to search. And in fact, by looking at the Lemmy server code, you can see that it only ever returns Post results.

https://github.com/LemmyNet/lemmy/blob/cb01427dcff14b3d88e30220695fc97978786a9a/crates/apub/src/api/search.rs#L183-L184

Therefore, I attempted to implement this as a smart toggle.

* When the search type is posts, the user can toggle whether to do a URL search or not.
* Furthermore, if the entered query is parseable URL, we will automatically select URL as the search type (of course, the user can always switch back to text search).

Hopefully this makes sense and is a reasonable implementation of this feature in our UI.

## Issue Being Fixed

Issue Number: #273

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/72018d87-c487-42d8-bd04-8f8d6e0f4192

## Checklist


hjiangsu (approved):
LGTM!

PR #887, opened 2023-11-09
Add comment search

micahmo:
## Pull Request Description

This PR introduces a basic comment text search within a post. There is a new FAB button from which you can initiate a search and enter a search term. After that, if there are any matches, the first matching comment will be highlighted. At that point, the main FAB button will become a "next result" button, allowing you to cycle through all the results. Then you can expand the FAB to cancel the search using the same button that initiated it.

Limitations
* We can currently only scroll to top-level comments, so if there is a nested result, it might not be obvious (but it will be highlighted). This is a general problem and not new here.
* With these changes, we are not specifically highlighting the matched word/phrase, only the whole comment.

Notes
* I am using the `newlyCreatedCommentId` property of the `PostState` to highlight the search results. Since this property is really only used for highlighting, it could be renamed to a more general `highlightedCommendId` or similar. But that would be more of a refactoring task for the future.
* I did ensure that, if cycling through results causes additional paging/fetching, we will stay in search mode.
* Performing any other action (vote, save, etc.) will take us out of search mode.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/93a5cac7-9604-41e9-a874-c000e487d010

## Checklist


ggichure (approved):
LGTM

hjiangsu (approved):
This looks really awesome! Thanks for taking this on, it'll definitely be useful to have a search function for comments.

> I did ensure that, if cycling through results causes additional paging/fetching, we will stay in search mode.

Just for a bit of clarification here, if additional fetching happens, will the search results be updated (e.g., if the additional comment fetch retrieves a comment with the given search term `firefox`, will that new comment will be added into the results of the existing search)?

micahmo:
> Just for a bit of clarification here, if additional fetching happens, will the search results be updated (e.g., if the additional comment fetch retrieves a comment with the given search term `firefox`, will that new comment will be added into the results of the existing search)?

Yes exactly. You can't tell in the video because the fetching happens off-screen, but at one point the "next result" button puts us within the threshold of fetching more comments. Some of those new comments contain the search term, so pressing "next result" again goes to the new one rather than back to the top.

hjiangsu:
> Some of those new comments contain the search term, so pressing "next result" again goes to the new one rather than back to the top.

Awesome! Let me know if this is all good to be merged 😄

PR #893, opened 2023-11-15
Ability to Set Post Language and Edit Existing Posts

hjiangsu:
## Pull Request Description

This is a draft PR which adds the ability to edit existing posts, and set the specified language for the post. It also adds a few UI changes to make the experience a bit more cohesive. This is currently in draft since more testing is required.

With the addition of language selection, a new autocomplete has been added in to select and change the language.

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu:
@micahmo Feel free to review the PR if you'd like or do some testing! I think most cases were caught, but since this is a fairly big refactor, some things may have been missed (which should hopefully be caught when this is released to the nightly builds)

lib/community/pages/create_post_page.dart
```diff
@@ -32,51 +41,102 @@ import 'package:thunder/utils/instance.dart';
 class CreatePostPage extends StatefulWidget {
   final int? communityId;
   final CommunityView? communityView;
-  final void Function(DraftPost? draftPost)? onUpdateDraft;
-  final DraftPost? previousDraftPost;
 
-  // used create post from action sheet
+  /// Whether or not to pre-populate the post with the [title], [text], [image] and/or [url]
+  final bool? prePopulated;
```
micahmo, line 46: I was wondering if we might want to think about consolidating all the different ways we can prepopulate a post. Right now I think we have drafts, edits, and shares, each with different things that are saved/restored. (For example, is language selection saved with drafts?) Maybe we can have a wrapper class around a `Post/View/Media/whatever`, so that it can contain every possible field, and then an enum saying what kind of thing it is. Just a thought.

micahmo (approved):
Very nice, I especially like the draft refactoring!

lib/community/pages/create_post_page.dart
hjiangsu: Thats a good idea! I think we could do this in a different PR since this one already contains a pretty substantial change if thats all good with you.

micahmo: Of course!

PR #905, opened 2023-11-21
Add ability to open links in Thunder from Safari extension

hjiangsu:
## Pull Request Description

This PR adds the ability to open Thunder using a Safari extension. Note that this is slightly different that #868. #868 allows for sharing and opening links in Thunder, whereas this is specifically used to open links via Thunder. Safari extensions are only available in iOS 15 and later, whereas the share extension from #868 can be used in older versions of iOS.

With the Safari extension, Safari will prompt the user whether they would like to open a Lemmy link through Thunder. As of right now, the list of Lemmy instances is static. However, we could add in a GitHub workflow which updates the list whenever we update the list of instances.

This is ~still in draft~ as the following is still required:
- [x] Update placeholder icons with Thunder icons
- [x] Additional testing with instance, post, and comment links

While working on this, I've noticed that there are cases where the `resolve_object` does not return back properly. This seems to be an existing issue, not one that was introduced with this change.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
> I've noticed that there are cases where the `resolve_object` does not return back properly.

It will fail for things that are not federated. If it's failing for things that are, then I'm not sure. 😊

ios/Open In Thunder/Resources/content.js
```diff
@@ -0,0 +1,100 @@
+let instances = [
+    "aussie.zone",
```
micahmo, line 2: Is there any way to put this list in another file by itself? That would make it easier to update via the GitHub Actions.

ios/Open In Thunder/Resources/manifest.json
```diff
@@ -0,0 +1,95 @@
...
+        "96": "images/icon-96.png",
+        "128": "images/icon-128.png",
+        "256": "images/icon-256.png",
+        "512": "images/icon-512.png"
+    },
+
+    "content_scripts": [{
+        "js": [ "content.js" ],
+        "matches": [
+            "*://aussie.zone/*",
```
micahmo, line 20: I'm guessing this probably can't be moved to its own file. If not, we can just have the workflow recreate this whole thing. 😊

micahmo (approved):
I don't have any way to test this, but the code looks good! Would love to se a demo!

ios/Open In Thunder/Resources/content.js
hjiangsu: There might be a way, but this was the simplest way I found to do this - I tried to place it in a different file but it wouldn't import it. Not sure if this is a restriction or a setting that I would have to tweak to allow that to work!

ios/Open In Thunder/Resources/manifest.json
hjiangsu: Yeah, this is similar to the other comment - I don't think there's a way to import things into JSON files from other files but I might be wrong

hjiangsu:
@micahmo I added 38ba707c8aaa01548694567ad442ba9fc10266d2 which adds the GitHub action to update the instances! I'm not entirely sure if it'll work properly since I don't know how to run GitHub actions locally, but I tested it through the command line and it seems to be okay.

Could you take a quick look over it and see if it all makes sense? Thanks!

micahmo:
> Could you take a quick look over it and see if it all makes sense? Thanks!

Yeah it seems reasonable to me! Worst case, the PR that the workflow opens will be bad and we'll just have to tweak it from there. 😊

hjiangsu:
Sounds good, I'll go ahead and merge it in!

hjiangsu:
Hmm, it seems like it's writing the new line character instead of treating it as a new line:

https://github.com/thunder-app/thunder/pull/923/files

micahmo:
> Hmm, it seems like it's writing the new line character instead of treating it as a new line:

Hmm, interesting. Maybe it needs to be escaped or something. Needs some more experimentation. If you want, we can add a [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) to the workflow and then we'd be able to trigger it manually on a testing branch.

hjiangsu:
> If you want, we can add a [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) to the workflow and then we'd be able to trigger it manually on a testing branch.

That sounds like a good idea!

micahmo:
Oh wait, it already has it! So you can run a test on any branch like this.

| image |
| - |

micahmo:
I think I got it! [Someone suggested](https://stackoverflow.com/questions/8467424/echo-newline-in-bash-prints-literal-n) `printf` and it works.

Workflow change: https://github.com/thunder-app/thunder/compare/develop...micahmo:thunder:test/action

Result: https://github.com/micahmo/thunder/pull/11/files

hjiangsu:
> I think I got it! [Someone suggested](https://stackoverflow.com/a/8467449/4206279) printf and it works.

Nice! Thanks for looking into it. I was also testing some alternatives but it didn't work  😅

PR #914, opened 2023-11-23
Fix search filter terms

micahmo:
This is a super small PR which updates the search filter terms to match the web UI.

See this discussion: https://github.com/thunder-app/thunder/issues/273#issuecomment-1815412234


hjiangsu (approved):
LGTM!

PR #916, opened 2023-11-24
Fix indexing into an empty display name

micahmo:
This is a super small PR which fixes an issue where we were attempting to index into an empty string (if the user's displayname was empty but not null).


hjiangsu (approved):
LGTM!

PR #917, opened 2023-11-27
Improve community picker

micahmo:
## Pull Request Description

This PR adds a subscriber count and subscription indicator to communities in the input dialog autocomplete dropdown.

I tried to figure out how to add this additional info without compromising the already limited horizontal screen real estate. For subscriber count, I ended up putting it under the community icon. And for the subscription indicator, I put it before the community full name. I'm open to suggestions for improving this!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/323d8371-f76c-4dab-8bb0-582d85177d9e

## Checklist

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -332,6 +332,9 @@
   "password": "Password",
   "@password": {},
   "pending": "Pending",
+  "@pending": { 
+    "description": "Describes the subscription status for 'Pending'" 
+  },
   "postBody": "Post Body",
   "@postBody": {},
   "postLocked": "Post locked. No replies allowed.",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -331,6 +331,7 @@
   "@overview": {},
   "password": "Password",
   "@password": {},
+  "pending": "Pending",
   "postBody": "Post Body",
   "@postBody": {},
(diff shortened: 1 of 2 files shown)


hjiangsu:
Hmm, I wonder how it would look like if we were to add an additional line to each `ListTile` to show subscription status and sub counts? Just a suggestion!

micahmo:
Sure, I like both of those suggestions. How's this?

| image |
| - |

lib/l10n/app_en.arb
```diff
@@ -331,6 +331,7 @@
   "@overview": {},
   "password": "Password",
   "@password": {},
+  "pending": "Pending",
```
hjiangsu, line 334: I would add in some metadata description for this to explain what or where this is used! It would be something like this:

```
"pending": "Pending",
"@pending": {
  "description": "Describes the subscription status for 'Pending'"
}
```

micahmo:
> I think that looks good!

Great! Should be good to go then!

hjiangsu (approved):
Just one comment for localization descriptions!

lib/l10n/app_en.arb
micahmo: All set!

Issue #45, opened 2023-06-22
Ability to view  images within the app hosted on external sites such as Imgur


Terkyz:
**Is your feature request related to a problem? Please describe.**
Yes, since it would be much more optimal to view the images directly from the app without having to open the integrated browser.

**Describe the solution you'd like**
Have the image rendered in the app like any other image hosted directly on a Lemmy instance.

**Describe alternatives you've considered**
I can open the link manually and it doesn't present any problem, but this is more optimal.

**Additional context**
It would be super cool to also have [Telegram-style previews of articles](https://i.imgur.com/KfsXPU1.jpg), instead of [this](https://i.imgur.com/CaDjOsR.jpg), but I'm not sure if this needs its own issue.

hjiangsu:
For this, it really depends on the image provider's APIs. 

Integration with Imgur is definitely possible, but it might require API access, and depending on their policies, there might be a cost associated with it (e.g., like what Reddit is doing, except much more reasonable)

>It would be super cool to also have [Telegram-style previews of articles](https://i.imgur.com/KfsXPU1.jpg), instead of [this](https://i.imgur.com/CaDjOsR.jpg), but I'm not sure if this needs its own issue.

That is definitely possible - currently theres only limited support for handling links (it'll fetch the image, and maybe the title of the article, but nothing more). I would suggest you to create a new issue for that to be tracked separately!

Generator:
in addition to this, maybe on pixelfed too, user login to his pixelfed account, and upload to there and share url image on lemmy.  
Not saying as a priority, maybe in a distance future

mario-g98:
May I add RedGIFs to the list? 😄

Incur and RedGIFs (and others probably) are nicely implemented in [Infinity For Reddit](https://github.com/Docile-Alligator/Infinity-For-Reddit), maybe just take a look in their code?

PR #919, opened 2023-11-27
Open images in-app

micahmo:
## Pull Request Description

This PR allows us to open images from links using the in-app image viewer.

## Issue Being Fixed

Issue Number: #45

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/65e3290e-0d77-449b-8591-8581d70b5e55

## Checklist

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -4,6 +4,7 @@ import 'dart:ui';
 import 'package:extended_image/extended_image.dart';
 import 'package:flutter/material.dart';
 import 'package:thunder/shared/image_viewer.dart';
+import 'package:thunder/utils/image.dart';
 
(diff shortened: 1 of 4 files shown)


hjiangsu:
Looks good! One thing I will note is that this doesn't fully solve what the original issue was since it only works with URLs that are images (end in an image format .png, .jpg, .jpeg, etc.).

This means it won't be able to handle URLs such as https://imgur.com/gallery/OQUvq1K which don't give the raw image URL. I do think that can be treated as a separate issue though (as we would need to implement a wrapper for the different APIs to retrieve that information)!

hjiangsu (approved):

micahmo:
Agreed! I didn't want to dive into any API stuff just yet. I figured if we have the raw image URL, we can at least show it in app!

Issue #913, opened 2023-11-23
Option to hide read posts by default


tj9fu5bu:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
An option to automatically hide read posts in subscriptions or per community.

**Describe alternatives you've considered**
Manually hiding all posts when the page is refreshed or app is reloaded.

**Additional context**
Issue might be encountered if a huge number of posts are already read.

hjiangsu:
I think the first step to this is to add an option which syncs up to your Lemmy account settings. Lemmy already has a setting to allow hiding or showing read posts.

This option wouldn't be as granular has being able to choose to hide posts in Subscriptions or on a community-basis, but that might be something to be explored in the future.

tj9fu5bu:
@hjiangsu this feature seems to be broken with v0.19.0 instances.

hjiangsu:
Just wondering, but did your instance recently upgrade to 0.19.0? If so, you may have to log out and log back in for things to work! There were authentication changes made behind the scenes so this might be what's causing the issues.

Let me know if that fixes the issue or not.

hjiangsu:
It seems like this may actually be on Lemmy's end with the 0.19.0 update: https://github.com/LemmyNet/lemmy/issues/4290

tj9fu5bu:
Oh yes, I've logged out and logged back in to get back my subscribed communities. I guess it's really on Lemmy's end then!

hjiangsu:
Could you let me know if this is fixed? It seems like a fix was pushed in Lemmy 0.19.1. If it is fixed, I'll mark this issue as closed!

tj9fu5bu:
I can indeed confirm that this has been fixed in 0.19.1 instances!

@hjiangsu

PR #921, opened 2023-11-27
Add setting to show or hide read posts

hjiangsu:
## Pull Request Description

This PR introduces a setting to toggle on and off read posts which syncs up with the currently active Lemmy account. I've added the general logic to allow fetching and updating the current user's account settings. This will allow us to more easily add in options in the future on a per-account basis.

Since this setting is synced with the Lemmy account, there's no additional logic required to check if the given post being fetched has been read or not.

An idea for a future improvement would be a account selector within this page to allow us to switch between the different logged-in accounts.

## Issue Being Fixed

Issue Number: #913

## Screenshots / Recordings

| 0.18.x | 0.19.x |
|-|-|
|simulator_screenshot_1F989691-6976-4D76-BF42-95C51D50D8F5|simulator_screenshot_ACF0A276-89B1-4B04-809F-AB6FD06E127C|

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Added
 - Added ability to collapse post in post page
 - Added ability to change app language in settings
+- Added ability to show/hide read posts in user settings
 
 ## 0.2.6 - 2023-11-22
 ### Fixed
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -591,7 +591,7 @@
   "username": "Username",
   "@username": {},
   "users": "Users",
-  "userSettingDescription": "These settings sync with your Lemmy account and are only applied on a per-account basis.\n\nYou are currently viewing the settings for: {user}",
+  "userSettingDescription": "These settings sync with your Lemmy account and are only applied on a per-account basis.",
   "@userSettingDescription": {
     "description": "Description which explains that the settings are applied globally to the current user."
   },
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -181,6 +181,10 @@
   "@feed": {},
   "fetchAccountError": "Could not determine account",
   "@fetchAccountError": {},
+  "filters": "Filters",
+  "@filters": {
+    "description": "Category to describe various filters (user, community, instance blocking)"
+  },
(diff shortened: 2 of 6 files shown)


micahmo:
Looks great!

Would it make sense to shrink some of the space here so that elements are relatively evenly spaced?

| image |
| - |

Also for identifying the current user, would it make sense to show the whole user widget like we do in the Create Post page?

hjiangsu:
> Would it make sense to shrink some of the space here so that elements are relatively evenly spaced?

I noticed that too - I can see if I can shrink it down a bit more!

> Also for identifying the current user, would it make sense to show the whole user widget like we do in the Create Post page?

Good idea, I'll see if I can add that in quickly

micahmo:
Looks great!

I wonder if we even need the "you are current viewing..." or if it would be sufficient to have the user widget at the top (maybe above the first paragraph). But I'll leave that up to you. It's good as is!

hjiangsu:
Done, removed current user text and moved the user indicator to the top of the page!
simulator_screenshot_EFC01939-3D20-4CDB-AAEA-692F0F670759

micahmo:
Yeah that looks really nice!!

P.S. It looks like there's an error on this page in 0.19. Not sure if it's new, so it can probably be addressed separately.

hjiangsu:
> P.S. It looks like there's an error on this page in 0.19. Not sure if it's new, so it can probably be addressed separately.

Hmm, what's the error? I can't reproduce it on my end

micahmo:
> Hmm, what's the error? I can't reproduce it on my end

Voyager seems to be having some issues at the moment, so I can't log in, but I will try again when I can!

micahmo:
Here you go, let me know if this helps!

https://github.com/thunder-app/thunder/assets/7417301/fa5b9c60-af78-487a-8e68-e9fc621d0d05

hjiangsu:
Just curious, but are you able to login to voyager.lemmy.ml (on the web UI)? I think the server was reset a while back so all existing accounts there were deleted.

I had to create a new account to log in because I had issues using my old credentials.

micahmo:
Yes I also had to recreate my account a while back, and yes I can still log into the web UI. However, I tried logging out in Thunder and back in, and that seemed to fix the issue. So I'll just blame it on some Voyager weirdness. 😊 Thanks!

PR #925, opened 2023-11-28
Add support for community and creator search filters

micahmo:
## Pull Request Description

This PR adds support for community and creator filters in search. This should bring Thunder's search in parity with the web UI!

## Issue Being Fixed

Issue Number: Not specifically called out in #273, but related.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/d8123a18-36f1-4223-b2ea-ebf833201c84

## Checklist


hjiangsu (approved):
Looks good to me! I do wonder if there might be a better way UX-wise to show these two filters. Right now, it seems like you have to scroll in order to see the options which might cause user's to not realize those options are available.

I do think that can be tackled at a later point in time, as this change allows us to be in parity with the web UI. Great work!

micahmo:
> I do wonder if there might be a better way UX-wise to show these two filters.

Agreed! We can get these two new options in using the same UI as the other options, then we can re-evaluate how best to show all of them.

PR #927, opened 2023-11-29
build for web

gwbischof:
There might be a couple more touch-ups to fix, but the app is running for web now.
this PR uses the package `sqflite_common_ffi_web` which is the web version of the package thunder is currently using for sqlite.

Is this a feature that you want to support?


hjiangsu:
Hey! Just did a very quick review of the changes, and I'm just curious about something.

```dart
(!kIsWeb && Platform.isIOS)
```

Are these checks necessary? I was under the assumption that the platforms are mutually exclusive, so these checks wouldn't be required. I haven't ran this locally myself so I'm not too sure but just wondering!

On a different note, I think `Platform.isWeb` is also available, so we could use that instead of `kIsWeb` for consistency: https://pub.dev/documentation/platform_builder/latest/platform/Platform/isWeb.html

gwbischof:
Good question, I just double-checked this.

This is the error that I was getting: https://stackoverflow.com/questions/58459483/unsupported-operation-platform-operatingsystem

These are the docs for the Platform class that is in Thunder: https://api.dart.dev/stable/3.2.2/dart-io/Platform-class.html
The Platform class from the platform_builder package is a different Platform class, the one that we are using doesn't have `isWeb` and raises an Exception when its used for web.

gwbischof:
I am going to close this PR and open a new one that also includes the docker stuff for web.

PR #938, opened 2023-12-01
Support SVG previews in markdown

micahmo:
## Pull Request Description

I was investigating issues I've seen displaying some images (most often in instance/community sidebars). It turns out that a common culprit is SVG files, which `ExtendedImage` cannot render.

This PR adds the ability to check whether an image is an SVG (getting only the headers, so the http request is lightweight), and if so, using a different widget to render the SVG (from [jovial_svg](https://github.com/zathras/jovial_svg)) in markdown. This works in post bodies, comment bodies, and sidebars. It doesn't affect when a post link references an SVG. Like the web, that is still treated as an external link. It also doesn't allow SVGs to be tapped and opened in an image viewer.

I also enhanced the error messaging for when images legitimately don't load.

> Note: I picked [jovial_svg](https://pub.dev/packages/jovial_svg) over the seemingly more popular [flutter_svg](https://pub.dev/packages/flutter_svg) because the latter failed to render the first thing I threw at it! :blush:

## Issue Being Fixed

Inability to render SVGs.

## Screenshots / Recordings

> Ignore the oversized Datadog image and render overflow. I would like to fix that separately!

### Before

https://github.com/thunder-app/thunder/assets/7417301/c9be6d4b-d919-4b22-9d3c-630a44a20e56

### After

https://github.com/thunder-app/thunder/assets/7417301/e3285a17-24d3-430c-b909-b87f4bb0ede3

## Checklist


lib/shared/common_markdown_body.dart
```diff
@@ -43,19 +45,32 @@ class CommonMarkdownBody extends StatelessWidget {
...
-            mainAxisAlignment: MainAxisAlignment.start,
-            children: [
-              ImagePreview(
-                url: uri.toString(),
-                isExpandable: true,
-                isComment: isComment,
-                showFullHeightImages: true,
+        return FutureBuilder(
+          future: () {
+            return isImageUriSvg(uri);
```
hjiangsu, line 50: I believe we can just do this?

```dart
future: isImageUriSvg(uri),
```

lib/shared/common_markdown_body.dart
```diff
@@ -43,19 +45,32 @@ class CommonMarkdownBody extends StatelessWidget {
...
+          }(),
+          builder: (BuildContext context, AsyncSnapshot<dynamic> snapshot) {
```
hjiangsu, line 52: I would change this to show the correct response type instead of `dynamic`!

```dart
builder: (BuildContext context, AsyncSnapshot<bool> snapshot) {
```

hjiangsu (approved):
LGTM! Just a couple of comments

lib/shared/common_markdown_body.dart
micahmo: Yes! Done.

lib/shared/common_markdown_body.dart
micahmo: Yep, done!

PR #939, opened 2023-12-01
Do not artificially stretch images

micahmo:
## Pull Request Description

As alluded to in #938, this PR fixes issues with image previews where small images could be stretched out, causing overflows and blurriness.

## Issue Being Fixed

Issue Number: N/A

### Before

| qemu-system-x86_64_Kw4rwYKKQq |
| - |

### After

| after |
| - |

## Checklist

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -92,11 +92,13 @@ class _ImagePreviewState extends State<ImagePreview> {
                           maxHeight: MediaQuery.of(context).size.width * 0.55,
                           maxWidth: MediaQuery.of(context).size.width * 0.60,
                         )
-                      : null,
+                      : BoxConstraints(
+                          maxWidth: MediaQuery.of(context).size.width - 24,
+                        ),
                   alignment: widget.isComment == true ? Alignment.topCenter : Alignment.center,
                   widget.url!,
                   height: widget.height,
-                  width: widget.width ?? MediaQuery.of(context).size.width - 24,
+                  width: widget.width,
                   fit: BoxFit.cover,
                   cache: true,
                   clearMemoryCacheWhenDispose: false,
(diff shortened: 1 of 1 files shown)


lib/shared/image_preview.dart
```diff
@@ -96,7 +96,7 @@ class _ImagePreviewState extends State<ImagePreview> {
                   alignment: widget.isComment == true ? Alignment.topCenter : Alignment.center,
                   widget.url!,
                   height: widget.height,
-                  width: widget.width ?? MediaQuery.of(context).size.width - 24,
+                  width: widget.width,
```
micahmo, line 99: Note that this removes any artificial image sizing, aside from what gets passed in (usually for feed previews). Also note that we still have a max width (enforced above this), so we should not exceed the width of the device. In other words, this should allow images to be smaller, but not bigger. I hope there aren't any unintended side-effects that I'm not thinking of. 😊

hjiangsu: Are you able to test this out with different post layouts?
- compact view
- card view: with view full height images, edge-to-edge images toggled on/off

I think if those look good, then it should be fine!

micahmo: Sure! Let me know if this is good.

### Compact

https://github.com/thunder-app/thunder/assets/7417301/70e2d2bd-23b8-4f7f-985a-d31e8e948da7

### Card, Full Height Off, Edge-to-Edge Off

https://github.com/thunder-app/thunder/assets/7417301/547e0389-3c31-4ff1-905c-33ca8a896750

### Card, Full Height On, Edge-to-Edge Off

https://github.com/thunder-app/thunder/assets/7417301/0053adf1-9111-44fb-b73d-f36ece9c8475

### Card, Full Height Off, Edge-to-Edge On

https://github.com/thunder-app/thunder/assets/7417301/d355a10e-0621-4333-9228-0f46e2060cd4

### Card, Full Height On, Edge-to-Edge On

https://github.com/thunder-app/thunder/assets/7417301/82919dc5-7caa-48cf-b815-e2ff0feb4d9c

hjiangsu: It seems to look good! Thanks again for doing this

hjiangsu (approved):

lib/shared/image_preview.dart
hjiangsu: It seems to look good! Thanks again for doing this

PR #946, opened 2023-12-04
remove dotenv

gwbischof:
The .env file is needed to build the app.
Github workflows to build the app will fail without it.

I did a bit of googling and I didn't find a way to change the pubspec.yaml (which uses the .env) to handle a missing .env file.

Adding an empty .env file removes a step from getting thunder running.
I left the .env in the gitignore so peoples changes to .env wont accidentally get pushed.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-# Empty Environment File
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+# Empty Environment File
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,3 @@ app.*.map.json
 /android/app/debug
 /android/app/profile
 /android/app/release
-
-# Environment files
-.env
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@
 
 <hr />
 <p>
-Hey there! Just wanted to let you know that this repo is currently my personal side project to build something cool while learning about Dart and Flutter.  
+Hey there! Just wanted to let you know that this repo is currently my personal side project to build something cool while learning about Dart and Flutter.
 </p>
 <p>
 Contributions to this project are always welcomed, and in fact, even strongly encouraged here! Since I am only able to work on this during my spare time, any contributions from the community is valuable. If you are a developer, feel free to tackle any issues present.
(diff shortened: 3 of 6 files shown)


micahmo:
> I left the .env in the gitignore so peoples changes to .env wont accidentally get pushed.

Unfortunately that doesn't work. 😞 The `.gitignore` only has effect when Git is considering which files to track. Once you've tracked a file (either it wasn't in `.gitignore` at the time, or you force added it, etc.), the `.gitignore` has no effect.

As a test, I grabbed your branch and appended to the `.env` file, and you can see that it has a pending change. And I tailed the `.gitignore` to show it's still there.

image

If the workflow needs the file, maybe it can create it on the fly?

gwbischof:
Ohh thanks! I didn't consider that.  I wonder if there is some better way to have a default .env file.

gwbischof:
I removed .env from the git ignore, because it do anything.
I did another search and couldn't figure out how to get this line not to error out with the .env doesn't exist: https://github.com/thunder-app/thunder/blob/develop/pubspec.yaml#L131

gwbischof:
Here is a related issue:
https://github.com/java-james/flutter_dotenv/issues/70

hjiangsu:
I believe we might just be able to get rid of `dotenv` package and related implementation completely. It was previously used for Sentry to allow me to detect errors in production but its no longer used.

In the future, we might want to add in secrets for API endpoints (Imgur, or other APIs). If that's the case, then we can revisit this discussion. Thoughts?

gwbischof:
This is a decent summary: https://www.sandromaglione.com/articles/how-to-use-environmental-variables-in-flutter

I think we can use `--dart-define` to pass environment variables or secrets.
The build scripts can pass the environment variables to `flutter build` with `--dart-define`.
If you are using github workflows to build, then you can use github secrets to store your secrets and pass them into the build: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions

And .vscode/launch.json can be used to pass variables for development.

So I think getting rid of the dotenv makes sense.

gwbischof:
Just updated it and removed it, I can change it back if that not what you want

hjiangsu:
Looks good to me!

hjiangsu (approved):

Issue #253, opened 2023-07-06
[Main issue thread] Long-press menu on comments


krestenlaust:
**Is your feature request related to a problem? Please describe.**
Some of the features missing would make sense to introduce in a comment-context menu.

**Describe the solution you'd like**

- [x] #252 - Report comment
- [x] #232 - Share comment
- [x] #231 - Delete own comment
- [x] #251 - Copy comment text
- [x] #958 - Navigate to user profile

**Describe alternatives you've considered**
The actions should also be implemented as swipe gestures, that the user can customize but aren't active by default.

**Additional context**
#189 (Original issue)

micahmo:
It looks like Report is the only outstanding action from this request. Would it make sense to create a new issue for just that one action?

hjiangsu:
Adjusting the priority here because I think we should have the ability to report and delete comments (reporting in particular). For deleting comments, one way to do it would be to edit the comment instead to show nothing (not fully ideal, but it is a workaround)

micahmo:
I thought delete was done in #415? And report is the only outstanding thing here?

hjiangsu:
Oh shoot, I forgot about that one 😆 - yeah, in this case, just reporting comments/users would the the last outstanding issue!

Toldry:
Please also add the option to go to the user's profile page from the comment context menu

krestenlaust:
Awesome work!!

PR #958, opened 2023-12-05
Improve comment/post actions

micahmo:
## Pull Request Description

This PR updates the Post and Comment actions to bring them more in parity with other functionality in the app and with each other.

> Review without whitspace.

#### Post - Added
 - Block User

#### Comment - Changed
 - Moved Share to top row (like Posts) to make room for more options

#### Comment - Added
 - Visit User Profile
 - Block User
 - Visit Instance
 - Block Instance (when supported)

I also added a Block/Unblock Instance to the Visit Instance page (when supported).

> P.S. In a future update, I would like to replace "user"/"community", etc. with the actual names.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/37ee7e91-ec33-4361-970d-0bb2562a453f

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 - Added the ability to render SVGs in markdown bodies - contribution from @micahmo
 - Added Safari extension to open Lemmy links in Thunder
 
+## Changed
+- Added new items to the Post and Comment actions sheet - contribution from @micahmo
+- Added ability to block instance from instance page - contribution from @micahmo
+
 ### Fixed
 - Fixed issue where custom tabs would not respect default browser when opening links
 - Fixed issue where initial app startup takes a long time to load (503 errors)
--- a/lib/community/utils/post_card_action_helpers.dart
+++ b/lib/community/utils/post_card_action_helpers.dart
@@ -21,6 +21,7 @@ import 'package:thunder/shared/advanced_share_sheet.dart';
 import 'package:thunder/shared/picker_item.dart';
 import 'package:thunder/shared/snackbar.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+import 'package:thunder/user/bloc/user_bloc.dart';
 import 'package:thunder/utils/instance.dart';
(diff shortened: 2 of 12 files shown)


machinaeZER0:
This is nice! I will say, as the list of things we can do on long press grows, it might make sense to start consolidating similar actions? Boost has a setup like this that i like quite a bit, though for some reason it's not accessed through a long press. Screenshots below:

Screenshot_20231205-230203
and on pressing one of the option groups:
Screenshot_20231205-230215

Thoughts on trying something similar here? I could see grouping the options above into "View..." (personal preference over Visit) and "Block...", or keep the Visit actions exposed and just consolidate the Block actions (the way Boost does).

The biggest benefit is saving space while keeping taps minimal - and that space could eventually be used to add other options like Boost's menu has, such as reporting/sharing/saving/etc (if desired). Would love to hear your thoughts!

micahmo:
> it might make sense to start consolidating similar actions

We can definitely work on optimizing the long-press menu. We already moved several options to the top row (imagine how big it would be if each of those were a separate row!). There are a couple approaches to take. Like Boost, we could group things by action. Sync takes a slightly different approach and groups things by entity (for example, you'd click on the person first, then you could block, visit, etc.).

It would be nice if we could minimize taps instead of requiring a second menu. At one point I tried putting secondary action buttons in the same row, like this (if you can picture what I mean). But it got a little too busy.

| Actions | |
|--------|--------|
| Visit Profile | 🚫 |
| Visit Community | 🚫 |

Anyway, as usual, we can get this PR in first with the functionality, and then discuss a possible UI upgrade! If anyone wants to make mockups, that could help a lot!

machinaeZER0:
For what it's worth, I actually primarily use the long press menu to block communities as I'm browsing the main feed, so I love that we currently have it exposed! Hear what you're saying though :)

hjiangsu:
Linking this issue as it relates to this: #253

PR #960, opened 2023-12-06
Fix select community alignment

micahmo:
## Pull Request Description

This is a small PR which adjusts the height of the "Select a community" placeholder on the post creation page. Currently, the placeholder is slightly smaller than the final community indicator, meaning that when the community is selected, the height changes and shifts all the elements down. Now everything stays put.

> Review without whitespace.


hjiangsu (approved):
LGTM!

PR #963, opened 2023-12-06
Disable post -> Block User when logged out

micahmo:
## Pull Request Description

This is a super tiny follow up to #958 which ensures that the Block User option is disabled when not logged in.


hjiangsu (approved):
LGTM!

PR #964, opened 2023-12-06
Show version on instance page

micahmo:
## Pull Request Description

This PR adds the version of the instance to the heading on the instance page.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| |
|--------|
| image |
| image | 

## Checklist

### DIFF
--- a/lib/account/pages/login_page.dart
+++ b/lib/account/pages/login_page.dart
@@ -90,7 +90,7 @@ class _LoginPageState extends State<LoginPage> with SingleTickerProviderStateMix
         instanceTextDebounceTimer!.cancel();
       }
       instanceTextDebounceTimer = Timer(const Duration(milliseconds: 500), () async {
-        await getInstanceIcon(_instanceTextEditingController.text).then((value) {
+        await getInstanceInfo(_instanceTextEditingController.text).then((value) {
           // Make sure the icon we looked up still matches the text
           if (currentInstance == _instanceTextEditingController.text) {
             setState(() => instanceIcon = value.icon);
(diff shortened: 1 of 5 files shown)


hjiangsu:
Nice! I was also thinking of adding this information somewhere - I wonder if we could also add in the instance version in the account selector bottom sheet (if its not too much work)? I was thinking maybe in between the instance name and the ping time

micahmo:
> I wonder if we could also add in the instance version in the account selector bottom sheet

Funny you say that, I actually tried that first! On my emulator (with an admittedly low DPI), there is often a little bit too much info to display horizontally.

image

But on a physical device it's probably fine. I can just make sure the data is properly truncated, and maybe put the whole thing in a tooltip. What do you think>

hjiangsu:
> But on a physical device it's probably fine. I can just make sure the data is properly truncated, and maybe put the whole thing in a tooltip.

I think it should be fine on a physical device as you said! But I agree with truncating the data if possible and adding a tooltip for it. I personally prefer the instance being to the left of the ping time since that'll be useful information for any instance-based issues, but I'll let you make the final call on where its placed!

micahmo:
> I personally prefer the instance being to the left of the ping time

Agreed!

> I agree with truncating the data if possible

Since truncating is a bit tricky (multiple widgets, no way to know which one needs to be truncated), what do you think about a `Wrap`? Again, this should only be an issue on low DPI devices.

| image |
| - |

hjiangsu:
> what do you think about a Wrap?

I think that works! It's always tricky to work with low DPI or large text scaling, but at least we'll be able to show all information on the page. It would be nice if there was a guide for how to handle UI for these cases 😅

micahmo:
Alright, just pushed these changes! Review without whitespace. 😊

PR #965, opened 2023-12-07
Some tweaks to the user page state

micahmo:
## Pull Request Description

This PR addresses a couple small frustrations with the user page.

1. Do not intercept the Android back button unless we're at the top of the navigation stack.
2. Remember what section (Overview/Saved) and tab (Posts/Comments) we're on across refreshes, tab switches, and account changes. This in particular is pretty frustrating when I go to Saved>Comments, for example, then refresh, it boots me back to Overview>Posts.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/f076beb2-6c79-46b7-a502-b95a3e109298

## Checklist

### DIFF
--- a/lib/user/pages/user_page_success.dart
+++ b/lib/user/pages/user_page_success.dart
@@ -15,6 +15,7 @@ import 'package:thunder/core/singletons/preferences.dart';
 import 'package:thunder/post/bloc/post_bloc.dart' as post_bloc;
 import 'package:thunder/post/utils/comment_action_helpers.dart';
 import 'package:thunder/shared/comment_reference.dart';
+import 'package:thunder/shared/primitive_wrapper.dart';
 import 'package:thunder/shared/snackbar.dart';
 import 'package:thunder/user/widgets/user_header.dart';
 import 'package:thunder/core/models/comment_view_tree.dart';
@@ -47,6 +48,9 @@ class UserPageSuccess extends StatefulWidget {
   final bool hasReachedPostEnd;
   final bool hasReachedSavedPostEnd;
 
+  final List<bool>? selectedUserOption;
+  final PrimitiveWrapper<bool>? savedToggle;
+
   const UserPageSuccess({
     super.key,
     required this.userId,
@@ -60,6 +64,8 @@ class UserPageSuccess extends StatefulWidget {
     required this.hasReachedPostEnd,
     required this.hasReachedSavedPostEnd,
     this.blockedPerson,
+    this.selectedUserOption,
+    this.savedToggle,
   });
 
   @override
@@ -71,10 +77,11 @@ class _UserPageSuccessState extends State<UserPageSuccess> with TickerProviderSt
   final _scrollController = ScrollController(initialScrollOffset: 0);
   bool hasScrolledToBottom = true;
 
-  // Keep these values static so that they persist between refreshes, tab changes, account changes, etc.
-  static int selectedUserOption = 0;
-  static List<bool> _selectedUserOption = <bool>[true, false];
-  static bool savedToggle = false;
+  int? selectedUserOption;
+  // Do not change the object reference of either of these.
+  // Only change members.
+  List<bool>? _selectedUserOption;
+  PrimitiveWrapper<bool>? savedToggle;
 
   @override
   void initState() {
@@ -98,6 +105,10 @@ class _UserPageSuccessState extends State<UserPageSuccess> with TickerProviderSt
 
   @override
   Widget build(BuildContext context) {
+    selectedUserOption ??= widget.selectedUserOption?.indexOf(true) ?? 0;
+    _selectedUserOption ??= widget.selectedUserOption ?? [true, false];
+    savedToggle ??= widget.savedToggle ?? PrimitiveWrapper<bool>(false);
+
     final theme = Theme.of(context);
     final DateTime now = DateTime.now().toUtc();
     final int? currentUserId = context.read<AuthBloc>().state.account?.userId;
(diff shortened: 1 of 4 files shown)


lib/user/pages/user_page_success.dart
```diff
@@ -71,16 +71,14 @@ class _UserPageSuccessState extends State<UserPageSuccess> with TickerProviderSt
   final _scrollController = ScrollController(initialScrollOffset: 0);
   bool hasScrolledToBottom = true;
 
-  int selectedUserOption = 0;
-  List<bool> _selectedUserOption = <bool>[true, false];
-  bool savedToggle = false;
+  // Keep these values static so that they persist between refreshes, tab changes, account changes, etc.
+  static int selectedUserOption = 0;
+  static List<bool> _selectedUserOption = <bool>[true, false];
+  static bool savedToggle = false;
```
hjiangsu, line 77: Hmm, I'm not sure what the conventions are for setting static values within a `StatefulWidget`.

Just curious, but how does this work then when we navigate to a user's page (rather than our own). Does it reset the state properly, or does it keep the same state as the ones assigned here since they're static?

One test I can think of is to go to your account profile, and navigate to the saved posts tab. Then go back to the feed and open a user's page and see what the result is there

micahmo: > One alternative way I can think of to do this is to use `AutomaticKeepAliveClientMixin` on the `AccountPage` widget.

Good idea! I didn't have a lot of luck with `AutomaticKeepAliveClientMixin` last time I tried to use it, but I'll give it another shot.

micahmo: Alright, it looks like this worked! I had to use a trick to wrap `bool` in a class, so that `UserPageSuccess` can change the same object instance that `AccountPage` is holding on to. But otherwise, it seems good! To test, I've demonstrated the same functionality as the original demo video, plus navigating to a different user page to show that the state is not applied in that case.


https://github.com/thunder-app/thunder/assets/7417301/fba86c50-bbcf-4bc9-b8be-30efedfe1742

hjiangsu: > I had to use a trick to wrap bool in a class, so that UserPageSuccess can change the same object instance that AccountPage is holding on to. 

Interesting, I wonder why that would be the case (maybe its a bug with Flutter itself?) Anyways, I think it looks good now so I'll go ahead and merge it in!

hjiangsu (approved):

lib/user/pages/user_page_success.dart
micahmo: Not a bug, just a difference between value and reference types. When the parent widget passes a primitive value down to the child, it's getting a copy of the object. (In other words, a new stack allocation, instead of a pointer to the same object on the heap.) So when the child changes the value, the parent has no idea. By wrapping it in a full object, both the parent and child now have a pointer to the same heap allocation, so any value changes can be seen in both places. That's also why I added a comment in the child widget to not change the reference! If we point to a new object, we'll again lose track of those changes in the parent.

Other languages solve this by allowing value types to be passed by reference, so that anyone can get a pointer to the same exact same object, even if it's a primitive value. But Dart has no such feature. 😞 

https://stackoverflow.com/a/18273525/4206279

hjiangsu: Ahh okay, I think I was misunderstanding what the issue was then (I thought it had something to do with `AutomaticKeepAliveClientMixin` not restoring state properly for primitive values)!

We don't have to change the logic since its already merged in but here's a potential alternative!

I think something that we can do in this case (if we want a child widget to update the state of a parent widget) is to use callback functions. Since `AccountPage` now handles `selectedUserOption` and `savedToggle`, we can pass in a new callback function `onChange(List<bool> updatedSelectedUserOption, bool updatedSavedToggle)` as a argument to `UserPageSuccess`.

Then anytime we need to change a value from `UserPageSuccess`, we can call the callback function which triggers a state update on `AccountPage` and rebuilds the `UserPageSuccess` with the updated values. This might work (although I haven't tried it).

Example:
```dart
UserPage(
  userId: accountState.personView!.person.id,
  isAccountUser: true,
  selectedUserOption: selectedUserOption,
  savedToggle: savedToggle,
  onChange: (updatedSelectedUserOption, updatedSavedToggle) {
    setState(() => {
        selectedUserOption = updatedSelectedUserOption;
        savedToggle = updatedSavedToggle;
    });
  }
)

micahmo: That sounds like a reasonable approach and something to keep in mind. I'll just mention two more things, and then we can probably move on to other things. 😊
* The child has to update its own state with these values, so it would have to both invoke the callback, and also do a `setState` for itself. This is true for `UserPageSuccess`es that are _not_ children of `AccountPage`, so we can't rely on our parent to rebuild us.
* As a result, the parent doesn't really need to be rebuilt. It would just need to keep track of the values. Its callback could store the values without a `setState`.

Because of these two things, it made more sense to me for the child to have a reference to the same object as the parent, where the parent would be able to hold on to (but not react to) state changes automatically. However, to your point, it's a bit non-obvious what's actually going in, and callbacks are a more standard mechanism for this kind of coordination.

hjiangsu: > This is true for UserPageSuccesses that are not children of AccountPage, so we can't rely on our parent to rebuild us.

Good point, it might just cause unnecessary re-builds in this case since the `UserPageSuccess` also has its own state!

> Because of these two things, it made more sense to me for the child to have a reference to the same object as the parent, where the parent would be able to hold on to (but not react to) state changes automatically.

Thats understandable! I think either method works in this specific scenario (I do like the trick that you did with wrapping a primitive value within a class to make it passable by ref 😄)

PR #966, opened 2023-12-07
Feature/settings account

micahmo:
## Pull Request Description

This PR adds access to the user account settings from the main settings page. Since they are not necessarily obvious from the profile page, and since this is where the rest of the settings live, I think it makes sense to have a link here too. The page does make it clear that these particular settings sync to the Lemmy account. There is also handling for when the user isn't logged in.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/0b7be227-1a99-4d9f-8a92-9586b5dc1af6

## Checklist


hjiangsu (approved):
LGTM!

PR #970, opened 2023-12-08
Changed default feed FAB action to New Post

hjiangsu:
## Pull Request Description

This PR changes the default feed FAB action to be New Post rather than dismiss read. Now that we can select a community to create a post from anywhere, I think it makes sense to make this the new default!

I think (correct me if I'm wrong) most other apps with FABs also default to the New Post action. This will make it easier for users who are more familiar with other apps to quickly find the action

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/settings/pages/fab_settings_page.dart
+++ b/lib/settings/pages/fab_settings_page.dart
@@ -60,7 +60,7 @@ class _FabSettingsPage extends State<FabSettingsPage> with TickerProviderStateMi
   bool postFabEnableSearch = true;
 
   /// The main single press action for the feed FAB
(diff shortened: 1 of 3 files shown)


hjiangsu:
> The fact that it's customizable (including to show the extended actions by default) means it doesn't matter a ton!

I think the main thing here is that some users may not be fully aware that the actions are customizable (we can probably figure out some way to guide the user to choose the default action somehow when they first install Thunder).

Because of this, they expect the FAB to either open the extended actions (like you mentioned above) or create a new post.

Issue #526, opened 2023-08-02
Add a "Favorites" category in sidebar


Ginjutsu:
**Is your feature request related to a problem? Please describe.**
One of the things that frustrates me with the current design is how when I want to go to a specific community, I have to scroll through a list of every community I'm subscribed to in order to find it. I'm personally subscribed to a _lot_ of communities, so this can be kind of an annoying process.

**Describe the solution you'd like**
Allow users to "favorite" communities that we frequent more often, which would move communities in the "Subscriptions" drop-down category to an additional "Favorites" category located above.

**Describe alternatives you've considered**
The search function can be a quicker way to find communities currently, but there are many situations where I'd rather just use the sidebar.

Coehill:
This is the #1 feature I've been missing from when I used RedReader for Reddit. Except, RedReader calls it "pinned".

To take it one step further, there could be a feed option for only your "favourited" Lemmy communities!

orangeNgreen:
Just wanted to add my support for this feature. Seems most of the communities I like to visit are located toward the bottom of my very long Subscriptions list (such as Thunder and Thunder Nightlies).

hjiangsu:
This is now in WIP! The main functionality is complete, but its still a bit rough so more testing is needed before this is introduced in a nightly build

hjiangsu:
Update: added a video to demonstrate the feature. Let me know what you think of it and if there are any changes that should be made before it's merged in. The video is attached to the PR: #971

orangeNgreen:
For some reason I’m unable to view the video. Nothing happens when I click play.

hjiangsu:
Interesting - are you on mobile? or on desktop? Sometimes, I have issues playing the video on mobile.

orangeNgreen:
Ah. I was on mobile. Worked on desktop, and I love it! Thanks!

Coehill:
Looks awesome! Thanks for implementing this feature.

orangeNgreen:
I ask this only out of excitement: any thoughts when the feature will hit a Nightly?

hjiangsu:
I'll try to get a nightly out within the next couple of days, just doing some final internal testing on things first!

hjiangsu:
A new nightly build has been released which contains this feature: https://github.com/thunder-app/thunder/releases/tag/0.2.7-5

Let me know if you encounter any bugs or issues with it!

PR #971, opened 2023-12-08
Added ability to favorite communities in drawer

hjiangsu:
## Pull Request Description

This PR adds the ability to favorite subscribed communities. When a community is favorited, it will show up in a new section on the drawer for easier access.

Some notes:
- Refactored the drawer to be more modular, and to follow existing conventions better
- Fixed the issue where tapping on the user indicator in the drawer did not navigate to the user page
- Favourite communities are only available when the user is logged in. It is disabled for anonymous users for the time being
- I opted to use the database rather than SharedPreferences since the data is slightly more complex (account id, community id). The downside here is that we don't have a method to restore or backup database files, so favourites would be lost on re-installs.

Some more testing would be ideal for the database migrations (adding in a new table for favorites) from 0.2.6 -> 0.2.7, and also for newly created databases.

## Issue Being Fixed

Issue Number: #526

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/8fd6131f-fe17-41ed-999d-9b9f26e35aa6

## Checklist


hjiangsu:
@micahmo Feel free to test this out - especially for database migrations to make sure things go smoothly from last version to this version

hjiangsu:
> Does this work with anonymous subscriptions?

> Favourite communities are only available when the user is logged in. It is disabled for anonymous users for the time being

I mentioned it on the PR description but nope, this only works with logged in users for now! I would need to add in some more logic to make it work with anonymous accounts, which could be an improvement for the future.

> There's a funny UI thing where an inkwell effect occurs on a different community after favoriting.

Interesting, I'll take a look at this! I didn't notice it but I'll see if I can reproduce it

> I am currently unable to test this feature because I get the following error.

Yup, I think that might be the case, I did some db changes in between so it might've messed things up. I think clearing db and relaunching it should do the trick!

micahmo:
> I mentioned it on the PR description but nope, this only works with logged in users for now! I would need to add in some more logic to make it work with anonymous accounts, which could be an improvement for the future.

Sorry I missed that! Makes sense though!

> Interesting, I'll take a look at this! I didn't notice it but I'll see if I can reproduce it

I can see it in your video! Look at the first time you favorite a community. After you press the star, the start next to the top unfavorited community blinks. Again it's something I've seen with Flutter before. I think it's related to it getting confused about which widget was activated after the UI rebuilds.

> Yup, I think that might be the case, I did some db changes in between so it might've messed things up. I think clearing db and relaunching it should do the trick!

Will do!

micahmo:
> I think clearing db and relaunching it should do the trick!

This works fine once, but it looks like we can't switch to an older version and switch back. I don't know if there's a way to conditionally create the table if it doesn't exist. If not, we'll just have to know that we don't support downgrading from and then upgrading back to this version.

Otherwise the feature works great! 😊

hjiangsu:
> After you press the star, the star next to the top unfavorited community blinks.

Ahh, I wonder if this is just due to the animation delay of tapping the icon! Regardless, I think this is a minor thing and shouldn't be too much of an issue

> I don't know if there's a way to conditionally create the table if it doesn't exist.

I can do a try-catch on the creation logic so that it doesn't hang there! I think that should generally work in the cases where the table schema is left untouched (e.g., we don't change the schema between versions)

hjiangsu:
This should be good now! I added try-catch when creating or upgrading the database. Feel free to give it a quick test and let me know if it works!

Note that this only catches the error and proceeds forward. It doesn't attempt to fix any issues that it catches. This should allow the user to at least load Thunder, and delete the database from the Settings page to reset things back to default if needed!

micahmo (approved):
LGTM!

micahmo:
I just noticed that we lost the full community name on the second line of the listings in the drawer, so we can only see the display name by default. Was that intentional?

hjiangsu:
> Was that intentional?

It was intentional! I was playing around with the idea of hiding the full community name in the drawer now that we have a favorites action.

Because of the new action, the amount of horizontal real estate we have to display information is reduced. So to compensate for that, I thought I'd remove the full community name since it feels a bit redundant to have (we already have the community name). To add on to that, we can always see the full community name through the tooltip if its needed! Do you have any thoughts on this change?

PR #952, opened 2023-12-05
Update LPG dependency

micahmo:
## Pull Request Description

This PR just updates the LinkPreviewGenerator dependency to get the fix from thunder-app/link_preview_generator#4.

PR #972, opened 2023-12-10
Set new LPG property

micahmo:
## Pull Request Description

This is a super small PR which is a follow up #952 and https://github.com/thunder-app/link_preview_generator/pull/4. I had updated the LPG dependency, but I forgot to set the new property I added!


hjiangsu (approved):
LGTM!

PR #973, opened 2023-12-12
Standardize and make configurable full names

micahmo:
## Pull Request Description

Right now Thunder is a bit inconsistent in how fully qualified user and community names are displayed. In some places (mostly for communities) we use the bullet symbol, `·`, but we also occasionally use the `@` (e.g., input prompt). On the flip side, usernames are more often displayed with `@` but also sometimes use the bullet (e.g., search results). :blush:

Since people have probably gotten used to one way or the other, I didn't want to impose a single style. Instead, I've added settings for user and community name style, allowing the user to decide! The default values for these settings should keep things mostly like they were before, but there will be some subtle differences, since we weren't entirely consistent.

These are the places that I captured names being displayed. Let me know if I missed any!

### Usernames
 - [x] Feed posts (N/A - instance not yet shown here)
 - [x] Post body tooltip
 - [x] Comments (when enabled, and tooltip)
 - [x] Header
 - [x] Mod list in community sidebar
 - [x] Search result
 - [x] Search filter
 - [x] Input prompt
 - [x] Blocks
 - [x] User indicator (e.g., new post, account settings)

### Communities
 - [x] Drawer (and tooltip)
 - [x] Feed posts
 - [x] Post body tooltip
 - [x] Header
 - [x] Modded community list in user sidebar
 - [x] Search result
 - [x] Search filter
 - [x] Input prompt
 - [x] Blocks
 - [x] Comment reference (e.g., inbox)
 - [x] Cross-posts

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Note that this demo only shows a couple of places with interesting changes. See the list above for more details!

https://github.com/thunder-app/thunder/assets/7417301/43bf1bcf-2ecf-4616-8cc7-977d5b12adf7

## Checklist

### DIFF
--- a/lib/core/enums/full_name_separator.dart
+++ b/lib/core/enums/full_name_separator.dart
@@ -13,36 +13,24 @@ enum FullNameSeparator {
 
 String generateUserFullName(BuildContext context, name, instance) {
   final ThunderState thunderState = context.read<ThunderBloc>().state;
-  if (thunderState.userSeparator == FullNameSeparator.dot) {
-    return '$name · $instance';
-  } else if (thunderState.userSeparator == FullNameSeparator.at) {
-    return '$name@$instance';
-  }
-
-  // Default format in case the setting isn't set
-  return '$name@$instance';
+  return switch (thunderState.userSeparator) {
+    FullNameSeparator.dot => '$name · $instance',
+    FullNameSeparator.at => '$name@$instance',
+  };
 }
 
 String generateUserFullNameSuffix(BuildContext context, instance) {
   final ThunderState thunderState = context.read<ThunderBloc>().state;
-  if (thunderState.userSeparator == FullNameSeparator.dot) {
-    return ' · $instance';
-  } else if (thunderState.userSeparator == FullNameSeparator.at) {
-    return '@$instance';
-  }
-
-  // Default format in case the setting isn't set
-  return '@$instance';
+  return switch (thunderState.userSeparator) {
+    FullNameSeparator.dot => ' · $instance',
+    FullNameSeparator.at => '@$instance',
+  };
 }
 
 String generateCommunityFullName(BuildContext context, name, instance) {
   final ThunderState thunderState = context.read<ThunderBloc>().state;
-  if (thunderState.communitySeparator == FullNameSeparator.dot) {
-    return '$name · $instance';
-  } else if (thunderState.communitySeparator == FullNameSeparator.at) {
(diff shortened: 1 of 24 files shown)


lib/core/enums/full_name_separator.dart
```diff
@@ -0,0 +1,48 @@
...
+  at('name@instance.tld');
+
+  final String label;
+
+  const FullNameSeparator(this.label);
+}
+
+String generateUserFullName(BuildContext context, name, instance) {
+  final ThunderState thunderState = context.read<ThunderBloc>().state;
+  if (thunderState.userSeparator == FullNameSeparator.dot) {
```
hjiangsu, line 16: Just a stylistic syntactic thing, but I feel like using a `switch` here makes things a bit cleaner. Thoughts?

```dart
switch (thunderState.userSeparator) {
  case FullNameSeparator.dot:
    return '$name · $instance';
  case FullNameSeparator.at:
    return '$name@$instance';
  default:
    return '$name@$instance';
}
```

lib/core/enums/local_settings.dart
```diff
@@ -41,6 +41,10 @@ enum LocalSettings {
   useAdvancedShareSheet(name: 'setting_use_advanced_share_sheet', label: 'Use Advanced Share Sheet'),
   showCrossPosts(name: 'setting_show_cross_posts', label: 'Show Cross-Posts'),
 
+  // Advanced Settings
+  userFormat(name: 'user_format', label: 'User Format'),
```
hjiangsu, line 45: If we're using `l10n` for localization, I think we can leave the `label` blank! That'll help a bit when we eventually deprecate the `label` field since it no longer applicable

hjiangsu (approved):
Just a couple of minor comments! Otherwise, LGTM!

lib/core/enums/full_name_separator.dart
micahmo: Good idea! In fact, I think I'll take it a step further and use a switch expression instead of statement.

lib/core/enums/local_settings.dart
micahmo: Done!

lib/core/enums/full_name_separator.dart
hjiangsu: > I think I'll take it a step further and use a switch expression instead of statement.

Even better, I didn't even know that was a thing 😅

PR #978, opened 2023-12-13
Fix/user spinner

micahmo:
## Pull Request Description

This is a super small PR which handles the case where a user has no posts.


hjiangsu (approved):
LGTM!

PR #979, opened 2023-12-13
Feature/mark all as read confirmation

micahmo:
## Pull Request Description

This PR adds a confirmation dialog before marking all messages as read. Since this could affect a large number of messages and since it may not be obvious what the button does, I figured it would be good to confirm. Even with the additonal tap required, it's still faster than manually marking messages.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/8e294ef5-56fd-4ea0-9c2d-487e113b3995

## Checklist


lib/inbox/pages/inbox_page.dart
```diff
@@ -48,26 +49,46 @@ class _InboxPageState extends State<InboxPage> {
...
-              semanticLabel: AppLocalizations.of(context)!.readAll,
+              semanticLabel: l10n.readAll,
             ),
-            onPressed: () {
-              context.read<InboxBloc>().add(MarkAllAsReadEvent());
+            onPressed: () async {
+              bool result = false;
+              await showThunderDialog<bool>(
+                context: context,
+                title: l10n.confirmMarkAllAsReadTitle,
```
hjiangsu, line 69: I think the title/headline "Mark All As Read?" should suffice since the dialog itself acts as the confirmation modal (although I did notice that the log out modal also has "Confirm", so maybe we can change the "Confirm Log Out" to just "Log Out?" to make it consistent)

It also shortens the title and makes it a single line which looks better imo! (https://m3.material.io/components/dialogs/guidelines also mostly shows single lined headline)

hjiangsu (approved):
Just a few minor comments!

lib/inbox/pages/inbox_page.dart
```diff
@@ -48,26 +49,46 @@ class _InboxPageState extends State<InboxPage> {
...
+                },
+                secondaryButtonText: l10n.cancel,
+                onPrimaryButtonPressed: (dialogContext, _) {
+                  result = true;
+                  Navigator.of(dialogContext).pop();
+                },
+                primaryButtonText: l10n.markAllAsRead,
+              );
+
+              if (result && context.mounted) {
```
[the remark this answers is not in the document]
micahmo: Good idea! Done!

lib/inbox/pages/inbox_page.dart
micahmo: Sure thing, I updated both.

| image |
| - |

hjiangsu:
Looks good now, thanks!

PR #982, opened 2023-12-13
Add ability to set app font

micahmo:
## Pull Request Description

This PR adds the ability to change the font used by Thunder across the app. It should respect all of the existing text styling, colors, and sizing.

I picked a few fonts from the most popular on Google Fonts, but of course we can include others that people may want.

@hjiangsu, Can you help to test iOS? Specifically can you check that the System theme looks right, both in the selection preview and in the app when selected? And of course anything else you want to test. :)

One lingering thing is that the custom fonts always feel slightly larger than the system font. I'm not sure if this is my mind playing tricks on me, or whether these fonts just have larger characters in general, or if there's something funky going on with our text scaling (maybe we could lower the default text scaling if a custom font is used). I'd be curious to see if there's a similar effect on iOS.

P.S. For the sake of reviewing, you can ignore things under the new `google_fonts` folder.

---

Tasks
 - [x] Show preview in selection modal
 - [x] Download fonts and build with app ~so [google_fonts](https://pub.dev/packages/google_fonts) doesn't try to scrape them from the web~
 - [x] Fix any formatting issues (e.g., bold usernames on comments)
 - [x] Add more fonts
 - [x] Localize new strings

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/16761cfe-f79e-45f3-a279-9393c9c5bb2c


hjiangsu:
This is pretty cool, thanks for working on this! I'll definitely test it out on iOS and let you know how it performs there.

hjiangsu:
So I did a quick test on this, and it seems like everything works!
- I feel like the system font has been unchanged, and if it has been, its hard to tell for me
- I can see the previews of the fonts in the bottom sheet
- Switching fonts seem to work as expected
- I don't really notice that much difference in terms of the size of the text. It might be slightly larger but it's hard to tell

micahmo:
If it looks good to you then I'd say it's ready for review/merge!

hjiangsu:
Hmm, so I just ran an app build for this, and this increases the build size by quite a bit. It goes from 29.6MB -> 40MB. This is most likely because of the font assets. I wonder if theres a way to reduce that, or limit the choices down a bit more?

We can eventually add more fonts to be downloaded on-demand rather than pre-packaged!

micahmo:
The [`google_fonts`](https://pub.dev/packages/google_fonts) package gets around this by downloading all fonts on-demand. However, I wasn't sure how people would feel about Thunder frequently reaching out to Google Fonts (using data, talking to Google, etc., etc.)! In addition, it doesn't have good support for font weights when applied via themes, which led me away from it.

https://github.com/material-foundation/flutter-packages/issues/35

We can shelve this for now, as I don't think the demand is very high, but I would be curious to know if there is an easy way to dynamically load local fonts!

micahmo:
@hjiangsu Regarding this PR, how would you feel if we let the users download fonts at runtime to save on app space? We could give a very clear warning that the app would be reaching out to Google servers, which may be undesirable for some.

Unfortunately, while the [`google_fonts`](https://pub.dev/packages/google_fonts) package supports this functionality internally, I don't think we would have a great opportunity to insert our own warning message at the right time. Additionally and more importantly, when using fonts in this way, weights are not correctly supported (see [here](https://github.com/material-foundation/flutter-packages/issues/35)). If we went with the solution proposed, we would have to build in some mechanism whereby we manually download and load the font file, which we'd have to figure out how to do in Flutter. But I just wanted to check if it sounds reasonable in principle.

If not, we can just abandon this change since it's probably not in high demand anyway.

hjiangsu:
Sorry for the delay in response (completely missed it)!

I think a middle ground here could be to reduce the number of options down, and slowly add more if there is demand for others! I believe there are currently eight fonts introduced - maybe we can narrow it down to 4-5?

micahmo:
> maybe we can narrow it down to 4-5?

Ok, that's a good compromise! I'll try to pick a few good ones (maybe one from each category of sans serif, serif, monospace, etc.). We'll see how the app size looks after that.

micahmo:
Alright I have whittled down the list of fonts, in addition to `System`, to these few, with one representative from each "category".
* Poppins (geometric sans serif)
* Roboto Slab (slab serif)
* Noto Sans (alternative to system font for sans serif)
* Noto Serif (serif)
* Noto Sans mono (Monospace)

If this is still too big, I can take out a couple more!

hjiangsu:
Hmm, I just tried to rebuild it with the changes, and the general app size is still roughly the same 🙃. I took a look at the diff, and it seems like the major culprits come from the Noto family (~500+ KB per `.ttf` file)

If this is the case, then maybe we can just go ahead and merge this in (irregardless of the app build size). I _believe_ that most users would likely download updates when on WiFi, so data might not be that big of an issue. My initial worry was with regards to data savings but considering that Lemmy itself is quite media/data heavy, this might not be a concern. I would still like to hear your thoughts on this though!

micahmo:
> the general app size is still roughly the same 🙃

Haha! I should've checked the size of the ones I was deleting vs. the ones I kept.

You make a good point about app size, and I have heard people complain that a seemingly "simple" app has a large size, even though it barely affects their storage or data usage, so I am still a bit torn on this. Add that to the fact that nobody's really asking for this feature, and I do hesitate to push it through. Would it be worth soliciting feedback in Matrix (since that's where we'd be most likely to reach a good number of interested users)?

hjiangsu:
Yeah I think that sounds like a good idea!

hjiangsu:
Update: One consideration here is also the fact that the IzzyOnDroid repo has a max size limit of 30MB, which adding this would cause it to go over by quite a bit (~10MB over). Unfortunately, I think we'll have to reconsider this and possibly limit it to on-demand downloads, or find ways to reduce the overall build size.

image


Reference: https://gitlab.com/IzzyOnDroid/repo/#what-are-the-requirements-an-app-must-meet-to-be-included-with-the-repo

micahmo:
Good call! I will potentially look into how we can load additional fonts on-demand!

micahmo:
Hey all, I decided to take another look at this PR, and I'm afraid we may have hit a wall. The problem lies in the fact that, if you don't bundle define your fonts in `pubspec.yaml`, then there's no way to associate fonts of different weights with the same font family. I experimented some more with `google_fonts` and also [`dynamic_fonts`](https://pub.dev/packages/dynamic_fonts), but they all suffer from material-foundation/flutter-packages#35 where, if you use any technique other than setting the `fontFamily`, such as using the `textStyle` or `textTheme`, it will only load one variant of the font.

If anyone else would like to give this a shot, please feel free! But otherwise I'll plan on closing this.

micahmo:
I'm going to close this one out. It turned out to be more trouble than it was worth, and the longer it sits here, the more it generates conflicts and clutters things. 😊

PR #984, opened 2023-12-15
Utilize favorites feature elsewhere in app

micahmo:
## Pull Request Description

This PR adds a few enhancements on top of the recent [favorite communities feature](https://github.com/thunder-app/thunder/pull/971).

1. The ability to add/remove favorites from the community feed page.

https://github.com/thunder-app/thunder/assets/7417301/89c6e715-15c0-4af4-8104-33bc52e293eb

2. In community prompts, indicate if a community is favorited. Also prioritize favorites at the top of the suggestions list.

https://github.com/thunder-app/thunder/assets/7417301/93ceb72b-fe31-45bf-8f2e-f92fa049c44f

3. On the search page, show favorite communities.

https://github.com/thunder-app/thunder/assets/7417301/a35916ad-b3f1-4d48-91fd-26524277a841

4. In search results, indicate if a community is favorited. Also prioritize favorites at the top of the results.

https://github.com/thunder-app/thunder/assets/7417301/e5a1dc7b-d414-496c-abe4-71d37e7f967c

**EDIT:** I added another commit which prioritizes favorites in the list of communities shown in the input dialog before the user has typed anything.


hjiangsu:
This is awesome, thanks for doing this! It'll really help bring some QoL improvements to the favourites action!

I do have one minor discussion point around UI/UX. This is with regards to the favorites button being on the app bar for a community. I think Material design generally only recommends having at most 3 actions on the app bar. For more than 3 actions, we would use a dropdown menu to show the additional actions.

This is the guidelines: https://m3.material.io/components/top-app-bar/guidelines#224e8ca0-4177-45f4-9f4f-eef1fbb0e4b2

Now, I don't know if we necessarily need to strictly follow their guidelines, which is why I want to bring up this discussion around this. If we do want to follow their guidelines, then we need to think about which actions should be placed in the dropdown menu! Otherwise, we could leave this for a future discussion (when we eventually have more actions that need to be placed at the top)

micahmo:
> I think Material design generally only recommends having at most 3 actions on the app bar. For more than 3 actions, we would use a dropdown menu to show the additional actions.

That's a good point, especially since it would truncate the community name even further. What do you think about putting Favorite (since it's new) and Refresh (since it's available via FAB) in the overflow?

hjiangsu:
> That's a good point, especially since it would truncate the community name even further. What do you think about putting Favorite (since it's new) and Refresh (since it's available via FAB) in the overflow?

Yeah, I think that sounds good!

micahmo:
> Yeah, I think that sounds good!

Great, should be all set now! On the main feed, refresh and sort are visible. In a community, Subscribe and sort are visible, while refresh and favorite are in the overflow!

https://github.com/thunder-app/thunder/assets/7417301/6bac4efa-ef80-4793-b2f9-28a497fcf2a0

hjiangsu (approved):
LGTM!

PR #985, opened 2023-12-16
LPG fix

micahmo:
This is a super small followup to #972 and #952. Aside from setting the new property, the `onTap` must be null to prevent the inkwell. Didn't think it would take three PRs to get this changed. 😆


hjiangsu (approved):
LGTM!

PR #987, opened 2023-12-16
Fix share preview

micahmo:
This is a small PR which fixes an issue where the image preview in the advanced share sheet would always show "Unable to load image". The problem was that we would return the error text regardless of the state of the image viewer. However, this was only a problem for image viewers displaying bytes, not ones loading an image from a URL.


hjiangsu (approved):
LGTM!

PR #989, opened 2023-12-18
Fix the color of the delete draft icon

micahmo:
This PR fixes the color of the delete draft icon to look better in light mode (previously it was way too dark).

|  | Before | After |
|--------|--------|--------|
| **Light** | image | image |
| **Dark** | image | image |


hjiangsu (approved):
LGTM!

PR #990, opened 2023-12-18
Fix too wide images in sidebars

micahmo:
This PR fixes an issue where images may be too wide in the sidebar. In #939 we removed the artificially enforced minimum size, which is still good, but now we also need an (optional) maximum size for constrained-width areas like the sidebar.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/lib/community/widgets/community_sidebar.dart
+++ b/lib/community/widgets/community_sidebar.dart
@@ -117,7 +117,11 @@ class _CommunitySidebarState extends State<CommunitySidebar> {
                       child: ListView(
                         shrinkWrap: true,
                         children: [
-                          CommonMarkdownBody(body: communityView.community.description ?? ''),
+                          CommonMarkdownBody(
+                            body: communityView.community.description ?? '',
+                            // The sidebar is 0.8 the width of the screen, so we'll be just slightly smaller than that
+                            imageMaxWidth: 0.7 * MediaQuery.of(context).size.width,
+                          ),
                           const SidebarSectionHeader(value: "Stats"),
                           Padding(
                             padding: const EdgeInsets.symmetric(horizontal: 8.0),
--- a/lib/community/widgets/community_sidebar.dart
+++ b/lib/community/widgets/community_sidebar.dart
@@ -31,6 +31,8 @@ import 'package:thunder/utils/date_time.dart';
 import 'package:thunder/utils/instance.dart';
 import 'package:thunder/utils/navigate_user.dart';
 
+const kSidebarWidthFactor = 0.8;
+
 class CommunitySidebar extends StatefulWidget {
   final GetCommunityResponse? getCommunityResponse;
   final Function onDismiss;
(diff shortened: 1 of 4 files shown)


lib/shared/image_preview.dart
```diff
@@ -13,6 +13,7 @@ class ImagePreview extends StatefulWidget {
   final bool nsfw;
   final double? height;
   final double? width;
+  final double? maxWdith;
```
hjiangsu, line 16: Slight typo here - should be `maxWidth`

lib/community/widgets/community_sidebar.dart
```diff
@@ -117,7 +117,11 @@ class _CommunitySidebarState extends State<CommunitySidebar> {
                       child: ListView(
                         shrinkWrap: true,
                         children: [
-                          CommonMarkdownBody(body: communityView.community.description ?? ''),
+                          CommonMarkdownBody(
+                            body: communityView.community.description ?? '',
+                            // The sidebar is 0.8 the width of the screen, so we'll be just slightly smaller than that
+                            imageMaxWidth: 0.7 * MediaQuery.of(context).size.width,
```
hjiangsu, line 123: I think a slightly better approach here would be to extract the sidebar width constant so that it can be used here. The container width is determined by `FractionallySizedBox`'s `widthFactor` parameter on line `77`. We can create a constant for that value and use it here.

This will reduce overhead in having to adjust two different values (`widthFactor` and `imageMaxWidth`) if we ever want to change the width of the sidebar! Thoughts?

```dart
const kSidebarWidthFactor = 0.8;

...

FractionallySizedBox(widthFactor: kSidebarWidthFactor)

...
CommonMarkdownBody(
    body: communityView.community.description ?? '',
    imageMaxWidth: (kSidebarWidthFactor * MediaQuery.of(context).size.width) - somePadding,
),

```

hjiangsu (commented):
Just a few comments!

lib/community/widgets/community_sidebar.dart
micahmo: Good idea, done!

lib/user/widgets/user_sidebar.dart
```diff
@@ -232,6 +232,8 @@ class _UserSidebarState extends State<UserSidebar> {
                                 ),
                                 child: CommonMarkdownBody(
                                   body: widget.userInfo?.person.bio ?? 'Nothing here. This user has not written a bio.',
+                                  // The sidebar is 0.8 the width of the screen, so we'll be just slightly smaller than that
+                                  imageMaxWidth: 0.7 * MediaQuery.of(context).size.width,
```
hjiangsu, line 236: The same should be applied here as `community_sidebar`!

micahmo: All set!

hjiangsu (approved):
LGTM! :D

PR #991, opened 2023-12-18
Highlight featured local posts

micahmo:
This PR highlights posts that are "featured locally" in the same way that we currently highlight posts which are "featured in a community". I believe this now matches the behavior of the web UI and, in fact, I think it's even more to feature local posts (which are sort of global across communities for a given instance).


hjiangsu (approved):
LGTM!

PR #994, opened 2023-12-18
Feature/condensed post body

micahmo:
## Pull Request Description

This PR introduces a new view type for post bodies. Inspired by Relay, this allows the post body to be condensed, meaning that any link or image previews will remain as a small thumbnail on the side. Text posts are still fully expanded (and collapsible).

The use case here is that you've already viewed the image/link from the main feed, so now you want to jump right to the comments. Seeing the full image again is often unnecessary.

> [!NOTE]
> I removed the duplicate "Show Cross Posts" setting from the General page, and made a new section on the Appearance > Posts page for post body settings.

> P.S. If there is a better name than "Card View" for the default post body appearance, let me know!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/548c73b2-a085-40dd-8ba5-928d8bf3ea08

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -28,7 +28,7 @@ enum LocalSettings {
   showThumbnailPreviewOnRight(name: 'setting_compact_show_thumbnail_on_right', label: 'Thumbnails on the Right'),
   showTextPostIndicator(name: 'setting_compact_show_text_post_indicator', label: 'Show Text Post Indicator'),
   tappableAuthorCommunity(name: 'setting_compact_tappable_author_community', label: 'Tappable Authors & Communities'),
-  useCompactPostBodyView(name: 'setting_general_use_compact_post_body_view', label: ''),
+  postBodyViewType(name: 'setting_general_post_body_view_type', label: ''),
 
   // General Settings
   showPostVoteActions(name: 'setting_general_show_vote_actions', label: 'Show Vote Buttons'),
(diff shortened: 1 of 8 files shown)


hjiangsu:
Thanks for this, this is really cool!

I just have one discussion about architecture for this. Since we're starting to add new views to the post body, I'm thinking this is a good chance to make the setting as adaptable as possible to any potential future post view types. Instead of using `useCompactPostBodyView`, what are your thoughts on changing it to something more flexible like `postBodyViewType`?

Then, we can store an enum value (e.g., `PostBodyViewType.condensed`, or `PostBodyViewType.mediaPreview`) and use that to build this new view. The benefit here is that we are no longer restricted to using a boolean, so we can add in more variations of views to the post body (e.g., default could be `PostBodyViewType.expanded`, a future one could be something like `PostBodyViewType.gallery`, etc.).

I know I used `useCompactView` for the feed (which admittedly was not the best way to do it in hindsight 😅) which is why I wanted to bring it up!

micahmo:
> Instead of using `useCompactPostBodyView`, what are your thoughts on changing it to something more flexible like `postBodyViewType`?

Great idea! All done!

https://github.com/thunder-app/thunder/assets/7417301/e0162f03-fd68-47e4-a9b1-b67247d6bf7b

micahmo:
Sorry I slightly misunderstood your comment and I used `condensed` and `mediaPreview` as the two options. I just pushed another change to make the options `condensed` and `expanded`!

hjiangsu (approved):
LGTM! Thanks for changing it :D

I'll hold off on merging this until I get the 0.2.7 general release out

PR #996, opened 2023-12-18
Add icons to popup menu

micahmo:
## Pull Request Description

This is a small PR which adds icons to the new feed popup menu. I also reorganized the code so that the tap handlers are more logically placed.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/cd1b4cfa-9cdb-48f1-ba9a-1b2a13e6a1b6

## Checklist

### DIFF
--- a/lib/feed/widgets/feed_page_app_bar.dart
+++ b/lib/feed/widgets/feed_page_app_bar.dart
@@ -114,12 +114,11 @@ class FeedPageAppBar extends StatelessWidget {
             itemBuilder: (context) => [
               PopupMenuItem(
                 onTap: () => triggerRefresh(context),
-                child: Row(
-                  children: [
-                    const Icon(Icons.refresh_rounded, size: 20),
-                    const SizedBox(width: 10),
-                    Text(l10n.refresh),
-                  ],
+                child: ListTile(
+                  dense: true,
+                  horizontalTitleGap: 5,
+                  leading: const Icon(Icons.refresh_rounded, size: 20),
+                  title: Text(l10n.refresh),
                 ),
               ),
               PopupMenuItem(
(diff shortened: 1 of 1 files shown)


lib/feed/widgets/feed_page_app_bar.dart
```diff
@@ -111,26 +111,30 @@ class FeedPageAppBar extends StatelessWidget {
...
-                  await toggleFavoriteCommunity(context, community, isFavorite);
-                  break;
-              }
-            },
             itemBuilder: (context) => [
               PopupMenuItem(
-                value: 'refresh',
-                child: Text(l10n.refresh),
+                onTap: () => triggerRefresh(context),
+                child: Row(
```
hjiangsu, line 117: Would it be better to use `ListTile` in this case if we want to add in icons to the text?

It seems like that is the recommended approach when reading the documentation for [PopupMenuItem](https://api.flutter.dev/flutter/material/PopupMenuItem-class.html):

> Typically the [child](https://api.flutter.dev/flutter/material/PopupMenuItem/child.html) of a [PopupMenuItem](https://api.flutter.dev/flutter/material/PopupMenuItem-class.html) is a [Text](https://api.flutter.dev/flutter/widgets/Text-class.html) widget. More elaborate menus with icons can use a [ListTile](https://api.flutter.dev/flutter/material/ListTile-class.html).

hjiangsu (commented):
Just one comment!

lib/feed/widgets/feed_page_app_bar.dart
micahmo: Sure thing, updated! I also modified some of the density properties of `ListTile` so that it takes up about the same amount of space as when using `Text`. (Otherwise, `ListTile` likes to be very big!)

hjiangsu (approved):
LGTM!

PR #997, opened 2023-12-19
Fix issue tapping on some usernames

micahmo:
## Pull Request Description

This PR fixes an issue where we couldn't navigate to some user pages because they were federated users from different platforms. For example, there is [`@remindme@mstdn.social`](https://lemmy.world/u/remindme@mstdn.social) which is a valid Lemmy user, but [`mstdn.social`](https://mstdn.social/) is not a Lemmy instance.

Previously we would check if the instance name was a valid Lemmy instance, so I removed that check. Note that this is still safe (in case we encounter an invalid user) because we already had a `try`/`catch` around `navigateToUserPage`. If it fails, we will still fall back to classic web navigation.

> This also speeds up the check, because `isLemmyInstance` would often attempt to make an API call to the instance.

> It's unclear whether we should also remove this check from `getLemmyCommunity`, since I'm unsure whether communities can be federated from non-Lemmy instances.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/56e9655f-0098-422c-a5c6-e8cbf8c331b3

## Checklist

### DIFF
--- a/lib/utils/instance.dart
+++ b/lib/utils/instance.dart
@@ -67,17 +67,17 @@ final RegExp username = RegExp(r'^@?(https?:\/\/)?((?:(?!\/u\/u).)*)@(.*)$');
 /// Otherwise, returns null.
 Future<String?> getLemmyUser(String text) async {
   final RegExpMatch? fullUsernameUrlMatch = fullUsernameUrl.firstMatch(text);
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM!

PR #998, opened 2023-12-19
Fix search post state issue

micahmo:
## Pull Request Description

This PR fixes an issue where certain post actions would not reflect in the search results list. This was a tricky one to track down, but it turned out to be a mismatch between the post lists held by the `SearchBloc` and `FeedBloc`. The `FeedBloc` would update its reference in `_onFeedItemUpdated`. Later, when it would index into the list and update an object in `_onFeedItemActioned`, the `SearchBloc` would not see that change, because it was holding a different list reference.

The fix was to modify `_onFeedItemUpdated` so that it doesn't return a new list, but returns a modified version of the same list.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/9ed968b8-57e8-4176-902c-ae72acbb3345

### After

https://github.com/thunder-app/thunder/assets/7417301/3f801c8f-376f-4b8b-b9ba-9c0a4a67ae9b

## Checklist

### DIFF
--- a/lib/feed/bloc/feed_bloc.dart
+++ b/lib/feed/bloc/feed_bloc.dart
@@ -231,15 +231,14 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
   Future<void> _onFeedItemUpdated(FeedItemUpdatedEvent event, Emitter<FeedState> emit) async {
     emit(state.copyWith(status: FeedStatus.fetching));
 
-    List<PostViewMedia> updatedPostViewMedias = state.postViewMedias.map((PostViewMedia postViewMedia) {
+    for (final (index, postViewMedia) in state.postViewMedias.indexed) {
       if (postViewMedia.postView.post.id == event.postViewMedia.postView.post.id) {
-        return event.postViewMedia;
-      } else {
-        return postViewMedia;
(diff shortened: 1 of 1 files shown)


lib/feed/bloc/feed_bloc.dart
```diff
@@ -231,15 +231,14 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
...
-        return event.postViewMedia;
-      } else {
-        return postViewMedia;
+        state.postViewMedias[index] = event.postViewMedia;
       }
-    }).toList();
+    }
 
-    emit(state.copyWith(status: FeedStatus.success, postViewMedias: updatedPostViewMedias));
+    //emit(state.copyWith(status: FeedStatus.success, postViewMedias: updatedPostViewMedias));
```
hjiangsu, line 240: We can remove this comment if its not being used!

hjiangsu (approved):
Thanks for this, I didn't even realize there was an issue with post actions on the search page. LGTM, just one comment!

lib/feed/bloc/feed_bloc.dart
micahmo: Oops, sorry I missed that! All set!

PR #999, opened 2023-12-20
Reset search for empty query

micahmo:
This is a super small PR which restores a previous functionality. When the search field is empty, the search should be reset (i.e., no results shown). This works when pressing the `x` but was broken when deleting the contents of the search field.


hjiangsu (approved):
LGTM!

Issue #219, opened 2023-07-04
Push notification


Tealk:
**Is your feature request related to a problem? Please describe.**
I would like to be informed when I receive responses.

**Describe the solution you'd like**
A push integration, preferably not from google but e.g. with [UnifiedPush](https://unifiedpush.org/)

trymeouteh:
Would like to see notifications work on degoogled devices.

shortwavesurfer2009:
This seems quite similar to #125

K4LCIFER:
> \[...\] with [UnifiedPush](https://unifiedpush.org/)

Agreed. This would require backend support from lemmy, however -- there is [an open issue](https://github.com/LemmyNet/lemmy/issues/2631) for this.

micahmo:
Hey all! An experimental version of notifications (for Android) is being worked on in #1000! It works completely without a notification server by having Thunder make a periodic poll. Unfortunately this won't work on iOS and has various limitations depending on Android manufacturer, but it appears like the solutions mentioned in LemmyNet/lemmy#2631 have similar issues, so this is at least a start on our part. 😊

K4LCIFER:
> Hey all! An experimental version of notifications (for Android) is being worked on in #1000! It works completely without a notification server by having Thunder make a periodic poll. Unfortunately this won't work on iOS and has various limitations depending on Android manufacturer, but it appears like the solutions mentioned in [LemmyNet/lemmy#2631](https://github.com/LemmyNet/lemmy/issues/2631) have similar issues, so this is at least a start on our part. 😊

I think it would still be a very good idea to have unified push supported. It isn't only push notifications that it adds. It also synchronizes the polling of all apps that supports it (this enables the device to remain sleeping for longer without being interrupted), and it supports the transition away from proprietary services like Google's, and Apple's push notifications.

micahmo:
@K4LCIFER and @Tealk, can you give a brief overview of how we could integrate UnifiedPush with Thunder?

* Would users self-host the server?
* Could we integrate the listener directly into Thunder, or would it be a separate app?

K4LCIFER:
> @K4LCIFER and @Tealk, can you give a brief overview of how we could integrate UnifiedPush with Thunder?
> 
>     * Would users self-host the server?
> 
>     * Could we integrate the listener directly into Thunder, or would it be a separate app?

It is explained well [here](https://unifiedpush.org/) (or, at least, far better than I could summarize :stuck_out_tongue_winking_eye: )

[...]

Issue #125, opened 2023-06-27
Automatic check for replies


shortwavesurfer2009:
**Is your feature request related to a problem? Please describe.**
The app should pull for any notifications on a user defined schedule like once per hour, once every three hours, etc. This would have to be an API call unless the lemmy server supports sending push notifications. In that case this could be done with UnifiedPush

**Describe the solution you'd like**
Thunder should check for notifications

hjiangsu:
Good idea - that is something I want to include as more features are integrated for the inbox! For the near future, my general plans is to support updating of notifications within the app itself (every time you open the app, it'll do a quick fetch of any new notifications and will show up as a small alert in the inbox icon)

I would like to get to push notifications eventually, but that requires a lot more work to set up (might need a dedicated server to poll for new notifications)

tcely:
A lot of the notifications infrastructure is already set up for you by UnifiedPush.

https://github.com/UnifiedPush/flutter-connector

trymeouteh:
> A lot of the notifications infrastructure is already set up for you by UnifiedPush.
> 
> https://github.com/UnifiedPush/flutter-connector

Would be nice if Thunder has its own push notification system and also had the option to use Unified Push and this can be changed with a simple setting in the app.

shortwavesurfer2009:
> > A lot of the notifications infrastructure is already set up for you by UnifiedPush.
> > https://github.com/UnifiedPush/flutter-connector
> 
> Would be nice if Thunder has its own push notification system and also had the option to use Unified Push and this can be changed with a simple setting in the app.

It could be set to poll for new notifications say once per hour unless UnifiedPush is detected. Then thunder could defer to the UnifiedPush provider

shortwavesurfer2009:
I am going to go ahead and close this because the polling every 15 minutes is good enough for me.

PR #1000, opened 2023-12-22
Feature/notifications

micahmo:
## Pull Request Description

This PR introduces OS notifications for inbox messages. All message checking and notification generation is done locally (i.e., no Firebase needed).

**New Packages**
 - [`flutter_local_notifications`](https://pub.dev/packages/flutter_local_notifications)
   - Has the ability to generate local notifications, manage permissions, etc.
 - [`background_fetch`](https://pub.dev/packages/background_fetch)
   - Allows the app to periodically run a process in the background (in this case, polling for new messages), even when closed.

**Notes**
> The following is a list of caveats and/or intentional limitations for this first round of implementation. Most of these things can be improved in the future, but this first round works great as is!
- I only set this up for Android. Perhaps @hjiangsu could do the iOS setup at some point, but it requires a bit of work for both packages. ([flutter_local_notifications setup](https://pub.dev/packages/flutter_local_notifications#-ios-setup) and [background_fetch setup](https://pub.dev/packages/background_fetch)).
- Enabling notifications will most likely increase battery usage due to the background checks. This is noted in the new setting description.
- There are some ["Release build configuration"](https://pub.dev/packages/flutter_local_notifications#release-build-configuration) changes suggested by `flutter_local_notifications` that I did not go through. I'd be curious to know if it works as-is (since it works fine in Debug), but we can test together.
- Clicking a notification only brings you to inbox page. It does not specifically focus the relevant message.
  - My idea to fix this in a future iteration is to add a new page which can display a single message and be shown on top of the navigation stack. This allows you to interact with the message, then go back to whatever you were doing in Thunder. (Right now, the page scroll to the Inbox page doesn't work so well if you're looking at a post, for example.)
- Notifications only work for replies, not mentions or messages.
- Notifications only work for the active account in Thunder.
- The interval of `background_fetch` is not super configurable (at all on iOS, and not without more permissions on Android). But I think the default of 15(-ish) minutes is fine!
   - For the sake of the demo video, the polling time was reduced. The code includes comments on how to do this.

## Issue Being Fixed

Issue Number: First step / partial implementation of #125 and #219.

### Demo of permissions flow

https://github.com/thunder-app/thunder/assets/7417301/57885668-7094-4a81-b27c-fef283f8a5e2

### Demo of receiving notifications while app is open

https://github.com/thunder-app/thunder/assets/7417301/f9d29af6-e0cf-4d35-b57f-f38efde20cda

### Demo of receiving notifications while app is closed

https://github.com/thunder-app/thunder/assets/7417301/825d5e6c-ad0a-46e0-9858-c9d63c3ff6d6

## Checklist


hjiangsu:
I would just like to say that this is awesome! I was also looking at ways to have notifications without having to host a server holding JWT tokens (for security purposes)

I'll definitely take a look at this when I have time and see if it can be adapted to iOS as well 😄

micahmo:
> I would just like to say that this is awesome! I was also looking at ways to have notifications without having to host a server holding JWT tokens (for security purposes)

Thanks, yeah, this seems to be a good enough solution, as long as we're ok with Thunder making the occasional API call!

If you'd like to try this out on your Android emulator to get a feel for how it works, I'd recommend a couple of changes to make testing easier.

* Change this value to `1`.
  https://github.com/thunder-app/thunder/blob/04cd9d7502551e0618e62e6bc84bc2adb0df26a5/lib/main.dart#L237
* Uncomment this line.
  https://github.com/thunder-app/thunder/blob/04cd9d7502551e0618e62e6bc84bc2adb0df26a5/lib/main.dart#L247
* Comment out the `where` part of this line.
  https://github.com/thunder-app/thunder/blob/04cd9d7502551e0618e62e6bc84bc2adb0df26a5/lib/utils/notifications.dart#L57
* Mark some messages as unread _or_ change this to `false`.
  https://github.com/thunder-app/thunder/blob/04cd9d7502551e0618e62e6bc84bc2adb0df26a5/lib/utils/notifications.dart#L49

With those things set, you should get some notifications for some/all inbox messages about once per minute whether the app is running or not!

hjiangsu:
So I've been looking into the packages, especially [background_fetch](https://pub.dev/packages/background_fetch), and it seems like there might be some inconsistencies in how this works (depending on platform and manufacturer)

For iOS, background fetching will not work when the app is terminated. This means that we can't really use this package reliably to fetch for notifications every ~15 mins. This will only work if the app is either in the foreground, or when its not terminated (in the background)

For Android, it seems like the reliability of background fetching varies depending on the device manufacturer (https://dontkillmyapp.com/). There seems to be some workarounds for the specific manufacturers but each one is different.
- https://github.com/transistorsoft/flutter_background_fetch/issues/355
- https://github.com/transistorsoft/flutter_background_fetch/issues/339

hjiangsu:
Yeah, unfortunately, I think the only viable solution on iOS would be to set up a server to handle push notifications. I could play around with it a bit and see how that would work, but needing to have the JWT token stored on the server in order to fetch notifications feels like a big security risk.

There might be a potential for self-hosting the server which reduces the security risk for those who are able to do so, but I don't know how that interacts with the push notification services (whether or not that's even possible)

I think we can still go ahead and push this local notifications since its really useful to have, but we should be clear that it's experimental and might not work on all Android devices (and provide some instructions on what to do to prevent the app from being killed in the background)

micahmo:
> we should be clear that it's experimental and might not work on all Android devices (and provide some instructions on what to do to prevent the app from being killed in the background)

Any recommendations for how we should communicate this? Maybe a popup warning when enabling the feature?

hjiangsu:
I think adding "Experimental" to the subheading of the option works well to let the user know that it may not fully work. We can also have a popup dialog which explains how the notifications work so that its more transparent
- Only happens on ~15 mins intervals, disable battery optimizations to give it the best chance to work properly, etc.

micahmo:
> I think adding "Experimental" to the subheading of the option works well to let the user know that it may not fully work. We can also have a popup dialog which explains how the notifications work so that its more transparent

Sounds good! How does this look?

https://github.com/thunder-app/thunder/assets/7417301/851e6302-5a75-47a9-88a4-21eab7a0ffad

Here's a screenshot of just the dialog so you can read it.

| image |
| - |

lib/settings/pages/general_settings_page.dart
```diff
@@ -662,7 +702,7 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
                   },
                   subtitle: enableInboxNotifications
                       ? !kIsWeb && Platform.isAndroid && areAndroidNotificationsAllowed == true
-                          ? l10n.backgroundCheckWarning
+                          ? null
```
micahmo, line 705: I took this out because this information is now covered in the dialog (and the experimental heading) and to keep the settings page more clean.

micahmo:
> It'll take some time for me to review this PR

No problem at all! I really appreciate that you're willing to review and accept this feature. Take all the time you need.

> I'll most likely just generate a release build for you to test

Sounds good!

android/build.gradle
```diff
@@ -1,12 +1,20 @@
 buildscript {
     ext.kotlin_version = '1.7.10'
+    // Versions recommended/needed by flutter_local_notifications and background_fetch
+    // These can be upgraded over time.
+    ext {
+        compileSdkVersion   = 33
+        targetSdkVersion    = 33
+        appCompatVersion    = "1.6.1"
```
hjiangsu, line 8: Just curious, but where did "1.6.1" come from?

lib/settings/pages/general_settings_page.dart
```diff
@@ -221,13 +236,24 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
...
   }
 
   @override
   void initState() {
     super.initState();
-    WidgetsBinding.instance.addPostFrameCallback((_) => _initPreferences());
+    WidgetsBinding.instance.addPostFrameCallback((_) async {
+      _initPreferences();
+
+      // Check whether Android is currently allowing Thunder to send notifications
```
hjiangsu, line 249: I would place all this logic into its own function (e.g., similar to `_initPreferences`). This way, you can perform the async operations!

hjiangsu (commented):
I've pushed some commits to update with the latest develop changes, and also to add in release build configuration! I can create a release build for you to test out if you'd like (I can send it over matrix)

I also did a quick code review and had some comments!

android/app/build.gradle
```diff
@@ -52,9 +55,12 @@ android {
         // You can update the following values to match your application needs.
         // For more information, see: https://docs.flutter.dev/deployment/android#reviewing-the-gradle-build-configuration.
         minSdkVersion 21
-        targetSdkVersion flutter.targetSdkVersion
+        // Use variable for version requested by background_fetch and flutter_local_notifications
+        targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
+        // The following line is required by flutter_local_notifications
+        multiDexEnabled true
```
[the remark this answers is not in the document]
micahmo: Good catch! I'll take it out and we'll see if things still work. 😊

android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
micahmo: I don't believe so! If you check the first bullet point under the heading you linked, it says...

> To schedule notifications the following changes are needed

We are not using the schedule functionality, so I don't believe anything there applies to us.

P.S. At first I was hoping that the schedule functionality allowed you to execute some code and perhaps conditionally generate a notification, but unfortunately, it only lets you schedule a predefined notification at a predefined interval, so it wouldn't be much help.

android/build.gradle
micahmo: It essentially comes from [this section](https://github.com/transistorsoft/flutter_background_fetch/blob/master/help/INSTALL-ANDROID.md#androidbuildgradle) of the `background_fetch` setup guide.

``` diff
+       appCompatVersion    = "1.4.2"           // or latest
```

Notice it allows for the latest. I believe I ran into an issue and someone suggesting upgrading, so I looked up the latest  [here](https://developer.android.com/jetpack/androidx/releases/appcompat), and it is 1.6.1.

Sorry I don't remember the specific issue that prompted me to upgrade.

micahmo:
> I've pushed some commits to update with the latest develop changes, and also to add in release build configuration! I can create a release build for you to test out if you'd like (I can send it over matrix)
> 
> I also did a quick code review and had some comments!

Thank you so much for reviewing and adding the release configs! I've addressed your PR comments. If/when you get a chance, feel free to shoot over a release build in Matrix and I'll see how things work!

android/build.gradle
hjiangsu: That's all good! I was just curious

android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
hjiangsu: Sounds good! Just double-checking 😄

lib/settings/pages/general_settings_page.dart
hjiangsu: Ahh, I think I misinterpreted the comment. We can do something like this so that it doesn't mutate `areAndroidNotificationsAllowed` directly:

Although in this case, it might be fine!

```dart
bool _areAndroidNotificationsAllowed = await androidFlutterLocalNotificationsPlugin?.areNotificationsEnabled();

setState(() {
  areAndroidNotificationsAllowed: _areAndroidNotificationsAllowed,
});
```

micahmo: Ok I updated it to do this! I think the end result is the same (the UI is redrawn with the current state), but I agree that this provides a clearer intention as to why `setState` is being called at all.

hjiangsu (approved):

PR #1001, opened 2023-12-22
Fixed issue where Safari extension would cause infinite reloads when loading lemmy instance

hjiangsu:
## Pull Request Description

This PR fixes an issue with Thunder's Safari app extension where loading a lemmy link would cause the page to reload. The main issue here seems to be that there was a change to a `URL`'s parameters, making most of the attributes read-only.

Because of this change, I've adjusted the logic so that it creates the modified URL with the custom URL scheme when initializing a new URL (rather than changing the `protocol` attribute after the fact). I've also refactored and simplified the code a bit more.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/ios/Open In Thunder/Resources/content.js
+++ b/ios/Open In Thunder/Resources/content.js
@@ -60,41 +60,28 @@ let instances = [
     "yiffit.net"
 ];
 
-const observeUrlChange = () => {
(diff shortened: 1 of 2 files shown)


micahmo (approved):
LGTM and I'll trust you that it works. :blush:

PR #1008, opened 2023-12-26
Improve number formatting

micahmo:
## Pull Request Description

This PR introduces a couple of changes related to number formatting.

I updated `formatNumberToK` to use the [intl](https://pub.dev/packages/intl) package's `NumberFormat.compact()` format rather than formatting it ourselves. There wasn't anything wrong with our code, but this is cleaner! It does make some small improvements like dropping trailing zeros (`49.0K` -> `49K`).

I also updated search results a bit. We previously displayed the full subscriber count unformatted, which was a bit hard to read (like `45827`). We could pare it down using `ToK`, but since we have plenty of room to display the full number, I used `decimalPatternDigits` instead. It would format the previous number as `45,827` (which I find much easier to read!). And since this is coming from `intl`, it will be locale-specific (e.g., `.` instead of `,`).

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| image |
| - |

## Checklist

### DIFF
--- a/lib/search/pages/search_page.dart
+++ b/lib/search/pages/search_page.dart
@@ -43,6 +43,7 @@ import 'package:thunder/utils/global_context.dart';
 import 'package:thunder/utils/instance.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:thunder/utils/navigate_user.dart';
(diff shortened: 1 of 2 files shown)


hjiangsu (approved):
LGTM!

PR #1010, opened 2024-01-01
Fix thunder being locked to 60Hz on 120Hz displays

mufeedali:
## Pull Request Description

Set Thunder to use the highest available display refresh rate.

## Issue Being Fixed

On several devices with high refresh rate displays, like my Nothing Phone 1,  Flutter apps get locked to 60 FPS even though my display mode is still supposed to be 120 Hz. This should fix it by using the [`flutter_displaymode`](https://github.com/ajinasokan/flutter_displaymode) plugin to set the highest available refresh rate.

## Checklist


micahmo (approved):
LGTM, thanks for doing this!

hjiangsu (approved):
Thanks for the contribution! Everything looks good to me - I made a small adjustment to the changelog so that it's under the correct version (this fix will be slated for 0.2.8)

Issue #1006, opened 2023-12-23
Unable to follow "Lemmy Links"


K4LCIFER:
**Description**  
You appear to be unable to follow/tap a [Lemmy Link](https://github.com/LemmyNet/lemmy/issues/2987) despite it being highlighted blue.

**How to Reproduce**  
1. Locate a Lemmy Link
2. Tap on it
3. Observer that nothing happens

**Expected Behavior**  
I would expect that the specific Lemmy link would be followed

**Screenshots**  
None, at the moment. I'll update this post with a video confirmation when I come across another Lemmy Link.

**Device & App Version:**  
 - Device: Google Pixel 6
 - OS: Android 14
 - App Version 0.2.7-6

**Additional Context** 
None

hjiangsu:
I wonder if this is related to #997

micahmo:
Ditto! As far as I know, we support Lemmy links for users, communities, posts, and comments in all of the web formats and hyperlinks, plus a few! If navigation doesn't work, it may be because it's federated from a different platform, like was fixed in #997, or because it hasn't federated to your current instance. Like @hjiangsu, and example of a bad link would really help!

K4LCIFER:
I think I can close this, as I haven't been able to find a link that confirms this issue.

PR #1014, opened 2024-01-04
User navigation fix and error handling

micahmo:
## Pull Request Description

This PR has a couple of fixes for issues related to navigation.

First, I noticed that our regexes for communities were accidentally matching a particular username syntax, like this: `/u/user@instance.tld`, because we were essentially looking for `anything@anything`. With this fix, if the community checker parses a name starting with `/u/`, it will be ignored.

Secondly, I added some error handling around the community navigation page. While we're now less likely to get into this error scenario, it would still be nice to know if the navigation is failing because we are trying to load a community we can't find, so now there is a snackbar indicating when a community load fails.

> Review without whitespace.

## Issue Being Fixed

Issue Number: Could possibly be related to #1006.

## Screenshots / Recordings

This video demonstrates the new error handling prior to fixing the community detection. This would clue us in that we're trying to load this user as a community, and something else is wrong to be investigated!

https://github.com/thunder-app/thunder/assets/7417301/8964f697-0689-43d1-a97b-08462813ca5e

This video demonstrates the correct user navigation after applying the community detection fix.

https://github.com/thunder-app/thunder/assets/7417301/38156397-f49a-4e1b-940e-85700857df24

## Checklist

### DIFF
--- a/lib/feed/bloc/feed_bloc.dart
+++ b/lib/feed/bloc/feed_bloc.dart
@@ -121,7 +121,7 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
 
   /// Handles clearing any messages from the state
   Future<void> _onFeedClearMessage(FeedClearMessageEvent event, Emitter<FeedState> emit) async {
-    emit(state.copyWith(status: FeedStatus.failure, message: null));
+    emit(state.copyWith(status: state.status == FeedStatus.failureLoadingCommunity ? state.status : FeedStatus.success, message: null));
   }
 
   /// Handles post related actions on a given item within the feed
@@ -304,7 +304,7 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
           } catch (e) {
             // If we are given a community feed, but we can't load the community, that's a problem! Emit an error.
             return emit(state.copyWith(
-              status: FeedStatus.failure,
+              status: FeedStatus.failureLoadingCommunity,
               message: getExceptionErrorMessage(e, additionalInfo: event.communityName),
               feedType: event.feedType,
             ));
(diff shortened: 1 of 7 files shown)


lib/feed/bloc/feed_bloc.dart
```diff
@@ -120,7 +121,7 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
 
   /// Handles clearing any messages from the state
   Future<void> _onFeedClearMessage(FeedClearMessageEvent event, Emitter<FeedState> emit) async {
-    emit(state.copyWith(status: FeedStatus.success, message: null));
+    emit(state.copyWith(status: FeedStatus.failure, message: null));
```
hjiangsu, line 124: Just curious, but why was this changed? From what I'm reading, if we get an error message/snackbar that shows up, the feed will become blank because of this line:

```dart
if (state.status == FeedStatus.failure)
SliverToBoxAdapter(
  child: Container(),
),
```

Is this the intended action? Or maybe I missed something here!

(e.g., we fail to fetch more posts from a feed -> which results in throwing a snackbar for the error -> the entire feed will become blank rather than just showing the snackbar and already loaded posts)

micahmo: Good catch. I made this change because, without it, if we fail to initially load a feed (such as in the first screen recording) we will be stuck with the circular progress bar forever, and I wanted to indicate that we tried and failed, and nothing else is loading.

However, you make a great point that if the feed is already populated, this will wipe it.

I will see if I can come up with another way to handle this!

micahmo: Alright, I just ended up adding a new status, since the new code is technically handling an error scenario that wasn't covered before (a failure to load the initial information about the community), which is different than a failure to load more posts, for example. The end result is that the behavior is the same as the first screen recording, while only for this scenario.

hjiangsu: Looks good now! Thanks for making those changes 😄

hjiangsu (approved):

hjiangsu:
I just noticed one issue with this as I was testing it - the drawer icon no longer opens the feed/subscriptions drawer.

This seems to be related to the `Scaffold` that was added in `feed_page.dart`. When tapping on the drawer icon, it attempts to look for the closest `Scaffold` parent in the widget tree which happens to be the one that was added in. Since that `Scaffold` does not have a drawer associated with it, it doesn't perform any action!

PR #1015, opened 2024-01-05
Fix prioritized community sort

micahmo:
## Pull Request Description

In PR #984, I added the ability for favorited communities to be prioritized in lists of communities like search results. The problem is that the sorting algorithm always fell back to overall subscriber count for non-favorites, making it essentially a `Top of all time` sort. This is fine for the auto-complete prompt, but not for search, where a specific sort maybe chosen.

This PR fixes that algorithm to respect the existing order of communities when prioritizing favorites, rather than re-sorting the whole list.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Observe that the selected sort type is now respected, both within favorites, which remain at the top, and within the rest of the list.

https://github.com/thunder-app/thunder/assets/7417301/f076ff56-f0f1-442e-acbd-b342ac03e544

## Checklist

### DIFF
--- a/lib/feed/utils/community.dart
+++ b/lib/feed/utils/community.dart
@@ -77,14 +77,36 @@ Future<void> toggleFavoriteCommunity(BuildContext context, Community community, 
 }
 
 /// Takes a list of [communities] and returns the list with any [favoriteCommunities] at the beginning of the list
-/// Note that you may need to call [toList] on any lists that are marked as readonly.
+/// Note that you may need to call [toList] when passing in lists that are marked as readonly.
 List<CommunityView>? prioritizeFavorites(List<CommunityView>? communities, List<CommunityView>? favoriteCommunities) {
-  return communities
-    ?..sort(
-      (a, b) => favoriteCommunities?.any((c) => c.community.id == a.community.id) == true
-          ? -1
-          : favoriteCommunities?.any((c) => c.community.id == b.community.id) == true
-              ? 1
-              : b.counts.subscribers.compareTo(a.counts.subscribers),
-    );
+  // This is the list we will return. If we don't do any prioritization,
+  // it will just be the initial list we received.
+  List<CommunityView>? prioritizedCommunities = communities;
+
+  // Only attempt to prioritize favorites if the list of communities and favorites is not empty
+  if (communities?.isNotEmpty == true && favoriteCommunities?.isNotEmpty == true) {
+    prioritizedCommunities = [];
+    final List<CommunityView> sortedFavorites = [];
+
+    // Iterate through the list in reverse so that we can insert things at index 0 while preserving the order.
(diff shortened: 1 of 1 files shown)


lib/feed/utils/community.dart
```diff
@@ -77,14 +77,36 @@ Future<void> toggleFavoriteCommunity(BuildContext context, Community community,
...
-/// Note that you may need to call [toList] on any lists that are marked as readonly.
+/// Note that you may need to call [toList] when passing in lists that are marked as readonly.
 List<CommunityView>? prioritizeFavorites(List<CommunityView>? communities, List<CommunityView>? favoriteCommunities) {
-  return communities
-    ?..sort(
-      (a, b) => favoriteCommunities?.any((c) => c.community.id == a.community.id) == true
-          ? -1
-          : favoriteCommunities?.any((c) => c.community.id == b.community.id) == true
-              ? 1
-              : b.counts.subscribers.compareTo(a.counts.subscribers),
```
micahmo, line 88: This was the problematic line from before. 😊

lib/feed/utils/community.dart
```diff
@@ -77,14 +77,36 @@ Future<void> toggleFavoriteCommunity(BuildContext context, Community community,
...
-    );
+  // This is the list we will return. If we don't do any prioritization,
+  // it will just be the initial list we received.
+  List<CommunityView>? prioritizedCommunities = communities;
```
hjiangsu, line 84: Hmm, I _believe_ there's a less verbose way to do this!

Instead of having to loop through the list in reverse, we can just perform a filter on the original `communities` list to get favorited and non-favorited communities. Once we have those, we can combine them into one. I feel like this makes it a bit easier to understand what's happening.

I wrote up some code to explain what I mean (its untested so not sure if it'll work properly) Let me know what you think!

```dart
// Create a set of the favorited community ids for filtering later
Set<int> favoriteCommunityIds = Set<int>.from(favoriteCommunities!.map((c) => c.community.id));

// Filters out communities that are part of the favorites, and keeps the same order
List<CommunityView> sortedFavorites = communities.where((community) => favoriteCommunityIds.contains(community.id)).toList();

// Filters out communities that are not a part of the favorites, and keeps the same order
List<CommunityView> sortedNonFavorites = communities.where((community) => !favoriteCommunityIds.contains(community.id)).toList();

// Combine them together, with favorites at the top
List<CommunityView> prioritizedCommunities = List<CommunityView>.from(sortedFavorites)..addAll(sortedNonFavorites);
```

micahmo: Sounds good! I had to make a few teaks, but I agree it's _much_ more readable! 😊

hjiangsu (approved):
LGTM!

PR #1016, opened 2024-01-05
Fix an issue where you could favorite a non-subscribed community

micahmo:
## Pull Request Description

This PR hides the "Add to favorites" menu item from the community view if the community is not already a subscription. Otherwise, favoriting has no effect.

Note: I considered putting the Refresh button back up on the main AppBar when you are not subscribed (since there would only be three icons, and since otherwise it would be the only item in the overflow menu). However, if you subscribe to a community from the community view, then the AppBar would redraw and the refresh button would jump into the menu, which I thought would look very funny!

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/5a4ad0ef-c5e3-4657-b3d5-18fb4b86c385

## Checklist

### DIFF
--- a/lib/feed/widgets/feed_page_app_bar.dart
+++ b/lib/feed/widgets/feed_page_app_bar.dart
@@ -121,19 +121,20 @@ class FeedPageAppBar extends StatelessWidget {
                   title: Text(l10n.refresh),
                 ),
               ),
-              PopupMenuItem(
-                onTap: () async {
-                  final Community community = context.read<FeedBloc>().state.fullCommunityView!.communityView.community;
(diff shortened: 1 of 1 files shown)


hjiangsu:
> if you subscribe to a community from the community view, then the AppBar would redraw and the refresh button would jump into the menu, which I thought would look very funny!

I agree! I also think we can start adding in more actions in that dropdown in the future anyways so it won't look as awkward having just a single action there (e.g., blocking community, triggering community sidebar, etc)

PR #1018, opened 2024-01-05
Fix a couple more search state issues

micahmo:
## Pull Request Description

This PR fixes issues with search where certain operations ((un)subscribe/(un)favorite/etc.) could cause the search screen to reload.

All of the scenarios I found where related to changes in #984 where we reset the search state if the list of favorites has changed. This is fine if we're on the search landing page, where the favorites are at the top, but it's jarring if we're currently in a search and everything resets. The fix is to only perform the reset if we are not currently searching.

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu (approved):
LGTM!

PR #1019, opened 2024-01-05
Show search type hints

micahmo:
## Issue Being Fixed

This PR adds a small QoL change which prompts the user to select another search type when there are no results.

This is a scenario that I encounter very often: I will search for something before realizing I am searching for the wrong type. I get no results, and then I have to pick a different type. By showing these prompts, the user can more quickly recognize *why* there were no results, and also pick a different type with one tap.

Note: I also refactored `SearchActionChip` because they all use the same several properties.

> Review without whitespace.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/a809d662-a141-468e-a783-61ee08d936d5

## Checklist


hjiangsu (approved):
Looks good, just a few minor comments!

PR #1034, opened 2024-01-10
Ability to create cross-posts from app bar button

micahmo:
## Pull Request Description

This PR adds the ability to create a new cross-post from the app bar menu. This is needed because, although the `CrossPosts` widget has a button for that, it is only displayed if the post is already a cross-post. This new menu item will allow the function to be accessed regardless of the state of the current post.

I decided to add this to an overflow menu rather than directly on the app bar because I figured we'd have more buttons soon!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/ed9cc747-e5a7-445a-8a3d-aabae32ba5a6

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -167,7 +167,7 @@ class _CreatePostPageState extends State<CreatePostPage> {
       _titleTextController.text = widget.title ?? '';
       _bodyTextController.text = widget.text ?? '';
       _urlTextController.text = widget.url ?? '';
-      _getDataFromLink();
+      _getDataFromLink(updateTitleField: _titleTextController.text.isEmpty);
 
       if (widget.image != null) {
         WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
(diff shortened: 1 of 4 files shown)


hjiangsu:
Thanks for getting to this! I do have one point for discussion, mainly on how cross-posts work on Lemmy.

It looks like we're only handling the case where cross-posts can be created for links/images. I believe we can also create cross-posts for text posts. In this case, the post body gets copied over and is placed into a block quote markdown. We can do something similar in this case, or just copy that over without any changes.

This is an example of a text cross-post: https://lemmy.world/post/10508184

hjiangsu:
> My only concern is that it's not a "real" cross-post in the sense that Lemmy has no knowledge of the original post nor any linkage between them.

I agree, but I still think we should try to match parity with what Lemmy UI does!

> Do you think we should follow it exactly? Like this?

I think thats fine!

micahmo:
All set!

https://github.com/thunder-app/thunder/assets/7417301/87d3165c-f662-460c-8196-ee3859e629c6

hjiangsu:
Looks good now. Thanks for doing that!

hjiangsu (approved):

PR #1035, opened 2024-01-10
Account selector improvements and minor fixes

hjiangsu:
## Pull Request Description
This PR adds the following changes/fixes:
- Fixes an issue where long pressing on the account option on the bottom navigation bar would often times show the tooltip instead of the account selector
- Fixes a minor visual bug where the drag handle of the account selector was mismatched
- Refactors the account selector modal to use slivers, and moved add account/instance to the top

Let me know what you think of the changes!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

simulator_screenshot_9F30DD9B-8AE1-4C93-B5B6-37BF005D14EE

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,13 @@
 - Added ability to subscribe/unsubscribe to community from long press action on posts
 - Added option to hide top app bar on scroll
 
+## Changed
+- Small UI adjustments for account switcher
+
 ### Fixed
 - Fixed issue where Thunder was being locked to 60Hz on 120Hz displays on Android
 - Fixed issue where subscriptions list in drawer is not respecting alphabetical order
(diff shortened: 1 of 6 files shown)


micahmo (approved):
LGTM!

My only concern is that the "add anonymous account" button might not be obvious and people might miss that the feature exists. Maybe in the future, we can have a single button, and within the account creation dialog, we can have an option to switch it to anonymous, which will just hide the user fields.

hjiangsu:
> My only concern is that the "add anonymous account" button might not be obvious and people might miss that the feature exists. Maybe in the future, we can have a single button, and within the account creation dialog, we can have an option to switch it to anonymous, which will just hide the user fields.

I tried to look for some more distinguishable icons for the anonymous accounts, but was unable to find any that suited. I do think your idea of having this in the account creation dialog makes sense!

PR #1036, opened 2024-01-10
search preferences/settings

ggichure:
## Pull Request Description

Added a way to search through preferences. 
## Issue Being Fixed

Issue Number: N/A

### Updated recording
https://github.com/thunder-app/thunder/assets/31619962/58e6af2f-6181-49ae-aeca-9d2433eb1139

## Checklist


ggichure:
Some strings are similar.
Image

I propose we use the `LocalSettings.name` to help distinguish them.
Image

Final result.
```dart
subtitle: Text(
  localSettings[index].name
    .replaceFirst('_', '')
    .replaceAll(RegExp(r'setting|general|gesture'), '')
    .replaceAll('_', ' ')
),
```

Image


@hjiangsu @machinaeZER0 @micahmo  thoughts.

micahmo:
> I propose we use the `LocalSettings.name` to help distinguish them.

Unfortunately the names are not translated (and neither are the labels). I think we need a more general way to assign the translated label to the setting enum. Then we can also assign a category and sub-category. Once that's done, you can display search results like this:

`Category -> Sub-Category -> Setting name`

For an example:

`Floating Action Button > Feed > Back To Top`

We started a discussion [here](https://github.com/thunder-app/thunder/pull/944#discussion_r1416088515) but it didn't get very far. 😊

machinaeZER0:
I love this idea! Guess I never noticed this behavior since I hadn't used the search bar for this purpose. We're pretty sure we wouldn't want to take the strings as a stopgap solution, given the translation issue? I do agree that @micahmo 's solution is probably the best final course of action, and if an interim solution would be messier/create unnecessary work then maybe it does make sense to wait. The issue is real, but I guess is more of an annoyance for now?

hjiangsu:
> Unfortunately the names are not translated (and neither are the labels). I think we need a more general way to assign the translated label to the setting enum.

I've added that in #1045. This only has mapping for the `LocalSettings` localizations, but we can also add mappings to category/subcategory. Let me know if this solution works! This is the general idea I have for how to do the category/subcategory.

`app_en.arb`
```json
{
  "defaultFeedType": "Default Feed Type",
  "general": "General",
  "feedTypeAndSorts": "Default Feed Type and Sorting"
}
```

`local_settings.dart`
```dart

enum LocalSettings {
  defaultFeedListingType(
    name: 'setting_general_default_listing_type',
    key: 'defaultFeedType',
    categoryKey: 'general',
    subcategoryKey: 'feedTypeAndSorts'
  );

  const LocalSettings({
    required this.name,
    required this.key,
    this.categoryKey,
    this.subcategoryKey,
  });
}

```

Here's an example of how you would get a localized string.

```dart
List<LocalSettings> localSettings = [
  LocalSettings.defaultFeedListingType,
  LocalSettings.defaultFeedSortType,
  ...
];

final l10n = AppLocalizations.of(context)!.
Text(l10n.getLocalSettingLocalization(localSettings[0].key)); // Retrieves "Default Feed Type"
Text(l10n.getLocalSettingLocalization(localSettings[0].categoryKey)); // Retrieves "General"
Text(l10n.getLocalSettingLocalization(localSettings[0].subcategoryKey)); // Retrieves "Default Feed Type and Sorting"
```

hjiangsu:
> Once https://github.com/thunder-app/thunder/pull/1045 Is merged , I'll make the necessary changes.

It should be merged now! I was just finishing up the first pre-release for 0.2.9 first before merging this one in.

micahmo:
Hey I just tested out this feature. Again, it's super awesome and something I've wanted for a while. Here are a few observations from my testing.
* The animation between the search widget and the search page is a tiny bit funny to me.
* If you have typed something into the search bar and then navigate back, and then into the search page again, the search text remains (is that intentional, or should we clear it?).
* The main search widget is typeable. If its goal is just to launch the search page, it doesn't need to be a real entry widget.
* Options that are not real "settings" (like the import action) are not searchable.
* If the result is off the page, it might not be obvious. I've seen other apps scroll to the setting and also flash it.
* Nested settings (like post appearance) are not searchable.

Note that these are just my observations and I haven't reviewed the code yet. Also I'm not asking you to change/fix anything. Just making notes. :blush:

ggichure:
> The animation between the search widget and the search page is a tiny bit funny to me.

Im using the default animation provided by the `SearchAnchor`

> If you have typed something into the search bar and then navigate back, and then into the search page again, the search text remains (is that intentional, or should we clear it?).

Its Intentional.  e.g On android devices settings.

> If the result is off the page, it might not be obvious. I've seen other apps scroll to the setting and also flash it.

 Currently I don't have a clear way to implement this. I'll need some help.

> Nested settings (like post appearance) are not searchable.

I reckon this are searchable e.g Comments from the video

> Options that are not real "settings" (like the import action) are not searchable.
 
This are not in the `LocalSettings` . I'll add them. 

> The main search widget is typeable, and the text stays there, and it is not in sync with the search page. This can cause funny results when typing, because it'll add a letter to the main widget, then open the page, then append the letter to what was previously searched.

 Didn't get this.

micahmo:
> > The animation between the search widget and the search page is a tiny bit funny to me.
> 
> Im using the default animation provided by the `SearchAnchor`

Ok if you're using something built-in then there's probably not much you can do about it. 😊

> > If you have typed something into the search bar and then navigate back, and then into the search page again, the search text remains (is that intentional, or should we clear it?).
> 
> Its Intentional. e.g On android devices settings.

Just for reference, my Android device settings search does not persist the search query (although it does have a history).

https://github.com/thunder-app/thunder/assets/7417301/df8974e5-5939-4670-b93d-63bcba86a50d

> > The main search widget is typeable, and the text stays there, and it is not in sync with the search page. This can cause funny results when typing, because it'll add a letter to the main widget, then open the page, then append the letter to what was previously searched.
> 
> Didn't get this.

Here is a demo of what I mean. It is odd to be that I can type in the main search widget, since it seems like it's only supposed to open the page. It's even more weird that the text is not in sync. I would say either typing there should not be allowed, or the text should be appended to the page.

https://github.com/thunder-app/thunder/assets/7417301/0c74c1df-4c57-489c-943c-76910f29a3bf

hjiangsu:
Thanks for doing this again @ggichure! It looks pretty awesome.

> If the result is off the page, it might not be obvious. I've seen other apps scroll to the setting and also flash it.

I believe we can split this off for a future enhancement, since it looks like it might take some more work to implement this feature.

> If you have typed something into the search bar and then navigate back, and then into the search page again, the search text remains (is that intentional, or should we clear it?).

Personally, I think we should clear it and reset the state. To me, that makes the most sense since navigating out of the search page means I'm done with changing whatever setting I meant to change in the first place.

> The animation between the search widget and the search page is a tiny bit funny to me.

> The main search widget is typeable, and the text stays there, and it is not in sync with the search page. This can cause funny results when typing, because it'll add a letter to the main widget, then open the page, then append the letter to what was previously searched.

I think this is following material's guidelines here: https://m3.material.io/components/search/overview. It looks like the main search text is supposed to be a text hint, rather than a dynamic traditional search bar. Maybe we can make it more apparent by not showing the cursor blink indicator.

micahmo:
> It looks like the main search text is supposed to be a text hint, rather than a dynamic traditional search bar. Maybe we can make it more apparent by not showing the cursor blink indicator.

Makes sense! Although see my last video demonstrating how you can type in it.

Also, would it make sense to remove the elevation here and on the search page to look a little more material?

lib/settings/pages/settings_page.dart
```diff
@@ -45,6 +51,101 @@ class SettingsPage extends StatelessWidget {
             toolbarHeight: 70.0,
             pinned: true,
           ),
+          SliverToBoxAdapter(
+            child: Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 18),
+              child: SearchAnchor(
+                viewHintText: l10n.search,
+                builder: (BuildContext context, SearchController controller) {
+                  if (Platform.isIOS) {
```
hjiangsu, line 60: Hmm, I think maybe we can leave this as the Material `SearchBar` if that's all good with you! Or remove it completely if we use `SearchAnchor.bar`.

My main reasoning for this is that we use `SearchAnchor` for suggestions and results, which is more of a "Material" component, rather than a component that uses Apple's HIG guidelines. Because of this, it might feel weird that we're switching between two different styles (Apple's HIG and Google's Material).

This is also consistent with the search page, where we use Material's Search UI!

lib/settings/pages/settings_page.dart
```diff
@@ -45,6 +51,101 @@ class SettingsPage extends StatelessWidget {
...
+                  localSettings.sortBy((setting) => setting.key);
+
+                  return List<ListTile>.generate(
+                      localSettings.length,
+                      (index) => ListTile(
+                            subtitle: Text(
+                                "${l10n.getLocalSettingLocalization(localSettings[index].category!.toString())} > ${l10n.getLocalSettingLocalization(localSettings[index].subCategory.toString())}"),
+                            onTap: () {
+                              String pageToNav = SETTINGS_GENERAL_PAGE;
+                              switch (localSettings[index].category) {
```
hjiangsu, line 99: A slightly cleaner approach here would be to use switch expressions (really up to you if you'd like to make this change, since its just styling)

e.g.,

```dart
switch (localSettings[index].category) {
  LocalSettingsCategories.posts => pageToNav = SETTINGS_APPEARANCE_POSTS_PAGE,
  LocalSettingsCategories.comments => pageToNav = SETTINGS_APPEARANCE_COMMENTS_PAGE,
  _ => pageToNav = SETTINGS_GENERAL_PAGE,
}
```

lib/settings/pages/settings_page.dart
```diff
@@ -45,6 +51,101 @@ class SettingsPage extends StatelessWidget {
             toolbarHeight: 70.0,
             pinned: true,
           ),
+          SliverToBoxAdapter(
+            child: Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 18),
+              child: SearchAnchor(
```
hjiangsu, line 57: You can use SearchAnchor.bar to make it more Material looking (e.g., proper padding values, etc). This also fixes the issue where it looks like the SearchAnchor is typeable!

https://api.flutter.dev/flutter/material/SearchAnchor/SearchAnchor.bar.html

hjiangsu (commented):
Just did a really quick review, and provided a couple of comments/feedback!

ggichure:
- No longer persisting the search view
- On back, hide keyboard and unfocus the searchbar .
- Set searchbar elevation to 0 and used `surfaceVariant` as the `barBackgroundColor`. 


https://github.com/thunder-app/thunder/assets/31619962/17d8ebf9-d1b8-4ac1-b2e3-97cc020b09fd

micahmo:
@ggichure I love the latest demo, nice work! 🥳

hjiangsu (approved):
LGTM now!

@ggichure Is it ready to be merged?

ggichure:
> LGTM now!
> 
> @ggichure Is it ready to be merged?

Yes.

PR #1042, opened 2024-01-12
Improve community and user name detection

micahmo:
## Pull Request Description

This PR removes another false-positive community name match by checking for `@`. It is similar to #1014.

I also added checks in the username detector for things that may be communities. They only reason we haven't needed those is just because the community check happens first and filters those out. But if we happened to call them in a different order, we would need those checks as well.

## Issue Being Fixed

Issue Number: Could possibly be related to #1006.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/91b08d58-6a15-4369-b126-649a51b68c33

## Checklist


hjiangsu (approved):
LGTM!

Issue #734, opened 2023-09-13
NullPointerException with gallery_saver package


hjiangsu:
**Description**  
There is currently an issue with gallery_saver which causes crashes on Android (according to crash logs). Looking at their repo indicates that the package is no longer actively maintained: https://github.com/CarnegieTechnologies/gallery_saver/issues

For this reason, we should either
- try to find an alternative package which can handle the same functions as gallery_saver to resolve these crash issues.
- implement a solution to prevent the crashes from occurring. This is the relevant issue I found on the repo: https://github.com/CarnegieTechnologies/gallery_saver/issues/152

```
Exception java.lang.NullPointerException:
  at carnegietechnologies.gallery_saver.GallerySaver.finishWithSuccess (GallerySaver.java)
  at carnegietechnologies.gallery_saver.GallerySaver.access$finishWithSuccess (GallerySaver.java)
  at carnegietechnologies.gallery_saver.GallerySaver$saveMediaFile$1.invokeSuspend (GallerySaver.java)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (BaseContinuationImpl.java)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.java)
  at android.os.Handler.handleCallback (Handler.java:942)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8762)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:604)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1067)
```

**How to Reproduce**  
Steps to reproduce the behavior: N/A

**Expected Behavior**  
Users should not be crashing when attempting to save an image.

**Screenshots**  
N/A

**Device & App Version:**  
 - Device: Android devices
 - OS: Android (unsure if this also affects iOS)
 - Version: 0.2.1, 0.2.2, 0.2.3

**Additional Context**

gwbischof:
Gallery saver also has a very old `http` dependency which is causing other dependency issues with this repo.

PR #1043, opened 2024-01-12
Migrate from gallery_saver package to gal for media downloads

hjiangsu:
## Pull Request Description
This PR migrates our image saving package from `gallery_saver` to `gal`. There are a couple of reasons for this change:
- Google Play services seems to indicate that there are occasional crashes occurring from the `gallery_saver` package.
- `gallery_saver` seems to be no longer actively maintained. Looking at their [repository](https://github.com/CarnegieTechnologies/gallery_saver/issues), there are a lot of open issues that have not been addressed by the author. Similarly, the package has not been updated in over two years.
- `gal` seems to be a good replacement as it handles the same features as `gallery_saver`, and is being actively maintained.

This PR also refactors a bit of the image saving logic to make it more readable. Once this is merged, and released to a general release build, the fork of `gallery_saver` that we have on the organization can be removed.

@micahmo Could you test this on a physical Android device? I tested it on an emulator and it seems to be functional. If possible, we should also perform a test on the minimum Android SDK that we support (I believe this would be API level 21?)

## Issue Being Fixed

Issue Number: #734

## Checklist

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,12 +6,14 @@
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
     <uses-permission android:name="android.permission.VIBRATE" />
+    <!-- requestLegacyExternalStorage is required for saving media to an album in API 29 -->
     <application
         android:label="Thunder"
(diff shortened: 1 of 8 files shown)


micahmo (approved):
The code LGTM and it also works great on my physical Android device!

ggichure (approved):
Also working as expected on a physical android.

PR #1045, opened 2024-01-12
Refactor local settings to use localization strings

hjiangsu:
## Pull Request Description

This PR adds a way to get a localized string based on a key/string. Flutter does not support mirrors, so this approach is more manual. It requires us to create a map between the key and the localization string. The following changes were made:
- Renamed `label` to `key` in LocalSettings enum. This was done because it makes more semantic sense.
- Created an extension on AppLocalizations with a new function `getLocalSettingLocalization`. This function takes in a key, and outputs the associated localization string.
- Refactored the existing code to handle the new changes

Note: for each new setting we add, we also need to add in the corresponding entry into the map so that it can be associates with the proper localization string.

Reference: https://stackoverflow.com/questions/55445624/how-to-get-a-property-by-this-name-in-string

## Issue Being Fixed

This relates to #1036 and hopefully provides a way to allow for localized strings for the search functionality.

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -1,138 +1,140 @@
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
 enum LocalSettings {
   /// -------------------------- Feed Related Settings --------------------------
   // Default Listing/Sort Settings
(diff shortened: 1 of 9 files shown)


micahmo (approved):
LGTM!

ggichure (approved):

Issue #858, opened 2023-10-27
Incognito link browsing


jgeorgeson:
**Describe the solution you'd like**
It appears the default, internal browser for links is Chrome Custom Tabs. I would like to have an option to make these Incognito Custom Tabs. 

**Describe alternatives you've considered**
Copy the link and manually open Chrome incognito window and paste the link.

**Additional context**

micahmo:
It looks like this is not a native feature of [`flutter_custom_tabs`](https://pub.dev/packages/flutter_custom_tabs). We'd have to see if there's a way to share to the browser as incognito, perhaps via an intent on Android.

Another option would be to revive our in-app browser, which wouldn't share history with any installed browser.

CTalvio:
Custom tabs will use whatever browser you have set to default.

With firefox at least you can make it use incognito tabs for this with a setting in firefox.

Settings>Private Browsing>Open links in a private tab

Perhaps Chrome has an equivalent setting?

hjiangsu:
I've looked a little bit into this and like @micahmo mentioned, there is no native way to handle incognito from the package we're using.

However, it is possible to get incognito mode to work on Chrome custom tabs if we fork the package and add some custom logic to it. We can attempt to make a feature request in the package's GitHub and see if the author is willing to add this in.

micahmo:
@jgeorgeson How do you feel about the solution in #1046 of using the in-app browser? Since it's not tied to any of your system browsers, the history won't be stored anywhere. And of course, you can always open the page in your system browser if you want to. Would that solve your use-case?

PR #1046, opened 2024-01-12
Bring back in-app browser

micahmo:
## Pull Request Description

This PR brings back the option to use an in-app browser, which I think this provides some current and future benefits, such as...
* The potential ability integrate better with in-app gestures, like swipe back ([#729](https://github.com/thunder-app/thunder/issues/729)) or swipe up for comments (like Relay), or even a button to jump to comments (like we can do from the media preview)
* No history stored by Thunder or system browser ([#858](https://github.com/thunder-app/thunder/issues/858))
* The ability to parse and re-render the page ourselves, potentially opening the possibility for a reader mode ([#760](https://github.com/thunder-app/thunder/issues/760))
* A more native look-and-feel

In addition to re-adding the ability to use the internal browser, I also improved it a bit.
* Added back/forward navigation buttons and moved original buttons to overflow menu
* Added page Title and URL to app bar

> Review without whitespace.

## Issue Being Fixed

Issue Number: No specific issue, but see above.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/16888ac6-a3f7-4cc9-984f-b8c8bc17bd64

## Checklist


lib/settings/pages/general_settings_page.dart
```diff
@@ -228,7 +229,15 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
       enableCommentNavigation = prefs.getBool(LocalSettings.enableCommentNavigation.name) ?? true;
       combineNavAndFab = prefs.getBool(LocalSettings.combineNavAndFab.name) ?? true;
 
-      openInExternalBrowser = prefs.getBool(LocalSettings.openLinksInExternalBrowser.name) ?? false;
+      browserMode = BrowserMode.values.byName(prefs.getString(LocalSettings.browserMode.name) ?? BrowserMode.customTabs.name);
+      // Migrate the openInExternalBrowser setting, if found.
```
hjiangsu, line 233: I noticed this is done in two different places. Could we maybe set up a system for migrating deprecated preferences?

My initial thought is to have a function that is ran on app startup (e.g., in the `main` function) which performs any required migrations. This can be done in main so that it gets triggered before `ThunderBloc` is initialized!

```dart
// main.dart
performSharedPreferencesMigration();
```

```dart
// lib/utils/preferences/migration.dart

performSharedPreferencesMigration() async {
  // Fetch system preferences
  // Perform migrations here
}
```

lib/core/enums/local_settings.dart
```diff
@@ -76,8 +76,10 @@ enum LocalSettings {
 
   scrapeMissingPreviews(
       name: 'setting_general_scrape_missing_previews', key: 'scrapeMissingLinkPreviews', category: LocalSettingsCategories.general, subCategory: LocalSettingsSubCategories.linksBehaviourSettings),
+  // Deprecated, use browserMode
   openLinksInExternalBrowser(
```
hjiangsu, line 80: Let's also make sure this no longer appears as part of a search result!

hjiangsu (approved):
Sorry for taking so long to look at this! I think the motivations for this make sense so I'm okay with adding this option back in.

I just made a couple of comments!

lib/settings/pages/general_settings_page.dart
micahmo: Good idea! I can see this being used for other things in the future (like a migration of compact mode to an enum). Done!

PR #1059, opened 2024-01-15
Add CI workflow to check for sorted translations

hjiangsu:
## Pull Request Description

This PR adds a workflow which checks if the translation entries are sorted in alphabetical order. It also removes this workflow: "Create Empty .env File" since its no longer needed.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Very cool!

We could also have a PR (like the instances one) which just opens a PR to fix this rather than failing existing PRs.

hjiangsu:
> We could also have a PR (like the instances one) which just opens a PR to fix this rather than failing existing PRs.

Should be possible! I'm just not too sure how it interacts with Weblate (whether it'll detect merge errors, etc) which is why I'm keeping it as part of the CI for now

PR #1062, opened 2024-01-16
Add ability to clear image cache on startup, and manually

hjiangsu:
## Pull Request Description

This PR introduces a way to automatically clear image cache on startup. One issue that I've noticed is that the app size grows over time since image cache gets stored on the device. Because of this, the size of Thunder can grow quite large over a short period of time. This PR introduces two things:
- A check at the start of the app to clear any old image cache (by default, this is set to image cache data older than 7 days but we may be able to make this configurable in the future)
- An option in Settings -> Debug to see the size of the image cache, and manually clear it. This will clear all the cache as opposed to the cleanup at the start of the app.

I'm not too sure if the automatic check will happen properly as I'll have to run it over the next little bit in order to monitor app size (without uninstalling/reinstalling the app).

@micahmo Could you also do a check to see if it works on Android? As part of this, let me know what the system reports back, and what Thunder reports back for the cache size. On iOS, the image cache size and the system size differs, but that might be due to other temporary files.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings
simulator_screenshot_DDE24F7E-9CB5-445E-882E-D90BFAA4D0BD

## Checklist

### DIFF
--- a/lib/settings/pages/debug_settings_page.dart
+++ b/lib/settings/pages/debug_settings_page.dart
@@ -2,18 +2,25 @@ import 'package:flutter/material.dart';
 
 import 'package:path/path.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:extended_image/extended_image.dart';
 import 'package:sqflite_common_ffi/sqflite_ffi.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
-import 'package:thunder/shared/dialogs.dart';
 
+import 'package:thunder/shared/dialogs.dart';
 import 'package:thunder/shared/snackbar.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
 import 'package:thunder/settings/widgets/settings_list_tile.dart';
+import 'package:thunder/utils/cache.dart';
 
-class DebugSettingsPage extends StatelessWidget {
+class DebugSettingsPage extends StatefulWidget {
   const DebugSettingsPage({super.key});
 
+  @override
+  State<DebugSettingsPage> createState() => _DebugSettingsPageState();
+}
+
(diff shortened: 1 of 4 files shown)


lib/settings/pages/debug_settings_page.dart
```diff
@@ -114,6 +121,33 @@ class DebugSettingsPage extends StatelessWidget {
               ),
             ),
           ),
+          const SliverToBoxAdapter(child: Divider(indent: 16.0, endIndent: 16.0)),
```
micahmo, line 124: Should we make this divider more like the other one we have in the appearance settings (or make that one like this one)?

https://github.com/thunder-app/thunder/blob/eade1f7a7f2c2167d31a4d04f889e357b43876fe/lib/settings/pages/appearance_settings_page.dart#L42-L50

micahmo (approved):
This is really cool!

I tested on Android and saw similar discrepancies between what Thunder and the OS report, but I'm sure there are other things going on, like you said.

| Before | After |
|--------|--------|
| image | image |
| image | image | 

Somewhat related: at some point in the future I'd like to introduce an advanced setting that allows for more aggressive image caching, for those who are willing to sacrifice more memory/disk usage in favor of performance. Just throwing that out there as something I had mind. :blush:

lib/settings/pages/debug_settings_page.dart
hjiangsu: Done! I've matched it to be the same as the appearance page

hjiangsu:
> I tested on Android and saw similar discrepancies between what Thunder and the OS report, but I'm sure there are other things going on, like you said.

I believe its other cached/temp files that are the discrepancies in this case. I do think the image cache plays the largest role in increasing the overall app size though, which should be a huge improvement! (casually scrolling through the main feed in compact mode for a few minutes brought my cache size to over 200MB depending on the amount of images)

PR #1066, opened 2024-01-18
Refactor: refactor of thumbnail badges and compact post cards

hjiangsu:
## Pull Request Description

This PR refactors some of the existing logic with post cards. The reasoning for this is to reduce dependencies across widgets, and to clean up some duplicated code. This will be one of many refactors for post cards to help reduce code complexity.

Main changes performed:
- Removed unnecessary parameters for `PostCardViewCompact`. A lot of the current parameters that are passed in can be derived from `ThunderBloc`. As such, it's unnecessary to pass that data down the widget tree.
- Complete refactor of `TypeBadge` to reduce duplicated code. `TypeBadge` no longer has a dependency on PostViewMedia, which will help in certain situations. Rather, it will have a more generic parameter `MediaType` which will determine the look of the badge.
  - An upside of this refactor is that it will make it easier to implement #787!
- The thumbnail logic has been moved to a separate widget `ThumbnailPreview`. This helps with reducing duplicated code, and allows us to use `ThumbnailPreview` elsewhere in the app if needed.

Sidenote: Moderator post reports `/post/report/list` only returns back a `Post` rather than a `PostView`. Because of this, a lot of our existing widgets are not compatible with the response of moderator reports since they take in `PostViewMedia`. This is one of the initial triggers for refactoring the post cards.

e.g., an example response of `/post/report/list`
<details>

```json
{
  "post_report": {
    "id": 2975,
    "creator_id": 2290976,
    "post_id": 10174984,
    "original_post_name": "Images",
    "original_post_body": "Same instance pictrs: https://lemmy.ca/pictrs/image/4ee47284-643a-434a-9e38-1fb49317aad4.webp?format=webp\n\nDifferent instance pictrs: https://lemmy.world/pictrs/image/aa5c4b44-7b27-4564-8557-67177c1e6dbf.png?format=webp\n\nImgur: https://i.imgur.com/UCYJhaY.jpeg\n\nCatbox: https://files.catbox.moe/1no740.png",
    "reason": "Testing moderation reporting",
    "resolved": false,
    "published": "2024-01-17T18:38:03.738424Z"
  },
  "post": {
    "id": 10174984,
    "name": "Images",
[...]

PR #1067, opened 2024-01-19
Fix a few things changed by type badge refactor

micahmo:
## Pull Request Description

This PR fixes a few things changed by #1066.

1. The profile page should not dim posts (because the assumption is that everything there is already read).
   * [Here](https://github.com/thunder-app/thunder/pull/1066/files#diff-72c407b991667133d9fcc39dcf72dc8f83fb716db0f6027d0b20d8e99ded72bdL204) (and other places) the `indicateRead` parameter was removed. But we can't get this just from the `ThunderState` since some places override it. When adding this value back, I did make it optional so that we should have to pass it around less.
1. I reverted the change to the image icon. I'm finding that `image_rounded` is very heavy, especially in light mode where it's essentially dark on dark. `image_outlined` looks better to my eye! Of course if you disagree, I can put it back!

| Before | After |
|--------|--------|
| image | image | 

Sorry for not catching these!


hjiangsu (approved):
Looks good! Thanks for catching these (I didn't realize I changed the `image_rounded`)

PR #1068, opened 2024-01-19
Fix issue with cross-posts

micahmo:
## Pull Request Description

This is a small PR which fixes an issue where some cross-posts could fail to be initialized properly because the original post was stored in the state, not the widget.


ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1069, opened 2024-01-19
Add a notification page to display individual messages

micahmo:
## Pull Request Description

This PR is a followup to the recent notifications PR, which adds the ability to navigate directly to a specific reply from a notification. This is accomplished by pushing a new page onto the navigation stack, which is especially useful when the user has already navigated in the app to another page. Previously, opening a notification would scroll to the Inbox tab, but if there were another page open, that action would not be visible. Now a new page is pushed on top.

If the user clicks on a group notification, the page will show all unread replies. If the user clicks on a specific reply, the page will show only that reply (similar to opening an email). Also note that it will show that reply whether it is unread or not (in case the user happened to mark it read outside of Thunder). From there, the user can perform all the standard actions with the reply.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

#### Opening group notification

https://github.com/thunder-app/thunder/assets/7417301/df1855be-c169-4b96-ad4f-468979d525cd

#### Opening single notification

https://github.com/thunder-app/thunder/assets/7417301/f1c17dc6-6d0e-4015-9c58-764e870418b0

## Checklist


lib/thunder/pages/thunder_page.dart
```diff
@@ -399,191 +403,198 @@ class _ThunderState extends State<Thunder> {
       ],
       child: WillPopScope(
         onWillPop: () async => _handleBackButtonPress(scaffoldMessengerKey.currentState),
-        child: BlocListener<DeepLinksCubit, DeepLinksState>(
```
ggichure, line 402: Here you can switch to a  [MultiBlocListener](https://pub.dev/documentation/flutter_bloc/latest/flutter_bloc/MultiBlocListener-class.html)

micahmo: Good idea! Done!

ggichure (approved):

lib/thunder/pages/notifications_pages.dart
```diff
@@ -0,0 +1,57 @@
...
+                flexibleSpace: const FlexibleSpaceBar(titlePadding: EdgeInsets.zero),
+                pinned: true,
+                title: ListTile(
+                  title: Text(l10n.inbox, style: theme.textTheme.titleLarge),
+                  subtitle: Text(l10n.reply(replies.length)),
+                ),
+              ),
+              SliverToBoxAdapter(
+                child: Material(
+                  child: Transform(
```
hjiangsu, line 42: Just curious but what is the `Transform` for?

lib/utils/notifications.dart
```diff
@@ -120,3 +125,48 @@ Future<void> pollRepliesAndShowNotifications() async {
   // Save our poll time
   prefs.setString(_lastPollTimeId, DateTime.now().toString());
 }
+
+void navigateToNotificationReplyPage(BuildContext context, {required int? replyId}) async {
```
hjiangsu, line 129: Would it be good to place this in a separate file? Similar to `navigate_comment.dart` or `navigate_instance.dart`?

hjiangsu (approved):
Looks good to me! Just a couple of comments.

I'll trust that it works on Android since I can't test it out myself!

lib/thunder/pages/notifications_pages.dart
micahmo: I found that there was a space between the `SliverAppBar` and the `InboxRepliesView`. If I use another widget, there's no space, so I don't think it's related to the app bar. But there is also no padding defined in the `InboxRepliesView` and seemingly nothing above the first divider/reply. I worked on tracking it down for a while but got stuck, hence this. Let me know if you have any other ideas!

lib/utils/notifications.dart
micahmo: Sure! Done!

lib/thunder/pages/notifications_pages.dart
hjiangsu: Is it possible that the `Column` in `InboxRepliesView` is causing this? Try to add this to the `Column`:

```dart
crossAxisAlignment: CrossAxisAlignment.start,
mainAxisAlignment: MainAxisAlignment.start,
```

micahmo: Just tried, that doesn't help. 😞  Here's what it looks like for reference.

https://github.com/thunder-app/thunder/assets/7417301/cf75ba12-e450-47a6-8e66-2bccd3a63673

hjiangsu: Could you try adding `padding: EdgeInsets.zero` to ListView.builder in `InboxRepliesView`? 

> By default, [ListView](https://api.flutter.dev/flutter/widgets/ListView-class.html) will automatically pad the list's scrollable extremities to avoid partial obstructions indicated by [MediaQuery](https://api.flutter.dev/flutter/widgets/MediaQuery-class.html)'s padding. To avoid this behavior, override with a zero [padding](https://api.flutter.dev/flutter/widgets/BoxScrollView/padding.html) property.

It looks like `ListView` may be adding padding to the top. Usually, we would use `SliverList` for lists in `CustomScrollView` but we can't do this in this case since it's being shared. For reference: https://api.flutter.dev/flutter/widgets/ListView-class.html

micahmo: That did it! And it doesn't mess up the Inbox page either. Thank you so much for tracking that down!

hjiangsu (approved):
LGTM now!

PR #1078, opened 2024-01-22
Fix an issue with loading favorites on startup

micahmo:
## Pull Request Description

This PR fixes an issue where initial usage of things like the community autocomplete and search page do not include favorites until the drawer is opened once.

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/thunder-app/thunder/assets/7417301/1e2bc288-ad7c-4699-bae3-f152a0cd9fc5

#### After

https://github.com/thunder-app/thunder/assets/7417301/89d78c3e-b613-49be-8e4e-1b1ce7eb9655

## Checklist

### DIFF
--- a/lib/account/bloc/account_bloc.dart
+++ b/lib/account/bloc/account_bloc.dart
@@ -43,9 +43,9 @@ class AccountBloc extends Bloc<AccountEvent, AccountState> {
   }
 
   Future<void> _refreshAccountInformation(RefreshAccountInformation event, Emitter<AccountState> emit) async {
-    add(GetAccountInformation());
-    add(GetAccountSubscriptions());
-    add(GetFavoritedCommunities());
+    await _getAccountInformation(GetAccountInformation(), emit);
+    await _getAccountSubscriptions(GetAccountSubscriptions(), emit);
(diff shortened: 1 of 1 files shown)


lib/account/bloc/account_bloc.dart
```diff
@@ -43,9 +43,9 @@ class AccountBloc extends Bloc<AccountEvent, AccountState> {
   }
 
   Future<void> _refreshAccountInformation(RefreshAccountInformation event, Emitter<AccountState> emit) async {
-    add(GetAccountInformation());
-    add(GetAccountSubscriptions());
-    add(GetFavoritedCommunities());
+    await _getAccountInformation(GetAccountInformation(), emit);
+    await _getAccountSubscriptions(GetAccountSubscriptions(), emit);
+    await _getFavoritedCommunities(GetFavoritedCommunities(), emit);
```
micahmo, line 48: The problem here is that adding these things as events causes them to execute asynchronously, which creates a race condition (specifically because calculating the favorites requires having all of the subscriptions first). By specifically invoking and awaiting, we ensure that everything executes in the right order.

hjiangsu (approved):
LGTM, nice catch!

PR #1085, opened 2024-01-25
Fixed issue where feed FAB can be triggered on other pages

hjiangsu:
## Pull Request Description

This is a small PR which fixes an issue where the feed FAB can be triggered from another screen. This is because the FAB is still on screen, but is just invisible.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
LGTM.

Does the FAB still appear on communities that are opened from another page?

hjiangsu:
> Does the FAB still appear on communities that are opened from another page?

It still works when navigating to a community on a new page since that uses a separate feed FAB!

PR #1095, opened 2024-01-29
DRAFT: Add initial implementation for iOS push notifications using APNs

hjiangsu:
## Pull Request Description

This is a draft PR which adds the ability to handle apple push notification service (APNs).

To do this, I'm using a package `push` which handles platform specific implementations for iOS and Android. For iOS, it uses APNs (apple push notification service). For Android, it uses FCM (Firebase Cloud Messaging). I'm only implementing the iOS platform specific implementation here. Ideally, we can ignore the Android specific portions, and use UnifiedPush for Android devices. Whether or not this is possible is up in the air as we would have multiple packages handling push notifications on Android (may clash with each other).

Additionally, implementing push notifications requires a separate server in order to perform notification polling and sending out push notifications for devices. I've set up a simple server to handle this for now, and the domain is https://thunderapp.dev.

There is still a lot of work required for this to function properly. I'll add a to-do list here:
- [ ] Set up endpoint on the server to store device token and JWT token. The device token is required in order for APNs to work since it needs to know which device to send push notifications to. The JWT token is required to poll for new notifications. Ideally, Lemmy can implement a better method authentication so that it does not require the JWT token.
- [ ] Setup encryption for device token and JWT. We want to make sure that both the device token and JWT are encrypted through transit to prevent malicious actors from accessing the information. I'll need to perform more research on this and understand the implications of it.
- [ ] Setup server to perform periodic polling for new notifications, and send out push notifications. This step should be relatively straight-forward.

While I'm developing the server code, I'll keep it closed-source. I'll most likely make it open-source once I'm done with all the local testing (to ensure that no keys are accidentally committed to the repository)

Another note: I've shifted the notification logic and moved it from 'main` into `ThunderApp`. I hope this still works with the local notification logic!

## Issue Being Fixed

Issue Number: #219

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/d741b0b9-8444-4e8f-86cb-ecfec2c1a40e

## Checklist

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -31,6 +31,8 @@ PODS:
     - FlutterMacOS
   - permission_handler_apple (9.1.1):
     - Flutter
+  - push_ios (0.0.1):
+    - Flutter
   - receive_sharing_intent (0.0.1):
     - Flutter
   - share_plus (0.0.1):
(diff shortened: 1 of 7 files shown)


micahmo (commented):
Exciting!!

I'll try to do a more thorough review and also re-test on Android when ready!

hjiangsu:
Closing this draft in favour of #1237.

PR #1098, opened 2024-01-30
Add notification check on startup

micahmo:
## Pull Request Description

This is a super small PR which adds a notification check on startup. This should help to avoid scenarios where you may start the app, see that you have messages, and then receive notifications for them at some time later. It ensures that once the app is launched, everything is "up-to-date".


ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1064, opened 2024-01-18
Fix exit confirmation snackbar

micahmo:
## Pull Request Description

This is a small PR which fixes an apparently recent regression where the exit confirmation snackbar would not appear (most likely due to all the scaffold changes we've made recently). I also tweaked the wording slightly because I have found it a bit confusing. 😊

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/293bc360-33d4-48c0-93ed-2a5e7d7a199f

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1044, opened 2024-01-12
Navigate to new post after creation

micahmo:
## Pull Request Description

This PR adds the ability to navigate to a new post directly after creation. This is especially useful when creating from the main feed, because the new post won't appear in the feed.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

This demo illustrates that the snackbar is shown and navigation to the post works after creating a new post from...
* The main feed FAB
* Cross-post menu
* Cross-post widget
* External share

https://github.com/thunder-app/thunder/assets/7417301/47c01a18-f91d-4a0e-a088-82aca7e342fc

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1102, opened 2024-01-31
Update snackbars

micahmo:
## Pull Request Description

- We currently have lots of issues showing snackbars on various pages. Sometimes things break and we lose them ([#1064](https://github.com/thunder-app/thunder/pull/1064)). Sometimes we have to create and pass around lots of scaffolds in order for them to appear on the right page when triggered from another page ([#1044](https://github.com/thunder-app/thunder/pull/1044/files#diff-a6ee13719781c85254549655f57fb08a1e9b043efd2274202635504c9c0e27c8) and others). Sometimes they appear underneath what you're viewing. (We introduced a whole `customState` argument for the snackbar method to get around some of this stuff, and it still doesn't always work.) Sometimes the snackbar jumps position depending on whether the bottom nav bar is present. The summary is that snackbars are hard to get exactly right. I'm also not a huge fan of the appearance (where it covers the whole bottom section), which I believe was only changed to fix an issue, not because it looked better!
- This PR switches our snackbar usage to [GetX](https://github.com/jonataslaw/getx). Their tagline is "Open screens/snackbars/dialogs/bottomSheets without context", which immediately grabbed my attention!
- We might even be able to use GetX for other things in the app, like navigation or simple state management, although its [future is a bit unknown](https://github.com/jonataslaw/getx/issues/2889). I think it's worth using just for snackbars though!

> One downside of this approach is that the snackbar now covers things like the FABs, instead of them intelligently jumping up in response. However, I think this is a reasonable tradeoff for more reliable snackbars, and I think sometimes the jumping was confusing anyway (when I was about to press a button, it would move).

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Here are a couple specific before and after videos demonstrating the improved usability of snackbars. But there should be many other scenarios that this change improves.

#### Before: Drafts

https://github.com/thunder-app/thunder/assets/7417301/36be03e3-3497-4bcc-a863-d9bb8f75ee58

#### After: Drafts

https://github.com/thunder-app/thunder/assets/7417301/59a4a8cf-e114-42f7-842d-b7440b4d72a9

#### Before: Error

https://github.com/thunder-app/thunder/assets/7417301/27c05722-b3c2-46d8-a15f-04d9441cd5e1

#### After: Error

https://github.com/thunder-app/thunder/assets/7417301/bc16fce8-966b-4829-9668-f8ef23471049

#### Before: Navigate to Post

> In this before example, the snackbar only works when the new post is initiated from the home feed.

https://github.com/thunder-app/thunder/assets/7417301/3959b3eb-f62e-450f-8bda-af8f4e13be48

#### After: Navigate to Post

https://github.com/thunder-app/thunder/assets/7417301/d4bb1f75-376e-4d77-ad50-2a21e58d238d

#### Before: Block

> In this before example, the snackbar only appears once we've navigated back to the home feed.

https://github.com/thunder-app/thunder/assets/7417301/e5788300-12e6-44e0-b0f8-912437b3c78a

#### After: Block

https://github.com/thunder-app/thunder/assets/7417301/f059eeba-ebbc-4b35-a977-3aafd2df5609

#### Dark Mode

https://github.com/thunder-app/thunder/assets/7417301/7fcf4e53-2250-4c76-a6e5-3d76dc37f0d2

These are all of the examples I can come up with right now, but I know I've seen the snackbar do other funky stuff!

## Checklist

### DIFF
--- a/lib/shared/snackbar.dart
+++ b/lib/shared/snackbar.dart
@@ -22,15 +22,15 @@ void showSnackbar(
     duration: duration ?? Duration(milliseconds: max(4000, 1000 * wordCount)), // Assuming 60 WPM or 1 WPS
     messageText: Text(
       text,
-      style: const TextStyle(color: Colors.white),
+      style: TextStyle(color: theme.colorScheme.onInverseSurface),
     ),
     icon: leadingIcon != null
         ? Icon(
             leadingIcon,
-            color: leadingIconColor,
+            color: leadingIconColor ?? theme.colorScheme.inversePrimary,
           )
         : null,
-    backgroundColor: backgroundColor ?? const Color(0xFF303030),
+    backgroundColor: backgroundColor ?? theme.colorScheme.inverseSurface,
     mainButton: trailingIcon != null
         ? SizedBox(
             height: 20,
@@ -55,7 +55,7 @@ void showSnackbar(
     borderRadius: 6.0,
     boxShadows: [
       BoxShadow(
-        color: (backgroundColor ?? const Color(0xFF303030)).withOpacity(0.5),
+        color: (backgroundColor ?? theme.colorScheme.inverseSurface).withOpacity(0.5),
         blurRadius: 5,
         offset: const Offset(1, 1),
       ),
--- a/lib/shared/snackbar.dart
+++ b/lib/shared/snackbar.dart
@@ -1,11 +1,11 @@
 import 'dart:math';
 
 import 'package:flutter/material.dart';
+import 'package:get/get.dart';
 
 void showSnackbar(
   BuildContext context,
   String text, {
-  ScaffoldMessengerState? customState,
   bool clearSnackBars = true,
   Duration? duration,
   Color? backgroundColor,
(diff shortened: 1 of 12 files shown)


lib/shared/snackbar.dart
```diff
@@ -15,55 +15,65 @@ void showSnackbar(
   IconData? trailingIcon,
   void Function()? trailingAction,
 }) {
-  int wordCount = RegExp(r'[\w-]+').allMatches(text).length;
-  SnackBar snackBar = SnackBar(
+  final ThemeData theme = Theme.of(context);
+  final int wordCount = RegExp(r'[\w-]+').allMatches(text).length;
+
+  GetSnackBar snackBar = GetSnackBar(
```
ggichure, line 21: Why not use the default `snackbar `?

micahmo: Please see the PR description for my rationale behind this change! Especially watch the before/after videos, and see how the code is simplified without having to worry about context and scaffolds! 😊

lib/shared/snackbar.dart
```diff
@@ -15,55 +15,65 @@ void showSnackbar(
...
-        Expanded(
-          child: Text(
-            text,
-          ),
-        ),
-        if (trailingIcon != null)
-          SizedBox(
+          )
+        : null,
+    backgroundColor: backgroundColor ?? const Color(0xFF303030),
```
ggichure, line 33: For the background color [M3 guidelines](https://m3.material.io/components/snackbar/specs#200c7ca0-830a-470e-afb4-4e3e3fef45c9)  recommends `onInverseSurface`

micahmo: I think you mean `inverseSurface` for the background and `onInverseSurface` for the foreground. That looks fine to me, so I updated.

| Light | Dark |
|--------|--------|
| image | image |

micahmo:
> Hmm, I'm a bit divided about making this change.

That's perfectly reasonable! I think I will respectfully disagree with your points (explained below), but of course the final call is yours and I'm happy with whatever you decide!

> We might be adding this package to solve an issue that could be solved by some simple architectural changes (one example I can think of is placing the `SnackBar`s in more appropriate locations based on where they should be shown, and derive the proper `context` to use)

I'm not convinced that we can ever have as elegant of a solution, even with a rearchitecture, as long as we're relying on scaffolds. I think there are too many different scenarios to account for and some we might not even think of. For example, I just hit a situation where I was viewing an image, then I got a notification which opened a page, then I navigated to a post, then a community, then I made a new post. The "navigate to post" snackbar appeared all the way back on the main feed under the image, which I just happened to catch. It was already hard enough to make the create post screen show a snackbar on the main feed, and that's really the only scenario where it works. Unless we can replicate their technique, I don't think anything will be as reliable or convenient as GetX.

> It feels to me like GetX is quite controversial amongst the community (some like it, some hate it, etc). Because of this, it can lead to situations where developers with strong opinions of the package might be reluctant from contributing.

If you're right, this would be very unfortunate! From what I've seen, most contributors are not Flutter experts and are happy to learn how we've done things and follow our patterns. Also if we isolate our usage to snackbars, then most of their objections shouldn't apply.

> Additionally, GetX seems to be a bit overkill for the issue we're attempting to solve as it attempts to solve a lot of things (state management, dependency injection, routing, localization, theming, etc.).

I disagree that it's overkill, as long as we're not using the additional functionality. If it required a huge refactor just to use their snackbars, maybe. But the only other thing I really had to touch was adding a key to our router. Just because the package can do more, doesn't mean it's overkill to use the part we want (unless you mean that by adding such a big package to our project, we'll increase the release size or something).

> While this can be good in certain situations, it also has drawbacks because it could cause a situation where we're mixing GetX solutions with other solutions.

I think if we decide not to use GetX for any of the other things that it can do, then we won't run the risk of mixing solutions. It doesn't seem like the kind of thing that can get in accidentally, so we would purposefully not add it ourselves or allow it in PRs. Contributors wouldn't even necessarily know it's there.

And with all that said, I'll leave it at that and let you make the final call. 😊 I'm happy to work on a different solution, especially if we can find something that is equally reliable.

ggichure:
> Instead of making this change, I'd like to give it a shot at re-architecturing some parts of the app to solve the issues with SnackBars, and make this change as a last resort if all else fails! Thoughts?

This will do. 

Also toasts will  always display on top and don't depend on scaffolds.

micahmo:
> Also toasts will always display on top and don't depend on scaffolds.

We could use toasts for some things! Unfortunately I don't believe they're interactable or super customizable.

micahmo:
That mockup looks great, thanks for trying that scenario! Does the snackbar stay on top if you navigate back through the stack?

Also do you think we can make it floating again? I think that looks much better and matches MD3 guidelines. I can't remember why we changed, but I feel like it was related to this issue.

Issue #1007, opened 2023-12-24
Mod Tools


orangeNgreen:
**Is your feature request related to a problem? Please describe.**

This request is not related to a problem. It’s a QOL feature request for admins and community moderators. Right now, to “pin” and “unpin” posts, lock threads, or delete posts/comments, etc., mods and admins need to either navigate to the browser or to another app that includes moderator tools. 

**Describe the solution you'd like**

Add moderator/admin tools: pin/unpin posts, see reported posts and comments, lock threads, delete posts and comments, etc. 

**Describe alternatives you've considered**

Right now I use Voyager for moderator stuff. But I prefer Thunder’s UI. So a one-stop-shop would be awesome!

**Additional context**

Nope, but happy to answer questions if I’m able.

hjiangsu:
Thanks for creating this! I think we can reuse this issue as an umbrella issue since it covers a lot of individual feature requests.

We can create smaller feature requests tackling on different parts of moderator actions so that others can potentially pick it up.

hjiangsu:
Just an update here - locking, removing, and pinning posts are now available in the latest pre-release. Other moderator activities (such as viewing reports and resolving reports, moderator feed, etc) will likely come in a future PR!

PR #1104, opened 2024-01-31
Initial moderator post actions (lock, pin, remove)

hjiangsu:
## Pull Request Description

This is a PR which introduces moderator actions to the post actions bottom sheet. Currently, it includes **locking, pinning, and removing posts**.

~~This is still in draft since there are a couple of things that still need to be addressed.~~
- [x] Set up a way to dynamically change the text of the action (from lock post to unlock post)
- [x] Clean up localizations and changes

## Issue Being Fixed

Issue Number: Partially solves https://github.com/thunder-app/thunder/issues/1007

## Screenshots / Recordings

Update (added trailing icon to moderator option)
Simulator Screenshot - iPhone 15 Pro - 2024-02-08 at 11 24 54

https://github.com/thunder-app/thunder/assets/30667958/2e6446c4-0032-4e57-87cf-fe9b56436c9b

## Checklist


micahmo:
I know this isn't ready for review yet, but I just wanted to comment that I like the overflow menu for moderator actions! I've been wanting to add some new actions to the long-press menus as well, as had been thinking about how to reorganize it. I'm thinking, for example, community, user, and instance actions can all overflow rather than having two entries each.

Should there be an arrow on the right-hand side like the "Top" sort menu item?

hjiangsu:
This is now ready for review! I've updated the body of the PR to include screenshots and a video

micahmo (approved):
Wow, this is so awesome!

The only usability question I had was about the "Reason" prompt. I do know we have a similar prompt for reporting comments, but it uses a modal. I was wondering if either that mechanism should use this dialog or vice versa. I was also wondering if the text input should be bigger by default? Or maybe a better question would be to ask what happens when you enter a lot of text? Does it grow vertically or overflow horizontally?

hjiangsu:
> The only usability question I had was about the "Reason" prompt. I do know we have a similar prompt for reporting comments, but it uses a modal.

Ahh good catch! I forgot that the report comment used a modal rather than dialog. I can try to adjust the logic on my end to do something similar to the comment reporting!

hjiangsu:
I created a new widget to hold the inline text-input. Right now, it'll be used for the mod removal but we can extend that to post report and comment report in the future!

One caveat with this right now is that cancelling the action closes the modals (does not bring you back to the options)

I also adjusted it a bit to match more with the current bottom sheet modal. Thoughts?

https://github.com/thunder-app/thunder/assets/30667958/9de2724d-5944-4270-8852-f6cc59a5fcf9

micahmo:
Looks great! Nice to immediately focus the keyboard as well!

hjiangsu:
If this is good, I'll go ahead and merge this in?

micahmo:
Go for it!

Issue #1101, opened 2024-01-30
Cache the feed contents


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
I noticed that when I open a post, and then return back to the feed, the whole feed flickers -- I assume that this is it reloading/re-fetching the feed contents. This causes me to think that the contents are not being cached.

**Describe the solution you'd like**
Cache the feed contents, so that the feed doesn't need to be completely reloaded.

**Describe alternatives you've considered**
None.

**Additional context**
I'm not sure if this is exactly what the problem is, so this feature request is rather malleable.

micahmo:
@K4LCIFER Any chance you could provide a screen recording of what's flickering? Is it mostly the images? If so, we have some plans for extending image cache time.

K4LCIFER:
Yeah it was the images that were flickering. I can still send a video if you would like!

micahmo:
Sure, thank you1

K4LCIFER:
> Sure, thank you!

Hrm, I appear to not be able to reproduce it now. Perhaps it was a symptom of me being on a slower network when I first experienced it?

micahmo:
It comes and goes for me too. I opened #1105, but I have kept it in draft mode because I haven't been able to prove definitively that it helps. 😆

PR #1105, opened 2024-01-31
Improve image caching

micahmo:
## Pull Request Description

This PR is an attempt to improve image caching. This has been a long-running discussion (see a few of the linked PRs below).
 - [#463](https://github.com/thunder-app/thunder/pull/463)
 - [#576](https://github.com/thunder-app/thunder/pull/576)
 - [#745](https://github.com/thunder-app/thunder/pull/745)

The most recent solution was to cache images with a timeout. This means that scrolling through the feed is nice, but navigating into a post for a while and then navigating back can still cause all the images to reload. In fact, this behavior recently prompted a new issue, [#1101](https://github.com/thunder-app/thunder/issues/1101).

Since it seems that there's no perfect solution which will both keep images cached and improve performance while also maintaining a low memory profile, I've added a new advanced setting which lets the user pick their preferred behavior.

The new default behavior will be like prior to #745, meaning that images may reload as you scroll up and down the feed, but should not reload when navigating back from a post (`clearMemoryCacheWhenDispose` is `true`).

The new optional behavior is to have more aggressive caching than even in #745, more like after #463, where `clearMemoryCacheWhenDispose` is `false` *and* `cacheMaxAge` is not set.

> [!IMPORTANT]
> The important thing is that neither mode should cause the flashing described in #1101, because that was purely due to the `cacheMaxAge`. It was forcibly invalidating cache, even before the widget was disposed.

## Issue Being Fixed

Issue Number: #1101

## Screenshots / Recordings

> This demo just shows the new setting. You need to run the app for a while to experience the difference in behavior.

https://github.com/thunder-app/thunder/assets/7417301/073ff729-efe7-48f9-a301-9cda27681e10

## Checklist


micahmo:
Since this one is a bit difficult to test, I think the best feedback we can get is from regular usage in the nightlies. Therefore, I'm opening this up for review.

lib/settings/pages/general_settings_page.dart
```diff
@@ -641,6 +650,31 @@ class _GeneralSettingsPageState extends State<GeneralSettingsPage> with SingleTi
               ),
             ),
           ),
+          SliverToBoxAdapter(
+            child: Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 16.0),
+              child: ListOption(
+                description: l10n.imageCachingMode,
```
hjiangsu, line 657: Since #1086 has been merged in, we can add in `highlightKey` for search results here!

hjiangsu (approved):
Just took a look at the changes and it LGTM (with a minor comment)!

One thought - since this seems to only be happening on Android, should we limit the setting to Android only for now?

lib/settings/pages/general_settings_page.dart
micahmo: You bet! Done!

micahmo:
> since this seems to only be happening on Android, should we limit the setting to Android only for now?

Sure, I will add that restriction. It would be easy enough to allow the setting on iOS later if needed. The default should also be fine for memory utilization on iOS since it defaults to clearing cache on dispose.

micahmo:
Should be all set now!

PR #1106, opened 2024-02-01
Fix in-app browser URL issue

micahmo:
## Pull Request Description

This is a super small PR which fixes an issue with the in-app browser where sharing / opening in system browser would use the original URL, not the current URL that the user may have navigated to.


hjiangsu (approved):
LGTM!

PR #1107, opened 2024-02-01
Fix issue with deferred comments

micahmo:
## Pull Request Description

Since LemmyNet/lemmy#3965 was done, I thought we wouldn't have any more issues loading nested comments, but I still do every now and then. I tracked it down to a section of code I had added as error handling. To be honest, the conditions I was checking make sense. It's whether the number of comments we've loaded are more than the total reported by Lemmy. And in fact, in the scenario I'm testing, we do seem to have loaded all the comments by number. So I'm not sure exactly what's going on. But by removing the artificial exception, we are able to successfully load the deferred comments. Therefore, I'm going to remove this exception. We should have time in the nightlies to see if this brings back the infinite spinner at all.

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/thunder-app/thunder/assets/7417301/6c1856d6-7a69-44aa-a810-8b6802237620

#### After

https://github.com/thunder-app/thunder/assets/7417301/7d618228-8919-4546-943c-6a7da013ff0c

## Checklist

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -289,13 +289,9 @@ class PostBloc extends Bloc<PostEvent, PostState> {
             if (!state.hasReachedCommentEnd && state.commentCount == state.postView!.postView.counts.comments) {
               emit(state.copyWith(status: state.status, hasReachedCommentEnd: true));
             }
-            if (event.commentParentId != null) {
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM! Thanks for double checking this issue!

PR #1115, opened 2024-02-05
Fix spoiler titles in instance view

micahmo:
## Pull Request Description

This is a super small PR which applies the same fix as #1033 to the instance view.


hjiangsu (approved):
LGTM!

PR #1121, opened 2024-02-12
Only do startup check if not group notification

micahmo:
## Pull Request Description

This is a super small PR which fixes a subtle use case. On Android, notifications can be grouped together. When tapped, all of the child notifications are dismissed. In Thunder, when the user taps the group, we show all unread messages in the notifications page. This means that if a new message happened to come in since the notifications were displayed, they will still see these messages.

However, notification check that occurs on startup could create new notifications for these new messages, even though the user started the app to view the notifications page. In other words, it creates a scenario where the user is already reviewing what the want to see, and the new notifications become superfluous.

I hope that makes some sense. If not, I can try to word it differently!


hjiangsu (approved):
I think I understand! LGTM

PR #1086, opened 2024-01-25
Improve settings seearch results

micahmo:
## Pull Request Description

This PR continues to build on the recent settings search feature, and adds the functionality of scrolling to and flashing the selected search result.

I also tested each possible search result and fixed up some things that were missing/wrong.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/c7ef228a-23af-463e-a15a-b41356e3fd49

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1122, opened 2024-02-12
Fix settings padding

micahmo:
## Pull Request Description

PR #1086 introduced padding into several common settings widgets. Accordingly, the surrounding padding was removed from the usage of those widgets on settings pages. However, those widgets are also used on pages like the advanced share sheet and the account page. This PR simply makes a similar change to remove the unnecessary padding around those widgets. In cases where the padding was around a group of widgets, it was removed from the group and added to the ones that didn't already have it.

> Review without whitespace.


hjiangsu (approved):
LGTM. Thanks!

Issue #1123, opened 2024-02-13
0.2.9-4 does not start


cooperaj:
### Bug Description

On my Pixel 6 Pro running Android 14 I'm unable to start 0.2.9-4. The app hangs at the logo.

Additionally, attempting to rollback to 0.2.9-3 give me an error that the app package is invalid. I have to uninstall Thunder first then that release installs and functions as I'd expect. 

### Expected Behaviour

Should be seeing a shiny new 0.2.9-4 release. 

### Steps to Reproduce

Install new version.
Start it. 

### App Version

0.2.9-4

### Device

Pixel 6 Pro

### OS

Android 14

Fmstrat:
Same issue, Pixel 6 (non-pro).

machinaeZER0:
This happened to me as well! Pixel 4a over here

hjiangsu:
Hey, thanks for mentioning this. It seems like it's not limited to Android as its affecting some iOS users as well. I'll take a closer look at this and push out a fix when I can!

hjiangsu:
Update: I've released 0.2.9-5 which should fix this issue. If you could try it out and let me know, that would be great. Thanks!

https://github.com/thunder-app/thunder/releases/tag/0.2.9-5

Fmstrat:
Works for me. Thank you, quick action is much appreciated with your free time!

machinaeZER0:
All fixed on my end, thank you! 😄

hjiangsu:
Closing this issue since it seems to be fixed!

PR #1125, opened 2024-02-13
Fix issue with Thunder stalling on splash screen

hjiangsu:
## Pull Request Description

This PR fixes an issue where Thunder stalls on the splash screen when upgrading/updating from a previous version. The main issue here seems to be that the browser migration was not properly applied.

Instead of using `browserMode.name` as our setting string, we were using `browserMode.toString()`. This in turn has slightly different outputs. `browserMode.name` outputs the name of the BrowserMode enum whereas `browserMode.toString()` outputs the full enum name (e.g., `external` vs `BrowserMode.external`).

I've applied a fix for this (both with the original migration, and for the case where the migration was already applied but was using the improper setting), I've also made `performSharedPreferencesMigration` async to ensure that the migrations are applied before other code execution, and removed the splash screen when ThunderPage is at a failure state.

@micahmo I'll go ahead and merge this in first so that I can push out a hotfix release for this!

## Issue Being Fixed

Issue Number: #1123

## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -91,7 +91,7 @@ void main() async {
   LemmyClient.instance.changeBaseUrl(initialInstance);
 
   // Perform preference migrations
-  performSharedPreferencesMigration();
+  await performSharedPreferencesMigration();
 
   // Do a notifications check on startup, if the user isn't clicking on a group notification
   if (!startupDueToGroupNotification) {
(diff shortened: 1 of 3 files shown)


micahmo (commented):
Looks good, thanks so much for catching and fixing this!

PR #1130, opened 2024-02-16
Fix search page keyboard on iOS

hjiangsu:
## Pull Request Description

This PR fixes a small issue on the Search page when on iOS. The keyboard was previously set to `url`, which causes the iOS keyboard to switch over to a `url` styled keyboard. This causes an issue because there's no way to add a space in the search field.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
LGTM!

Issue #1053, opened 2024-01-14
Explore


trymeouteh:
**Is your feature request related to a problem? Please describe.**
No way to explore an instance or the fediverse.

**Describe the solution you'd like**
In the search tab, when you do not have anything typed in the search box, you can explore all of the communities, users, posts and comments.

You can filter this out with the filters by only exploring the local instance, the entire fediverse, etc.

**Describe alternatives you've considered**
None

**Additional context**
None

K4LCIFER:
> No way to explore an instance

I agree -- this would be a very useful feature.

hjiangsu:
> In the search tab, when you do not have anything typed in the search box, you can explore all of the communities, users, posts and comments.

I believe there is an endpoint which lists the communities (all, local, subscribed) for an instance via `/communities`. However, there is no endpoint that I know of which lists users or comments.

For posts, I believe they are already covered by using the Feed -> Local.

CTalvio:
I've been thinking that as well as Trending, the top five federated communities could be shown in search, and should have an entry to a list where you can just scroll a list of all federated communities.

Browsing by comments is right there in the default webUI, so it's certainly a thing: https://sopuli.xyz/?dataType=Comment&listingType=Subscribed&sort=Hot

micahmo:
Hey @trymeouteh, I just wanted to clarify your request. I believe the thing you're demonstrating, which we don't currently support, is the ability to see search results of all types mixed together (communities, users, posts, comments), without having to pick a specific type of thing to search. Is that right? If so, I believe that's covered by #1050. But please let me know if I misunderstood!

PR #1133, opened 2024-02-16
Add instance explorer

micahmo:
## Pull Request Description

This PR adds support for an "instance explorer" which allows viewing the communities, ~~users~~, posts, and comments federated with a given instance. This is part of the existing instance page, which can be accessed by long-pressing posts/comments and pressing "Visit Instance". When an entity is tapped, we use the `ResolveObject` endpoint to find it on the user's local instance so they can proceed to interact with it as normal.
For the current instance, this explorer can also be accessed via the search page, where there is a new "Explore instance" button at the bottom that will navigate to the same instance page.

Future enhancements:
* Add sorting to each view mode (currently everything is `TopAll`; however, note that the `/communities` page on the web UI  doesn't have sorting either, so we're not much worse :blush:).
* Ensure that long-press and swipe actions are supported. This will take some extra work because we don't always have the expected blocs in place, plus we're dealing with additional complexity of entities from other instances.

**NOTE:** After I finished working on this, I discovered that `ListingType.local` does not seem to work when the `Search` endpoint is called for `SearchType.users`. It always returns users from all instances. As a result I have disabled this section of the instance page. However, all of the code is there, and it is not commented out (just hidden with a conditional), so we can maintain it going forward and unhide if/when that endpoint is fixed. (The demos still include the user section.)

> Review without whitespace.

## Issue Being Fixed

Issue Number: #1053

### Demo 1: Exploring the current instance via the search page

https://github.com/thunder-app/thunder/assets/7417301/0aba38d6-e82b-45ce-b919-46f55a5359b2

### Demo 2: Exploring a remote instance (with object resolution to current instance)

https://github.com/thunder-app/thunder/assets/7417301/99551156-9ea6-4823-8656-8739cb1038d3

## Checklist


hjiangsu:
I haven't reviewed the code yet but this looks really good! Thanks for working on this, this will definitely make instance exploration more accessible within the app!

Just a note, I'll probably look at this after the next general release is out (just so that I can focus on fixing some existing issues and get the next version rolled out as soon as I can)

lib/instance/pages/instance_page.dart
```diff
@@ -0,0 +1,370 @@
...
+                          SliverToBoxAdapter(
+                            child: Padding(
+                              padding: const EdgeInsets.all(20),
+                              child: Material(
+                                child: InstanceView(site: widget.getSiteResponse.siteView.site),
+                              ),
+                            ),
+                          ),
+                        if (viewType == SearchType.communities)
+                          SliverToBoxAdapter(
```
hjiangsu, line 252: Is it possible to use `SliverList` or some form of a Sliver-based widget to generate the individual entries? (for communities, users, etc.)

The reasoning here is that with large lists, using `ListView.builder` and `shrinkWrap: true` causes performance issues (which is the reason I switched over to using Slivers in the FeedPage refactor). When using `shrinkWrap`, the whole list is kept in memory and Flutter is unable to remove parts of the list that are not visible to the screen. 

This won't be noticeable for small lists, but is more apparent in larger/infinite lists

https://api.flutter.dev/flutter/widgets/ScrollView/shrinkWrap.html

lib/comment/widgets/comment_list_entry.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:thunder/core/auth/bloc/auth_bloc.dart';
+import 'package:thunder/core/enums/local_settings.dart';
+import 'package:thunder/core/singletons/preferences.dart';
+import 'package:thunder/post/bloc/post_bloc.dart' as post_bloc;
+import 'package:thunder/post/pages/create_comment_page.dart';
+import 'package:thunder/shared/comment_reference.dart';
+import 'package:thunder/shared/snackbar.dart';
+import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+
+Widget buildCommentEntry(
```
hjiangsu, line 20: Would it be possible to convert this to a Stateless/Stateful Widget? Using Stateless/Stateful widgets are usually more performant than using a function that returns a widget! This applies to the comment, community, user, etc.

https://www.youtube.com/watch?v=IOyq-eTRhvo

hjiangsu (commented):
Just a couple points of discussion!

One more general point would just be to add some more comments on top of the widget to explain what its being used for (e.g., `buildCommentEntry` is used to display a preview of a comment for the search/instance explorer)

micahmo:
> add some more comments on top of the widget to explain what its being used for

Done!

lib/instance/pages/instance_page.dart
micahmo: I tried changing from
* `SliverToBoxAdapter` -> `Material` -> `ListView.builder` -> `<widget>`

to

* `SliverList` -> `SliverChildBuilderDelegate` -> `Material` -> `<widget>`

and it seems to work fine. Is that what you had in mind?

c6831c68e3325741e4a8831faaed348b69ea43fe

hjiangsu:
> Looks amazing! Please add an option to explore communities, posts, comments, users on the entire fediverse as well, not just a specific instance.

I don't believe that the Lemmy API gives us direct access to this information, so this might be more related to https://github.com/thunder-app/thunder/issues/508

lib/comment/widgets/comment_list_entry.dart
micahmo: > Using Stateless/Stateful widgets are usually more performant than using a function that returns a widget!

Good to know! Done!

48313fb964e11194850865ec775de6fdddf2262d

> https://www.youtube.com/watch?v=IOyq-eTRhvo

Very informative, thank you!

micahmo:
> > Looks amazing! Please add an option to explore communities, posts, comments, users on the entire fediverse as well, not just a specific instance.
> 
> I don't believe that the Lemmy API gives us direct access to this information, so this might be more related to #508

We may be able to use a similar technique as I'm using here. I am just calling `Search` without a query string, and it returns "everything". In this case, I am filtering by `Local`, but we could also filter by `All` to see "everything". Note that this would only be things federated with the current instance, so not really _everything_, but maybe good enough for the feature being requested.

Either way, that would be a separate change and may also make more sense on the search page, like how we have trending communities now.

hjiangsu (approved):

lib/instance/pages/instance_page.dart
hjiangsu: Yup, that looks good!

hjiangsu:
Just double checking - is this ready to be merged in?

micahmo:
Yes, should be good to go!

hjiangsu:
> After I finished working on this, I discovered that ListingType.local does not seem to work when the Search endpoint is called for SearchType.users. It always returns users from all instances. 

Linking this issue since it's relevant to what you mentioned (seems like it'll be fixed in an upcoming release): https://github.com/LemmyNet/lemmy/issues/4146

PR #1090, opened 2024-01-26
Remove usesCleartextTraffic attribute on AndroidManifest

hjiangsu:
## Pull Request Description

This PR removes `usesCleartextTraffic` from AndroidManifest as mentioned in https://github.com/thunder-app/thunder/issues/1060. We'll release this to a nightly build and see if there are any issues with removing this attribute.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1060

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1152, opened 2024-02-28
Fix in-app browser issue

micahmo:
## Pull Request Description

This PR fixes an issue when using the in-app browser where URLs with schemes intended for other apps are attempted to be handled as web pages, which fails. The fix is to ensure the protocol is one we can handle, otherwise we redirect to the "system" browser (which may be another app on the device). I also took this opportunity to exclude `http`, which would probably not work now due to #1090. We can always kick http links to the external browser and let it handle them. Since this might be an Android-only problem, I did add an OS check. If you can think of a similar test case for iOS, let me know.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/9de774ef-faec-4879-acc5-216df2e2579b

### After

https://github.com/thunder-app/thunder/assets/7417301/e4ba7578-a45a-43b8-a627-4af9defd0aad

## Checklist

### DIFF
--- a/lib/shared/webview.dart
+++ b/lib/shared/webview.dart
@@ -1,6 +1,8 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:back_button_interceptor/back_button_interceptor.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:share_plus/share_plus.dart';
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
Interesting! LGTM

PR #1149, opened 2024-02-27
Fix community search issues

micahmo:
## Pull Request Description

This PR includes a few changes for the community search page.
1. Focus the search box immediately. Unlike the main search page, there is no other content here, so I think it makes sense to jump right to searching.
2. Fix an issue with search state being lost when navigating back via the Android gesture.
   * I discovered that, when using the Android back button, the `SwipeablePageRoute` will rebuild itself, causing the blocs that it creates to be reconstructed from scratch. The fix is simply to construct the bloc outside of the `SwipeablePageRoute`. (This could possibly help with solving [#1109](https://github.com/thunder-app/thunder/pull/1109) as well!).

## Issue Being Fixed

Issue Number: N/A

### Issue 1

https://github.com/thunder-app/thunder/assets/7417301/2ae30f96-81e9-45cd-bb02-a4b689e6e8cb

### Issue 2 Before

https://github.com/thunder-app/thunder/assets/7417301/2fdaf62c-8f34-434d-a1bb-db5e3dc7f57a

### Issue 2 After

https://github.com/thunder-app/thunder/assets/7417301/df6b251f-db61-48e0-beac-ec963503b795

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1153, opened 2024-02-28
Fix notifications page state issue

micahmo:
## Pull Request Description

This PR is the culmination of a lot of effort to fix an issue that's been really bugging me lately. See #1109 for the full history.

Long story short, I was finally able to reproduce this issue consistently! None of my previous attempts to fix were related to the problem. In fact, the problem only occurred when using the Android back button, which is probably why I happened to see it on my physical device but not the emulator, where it's a bit clunky to use the gesture. It also turns out to be the exact same problem as #1149. I still can't explain *why* using the gesture causes the `SwipeablePageRoute` to completely rebuild, but at least I know how to fix it now.

This PR will supersede #1109, which can probably be closed now.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/c67a5f09-1cc4-4f3a-8409-96d36f0f3d08

### After

https://github.com/thunder-app/thunder/assets/7417301/5e40a98e-b3d5-49e2-9258-80852c7579da

## Checklist

### DIFF
--- a/lib/utils/notifications_navigation.dart
+++ b/lib/utils/notifications_navigation.dart
@@ -40,6 +40,8 @@ void navigateToNotificationReplyPage(BuildContext context, {required int? replyI
   }
 
   if (context.mounted) {
+    final NotificationsReplyPage notificationsReplyPage = NotificationsReplyPage(replies: specificReply == null ? allReplies : [specificReply]);
+
     Navigator.of(context)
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
Nice catch! LGTM 😄

PR #1156, opened 2024-02-29
Draft: Instance and community modlogs

hjiangsu:
## Pull Request Description

This draft PR introduces modlogs into Thunder. Most of the changes here shouldn't conflict with existing code since its _mostly_ new (with the exception of some logic which adds the modlog action to the app bar)

As of right now, instance and community modlogs have been implemented. In the future, we can add additional modlogs (e.g., user modlogs) but that is out of scope for this PR. To access the modlogs, I've added an additional action to the respective app bar (general feed, and community). General feeds will navigate to the instance modlog, whereas communities will navigate to the community modlog.

Notes:
- I had to make some minor changes to the Lemmy API because of wrong types: see https://github.com/thunder-app/lemmy_api_client/commits/fix/mod-ban-community/

Todo:
- [ ] Check to see that all the different moderation actions are included and parsed properly
- [ ] Have a way to distinguish between instance/community modlog
- [ ] Add more context to each mod action (e.g., link to the relevant post, comment, community, etc)
- [ ] Double check compatibility between 0.18 and 0.19. I'll try to mainly focus on 0.19 support as a lot of instances have migrated already, but I'll still do some checks to make sure it doesn't completely break on 0.18
- [ ] Fix icons with mod action filter

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

I'll post some initial screenshots once this is more polished!

## Checklist

### DIFF
--- a/lib/feed/widgets/feed_page_app_bar.dart
+++ b/lib/feed/widgets/feed_page_app_bar.dart
@@ -18,12 +18,14 @@ import 'package:thunder/feed/bloc/feed_bloc.dart';
 import 'package:thunder/feed/utils/community.dart';
 import 'package:thunder/feed/utils/utils.dart';
 import 'package:thunder/feed/view/feed_page.dart';
+import 'package:thunder/modlog/view/modlog_page.dart';
 import 'package:thunder/search/bloc/search_bloc.dart';
 import 'package:thunder/search/pages/search_page.dart';
 import 'package:thunder/shared/snackbar.dart';
 import 'package:thunder/shared/sort_picker.dart';
 import 'package:thunder/shared/thunder_popup_menu_item.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+import 'package:thunder/utils/swipe.dart';
 
 class FeedPageAppBar extends StatelessWidget {
   const FeedPageAppBar({super.key, this.showAppBarTitle = true, this.scaffoldStateKey});
@@ -120,6 +122,35 @@ class FeedPageAppBar extends StatelessWidget {
                   );
                 },
               ),
+              if (feedState.feedType == FeedType.general)
+                IconButton(
+                  icon: Icon(Icons.shield_rounded, semanticLabel: l10n.sortBy),
+                  onPressed: () async {
+                    HapticFeedback.mediumImpact();
+
+                    AuthBloc authBloc = context.read<AuthBloc>();
+
+                    ThunderBloc thunderBloc = context.read<ThunderBloc>();
+                    final ThunderState state = context.read<ThunderBloc>().state;
+                    final bool reduceAnimations = state.reduceAnimations;
+
+                    await Navigator.of(context).push(
+                      SwipeablePageRoute(
+                        transitionDuration: reduceAnimations ? const Duration(milliseconds: 100) : null,
+                        backGestureDetectionStartOffset: !kIsWeb && Platform.isAndroid ? 45 : 0,
+                        backGestureDetectionWidth: 45,
+                        canOnlySwipeFromEdge:
+                            disableFullPageSwipe(isUserLoggedIn: authBloc.state.isLoggedIn, state: thunderBloc.state, isPostPage: false) || !thunderBloc.state.enableFullScreenSwipeNavigationGesture,
+                        builder: (otherContext) {
+                          return MultiBlocProvider(
+                            providers: [BlocProvider.value(value: thunderBloc)],
+                            child: const ModlogFeedPage(),
+                          );
+                        },
+                      ),
+                    );
+                  },
+                ),
               if (feedState.feedType == FeedType.community)
                 PopupMenuButton(
                   itemBuilder: (context) => [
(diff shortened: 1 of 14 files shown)


micahmo:
This is so awesome!!

machinaeZER0:
Aw, I wish I could see the original implementation for comparison (there was a video here before, correct?). As a quicker take (on my phone), I think you've captured a lot of the functionality perfectly, but the button/pill setup feels like it could be done in a way to reduce duplication of information, if that makes sense? Being showing a blocked username in the body, then duplicating it in a pill below, showing a reply that requires action, then showing that text again in a pill below, etc. 

I think there may be a more elegant way to handle this page than pills, but one baseline suggestion would be that since the pills handle all the relevant actions/links, and the inciting incident is broadly covered by the section highlighted in red (the "title"?), removing the non-pill section between those would reduce duplicative info. Thoughts on that approach, at least as a first step?

I could then also hidden comments being displayed in between those in non-pill form (still expandable, but with a "go to comment" link), but that might be personal preference. Mostly thinking that would feel closer to the inbox experience, which I think would be positive for UX consistency

You've laid down a great foundation here, so I do think it's just deciding where it can be streamlined :) great job so far!

hjiangsu:
Thanks for all the feedback! The original video is basically the same, except without the button/pills that you mentioned (community, user, comment, post)

> I think you've captured a lot of the functionality perfectly, but the button/pill setup feels like it could be done in a way to reduce duplication of information

I agree with you here! Now that I look back at it, it does feel like there's a lot of information that is being duplicated.

I can maybe mess around and see what other layouts I can come up with, but I'm not too good at UI/UX so I appreciate any feedback 😅 If you do have any thoughts on alternatives, by all means let me know!

One thing I will note is that there is often a "reason" attached to each entry which is displayed in the body, so I'd have to figure out some place to put that if I were to get rid of the body section.

> still expandable, but with a "go to comment" link

Ahh, the reason I did this instead of having a link to the comment is that often times, the comment content is removed. Although, maybe having the link is good so that it can provide some more context for the comment.

Thanks again for the feedback! I'll take those things into consideration 😄

hjiangsu:
So I played around with a few more layouts, and this is what I have so far! This is only for posts/comments. I haven't figured out how to make the layout yet for users (e.g., banned users). Let me know what you think of the changes!

Posts:

https://github.com/thunder-app/thunder/assets/30667958/7fe75877-4a22-4692-91d1-74333ad8a01f

Comments:

https://github.com/thunder-app/thunder/assets/30667958/abf1c0f7-3e7b-4e82-b947-96ce109e328a

hjiangsu:
Another update, I've added a different layout for users/communities:

Users:

https://github.com/thunder-app/thunder/assets/30667958/de896e70-9acc-4217-84f4-58e654790208

Communities:

https://github.com/thunder-app/thunder/assets/30667958/a552de61-0c4d-42f8-a185-229daa8dfa19

machinaeZER0:
I think this looks really good! I wonder if it would be helpful to use either our profile or inbox pages as a jumping off point for making some sort of "landing page" from which a mod could browse these different categories (posts/comments/users/etc). The landing page could also have a section for admin stuff like setting/managing roles in a community, perhaps? 

I do like the idea of being able to take actions like blocking/unblocking stuff on a long press or something (or exposing other interactions) to keep these sections compact, though if a lot of actions need completing then hiding stuff in a sub menu will add a lot of steps for mods... Something to think about. Maybe a (collapsible?) icon bar under each item with intuitive icons would work for any "quick actions."

It will definitely be a lot of Stuff, so it's quite an undertaking to make sure everything is laid out in an intuitive way! Do we know of any examples in other apps that could be used as reference?

hjiangsu:
> I wonder if it would be helpful to use either our profile or inbox pages as a jumping off point for making some sort of "landing page" from which a mod could browse these different categories (posts/comments/users/etc).

Hmm, this might be possible, but it might be slightly harder to filter out events based on categories. From a technological standpoint, the API accepts different filters, but those filters are based on actions rather than categories if that makes sense (e.g., the API allows us to filter events for featuring/unfeaturing posts, removing/restoring posts, etc. but it doesn't provide a filter for general "post" related actions).

> I do like the idea of being able to take actions like blocking/unblocking stuff on a long press or something

I like the ideas that you mentioned - in particular, being able to perform actions on this page (given that you have the correct role). For example, an instance admin could navigate to the instance modlog and perform related actions there (undo/redo a specific event). Similarly, a community moderator could navigate to the community modlog.

I think if the general concept and design looks good, we can use this as the starting point for modlogs? We can then further iterate and add more features into here (like the quick actions that you mentions) in separate PRs!

machinaeZER0:
That makes sense to me!

hjiangsu:
This should be ready to go now. I'll go ahead and merge it in first.

@micahmo Feel free to review if you'd like, but no need to since I know there's quite a bit of changes in this one

micahmo (approved):
I didn't look too closely but it all LGTM. 😊

micahmo:
@hjiangsu How would you feel about putting the new modlog button under a popup menu on the main feed like it is on community feeds?

Pros:
* We will probably have more actions there at some point, so it will have to be overflowed eventually, and then users might be upset that it moved.
* The icon is a bit heavy compared to the rest of the UI in that area.

Cons:
* It might hide the discoverability of a cool new feature.

Just a point for discussion. 😊

hjiangsu:
Definitely! I'm all for placing the modlog into a popup menu

Issue #341, opened 2023-07-12
Switch accounts at the post level


hiramfromthechi:
**Is your feature request related to a problem? Please describe.**
When you have multiple accounts (whether for personal or business), it's difficult managing all of them.

Oftentimes, I'll come cross posts (under my personal account) that I wanna upvote, comment, or save—but to a different account.

Here's an example:
I'm on my personal account, hiramfromthechi. I come across a basketball post that I'd like to interact with, but I'd like to interact with that post with my basketball page, InThePaintCrew.

Currently, I need to switch accounts (which might refresh the feed, or least lose the current position of the feed), and then find the post I wanna interact with so that I can do it with my InThePaintCrew account.

**Describe the solution you'd like**
At the post level, it'd be great to choose the account you wanna interact with first, before submitting the interaction. Especially when it comes to commenting and saving posts.

**Describe alternatives you've considered**
N/A

**Additional context**
Both Boost for Reddit and Slide for Reddit had this. Would love to see this functionality make it to Thunder for Lemmy.

hjiangsu:
This is an interesting idea! Do you happen to have any screenshots or videos on how this might potentially work (from Boost/Slide) This would help with visualizing how we go about implementing this in the future

I would have to see how this works when you're interacting with posts on a different instance. I believe it should be possible, but I'm not sure if there are any side effects that have to be considered here.

jordan1776:
Liftoff has this feature somewhat, where you can change what instance you are viewing a post from.
Putting this into the reply would be a nice feature.
![Screenshot_20230712-195949~2](https://github.com/thunder-app/thunder/assets/6100101/66d7107c-6935-48c1-91da-c5b8f4d7d9e3)

At the very least the comment drawer should say what user your using to reply with.

Coehill:
RedReader also has this functionality for both posting and commenting.  Personally, I find being able to switch accounts when commenting the most useful since it happens with more frequency than posting.  However, both would be convenient.

Glazed:
This feature worked great in reddit apps, but doesn't necessarily work on Lemmy. A least not without caveats. It'll be fine if both of your accounts are on the same instance, but won't necessarily work if they aren't.

You can never use an account from a different instance to comment on a post. You must make your comment on *your instance's copy* of a post and then it will get federated back to the original post. (Or sometimes not... when federation fails, and only people on your instance will ever see your comment.)

[...]

PR #1159, opened 2024-03-01
Add support for posting as a different user

micahmo:
## Pull Request Description

This PR adds the ability to create posts under a different account than is currently selected. It does so by using a new technique whereby we can change the current account without reloading the app. When the user navigates back from the create post page, the previous account is restored and their feed is left untouched. (One drawback is that you won't see the "navigate to post" snackbar since the post might not be federated to the original account's instance.) We can hopefully use this same technique for other "temporary account" operations.

To perform the account switching, I reused the profile modal switcher, and disabled the following things.
* Ability to add new accounts/instances
* Ability to delete accounts
* Ability to see or select anonymous instances

There is also handling for if the selected community is not available on the instance of the desired account.

Future tasks:
* Add support for commenting using a different account (should be pretty similar).
* Add support for performing other actions (upvote/save/etc) using a different account (might be a bit different).

## Issue Being Fixed

Issue Number: Partly related to #341

### Demo 1: Show posting as a different user, show that the original user is restored and feed is untouched, see that we can view the new post

https://github.com/thunder-app/thunder/assets/7417301/7bf13d2b-3ace-4ac0-ba57-66bd0d0d0e12

### Demo 2: Show that we handle when the selected community is not yet federated to the new user's instance

https://github.com/thunder-app/thunder/assets/7417301/c81a7448-c9b0-4b70-bd26-024fd1e84922

## Checklist


lib/account/widgets/profile_modal_body.dart
```diff
@@ -18,8 +18,18 @@ import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 class ProfileModalBody extends StatefulWidget {
   final bool anonymous;
   final bool showLogoutDialog;
```
ggichure, line 20: some comments on how this constructors are used will go a long way.
`quickSelectMode`, `reloadOnSwitch` etc.

ggichure (approved):

hjiangsu (approved):
I would just like to mention again that is super cool - thanks for working on this! LGTM other than the one comment that @ggichure already mentioned.

As a side note: this prompted me to work on the comment page refactor so that we can hopefully apply this in the future as you mentioned

lib/account/widgets/profile_modal_body.dart
micahmo: All set! This should be good to go now!

Issue #1012, opened 2024-01-04
Comment language selection


Tealk:
just like the posts, the comments should also have a language selection see issue #157

**Describe the solution you'd like**

Add a language selection dropdown

Issue #738, opened 2023-09-14
Error "Could not determine post to comment to" when editing comment from own profile


everdred:
**Description**  
When editing a comment using the pencil icon next to a comment appearing on my profile, I always get the error "Could not determine post to comment to." 

**How to Reproduce**  
1. View my own profile.
2. Tap the pencil icon next to a comment.
3. Wait for my username and instance to load, just in case.
4. Modify comment, tap Post button, receive above error.
5. Go back, tap the comment to view it in context on the post.
6. Edit comment successfully using the pencil icon appearing there.

**Device & App Version:**  
 - OS: Android 13
 - Version: 0.2.4-2

micahmo:
@everdred Thanks for the report!

For the devs, there's some background in #700. Essentially the problem is that we have a `UserBloc`, not a `PostBloc`, so the comment editor can't find the parent post. I'm not sure if we can use `PostBloc` too, or make the comment editor handle either bloc type, or something else.

hjiangsu:
So I thought about this, and I think we need to do a refactor on how blocs work in the context of `CreateCommentPage`. 

This is what I'm imagining: We should create a new bloc for this screen for the sole reason that it'll be easier to manage data flow and reduce complexity. Since this is now it's own separate page, rather than a bottom sheet, it makes sense to do the following:

1. Create a new bloc to handle the state of the `CreateCommentPage`
2. Add in the relevant events that should happen within the context of `CreateCommentPage`. A few events that I can think of are creating comments, and editing comments. Things like delete and reporting comments should go in their respective blocs (post bloc)
3. For this new bloc, we should be able to pass in the required information for the given context. Some examples:
  a. The `postView` should be passed in if we're replying directly to a post
  b. The `commentView` should be passed in if we're replying directly to another comment, or if we are editing our own comment
  c. Any other context that we need in order to fill out the information on the create comment page
4. Refactor some of the current logic in the `CreateCommentPage` to work properly with this change. Since we don't have to propagate the error messages up anymore (this was a requirement before since we were using a bottom sheet), it'll hopefully simplify a lot of the logic that we have in place

This is just an initial writeup of my thoughts and can change! @micahmo I can go ahead and work on this to see how feasible it is first, and then when I have a working refactor, you can take a look at it!

micahmo:
Sounds like a great plan! 🎉

PR #1165, opened 2024-03-04
Create comment page refactor

hjiangsu:
## Pull Request Description

This PR refactors the existing create comment page and attempts to solve the following issues:
- It updates the create comment page to match logic similar to the create post page. This means that draft logic is all contained within the create comment page. It also means that all logic related to comment creation is contained within the page itself. This allows us to perform comment actions from anywhere without relying on passing in the correct blocs.
  - When a comment is created or edited successfully, it calls a callback function `onCommentSuccess`. The parent widget can use this callback function to update the state of the parent (e.g., add comment to list, update comment content for edits)
- Adds the ability to select a language for a comment. This brings us up to parity with the create post page.
- Updated the UI/UX to better match the overall UI of Thunder.
  - When replying to a post, the post information will be displayed in a similar fashion to the post page. It uses the pre-existing `PostSubview` widget, so any changes to the `PostSubview` will reflect on this page!
  - When replying to a comment, the comment information will be displayed in a similar fashion to the comment cards. It uses the pre-existing `CommentContent` to display the comment, so any changes there will also reflect on this page!

Most of the logic is now in place, and should be more or less functional! Some standing issues to be fixed:
- Need to add logic to update comment from the account page

Additional notes:
- I've disabled the comment action on the search page, and on the user feed page. My main reasoning here is that I'd like for users to visit the comment within it's full context in order to perform actions on it (vote, reply, etc). This also simplifies some of the overall logic and reduces edge cases! Do let me know your thoughts on this change though.

## Issue Being Fixed

Issue Number: #1012, #738

## Screenshots / Recordings

| Before | After |
|--------|--------|
| simulator_screenshot_11035348-06C9-4B12-BE0B-BEEFBE4E4E40 | simulator_screenshot_44F7F5FD-9615-4620-9B43-32BCE9C7407B |
| simulator_screenshot_04CEDB6F-B77B-47A1-9AF5-457395972889 | simulator_screenshot_71311E3F-11F5-447E-94F5-F703E6837840 |
| simulator_screenshot_82411B07-723F-4D80-8BC1-618C4382C51E | simulator_screenshot_A097A0DD-4BA7-4068-AE4D-9B1E1E455953 | 
| simulator_screenshot_B5B4150B-7D1D-460C-BF91-E161712B17AA | simulator_screenshot_7112944A-DF17-48AF-94B4-29F382A1F2FD | 

## Checklist


micahmo:
Just an FYI, I want to work on a "comment as other user" feature as some point. Let me know if you think I should wait for this PR or go ahead now.

hjiangsu:
I think it might be good to hold off on that until this is complete! I'll probably get back on this soon

hjiangsu:
@micahmo this should be more or less ready now (with the exception of the standing issues). I've updated the description to provide some more information/context. This is a relatively large change, so up to you if you want to do a code review on this!

@machinaeZER0 since I've updated the overall UI/UX, please let me know your thoughts on this!

machinaeZER0:
Hello! This is looking really nice :) Just a few thoughts/clarifications from me:

- I definitely see the usefulness of a language setting given the issues its absence is causing - I do wonder if there's a classier way to integrate it, though. are we able to default the language in a given community/instance to something acceptable, or no? And what are your thoughts on making the language selector a dropdown icon next to your personal info, rather than a text dropdown between your info and the composition section? (mostly unsure of what the icon would be - maybe just a three dot guy?)
image

- I think this is naturally improved having switched from an outline to a filled box style, but I think the quoted text isn't vertically centered. If that's true (might be my eyes playing tricks on me), is it possible to add a few pixels of padding to the bottom? 
image

- I'm sure this is outside the scope of this PR, but I still wish the send/post button was closer to the bottom, lolz. I think I'm in the minority on that one, but I like where it lives in most messaging apps from a reachability standpoint, rather than at the top of the screen. Maybe something to ponder for the future, but happy to keep things as they are on that front for the foreseeable :)

Let me know if I missed anything else you specifically wanted reviewed, but hope this was helpful!

micahmo:
Hey all, I'm still reviewing this PR, but I wanted to respond to @machinaeZER0's comments real quick.

> * are we able to default the language in a given community/instance to something acceptable

Given that the Lemmy UI does not, we probably wouldn't either.

| image |
| - |

> * what are your thoughts on making the language selector a dropdown icon next to your personal info

As long as it's obviously a language selector, I'm totally up for UI suggestions. Just keep in mind that this currently matches the Create Post page, so we'd have to change both. Also keep in mind that this page will also eventually have the "comment as other user..." feature, so there will be a chevron to the right of the username (where you circled) indicating you can pick a different user.

> * I still wish the send/post button was closer to the bottom

I also agree with this! A while back I did back I tried a mockup where the send button integrated with the preview button. I'll have to dig that up. Again, that would be a common change between this and the Create Post page, so probably a different PR as you mentioned.

machinaeZER0:
That all makes sense to me - if we're matching the "normal" Lemmy experience, thoughts on matching the language and styling of the dropdown here? A single dropdown (maybe pilled) that says "Select Language," rather than the current "Language: Select Language," I mean
image

I think it being shorter and contained within a pill would be positive changes, but also wouldn't be opposed to just doing the latter and keeping the longer copy of the former :)

machinaeZER0:
Made issue #1309 as a place to discuss the send/post button idea, also :)

lib/comment/view/create_comment_page.dart
```diff
@@ -0,0 +1,561 @@
...
+                                  MarkdownType.bold,
+                                  MarkdownType.italic,
+                                  MarkdownType.blockquote,
+                                  MarkdownType.strikethrough,
+                                  MarkdownType.title,
+                                  MarkdownType.list,
+                                  MarkdownType.separator,
+                                  MarkdownType.code,
+                                  MarkdownType.username,
+                                  MarkdownType.community,
```
micahmo, line 397: Did we lose spoiler here?

lib/comment/view/create_comment_page.dart
```diff
@@ -0,0 +1,561 @@
...
+      }
+    }
+  }
+}
+
+/// Creates a widget which displays a preview of a pre-selected language, with the ability to change the selected language
+///
+/// Passing in [languageId] will set the initial state of the widget to display that given language.
+/// A callback function [onLanguageSelected] will be triggered whenever a new language is selected from the dropdown.
+class LanguageSelector extends StatefulWidget {
```
micahmo, line 480: Doesn't have to be now, but is there any way to share this with the similar create post widget?

lib/shared/snackbar.dart
```diff
@@ -28,30 +28,33 @@ void showSnackbar(
...
-                          ? () {
-                              OverlaySupportEntry.of(context)?.dismiss();
-                              trailingAction();
-                            }
-                          : null,
-                      icon: Icon(
-                        trailingIcon,
-                        color: trailingIconColor ?? Theme.of(context).colorScheme.inversePrimary,
+            content: GestureDetector(
+              onTap: () => OverlaySupportEntry.of(context)?.dismiss(),
```
micahmo, line 32: Does this dismiss snackbars on tap? I wonder if we should have an X button for that.

unnamed

https://m3.material.io/components/snackbar/specs

micahmo (approved):
Awesome work, I know this was a huge effort! I did review the code but haven't tested yet (I'm ok with getting this merged in to test).

> I've disabled the comment action on the search page, and on the user feed page.

Does this also need to be done for inbox replies and mentions (the latter of which uses a totally different UI, not swipe gestures)?

hjiangsu:
Thanks for the feedback @machinaeZER0!

> are we able to default the language in a given community/instance to something acceptable, or no?

As @micahmo mentioned, we're mostly following what Lemmy UI does. I do believe we can make slight improvements to the language selector in future PRs (such as limiting the language options based on what is enabled on the instance/community).

Selecting a default language would be difficult because instances/communities can have more than one language. However, if the instance/community only has one language enabled, I can see making that the default!

> And what are your thoughts on making the language selector a dropdown icon next to your personal info, rather than a text dropdown between your info and the composition section?

As @micahmo mentioned,  the only issue here is that we'll be implementing ways to choose a different user to post the comment as (in the future). This would likely follow similar UI patterns to how the create post page looks like (with the chevron icon)

> is it possible to add a few pixels of padding to the bottom?

I can try to adjust that a bit more - it could be because of the spacing of the title text which causes an optical illusion

hjiangsu:
> Does this also need to be done for inbox replies and mentions

It doesn't apply to the inbox/mentions. The reasoning here is that if you get a reply or a mention, you might already have the general context of what the original discussion was about, unlike the user feed page or the search page. Additionally, sometimes you might want to just quickly reply to a reply/mention to acknowledge that you've seen it!

I'm not sure that I've tested all the edge cases everywhere, but I'm hoping that most of it gets caught during testing/nightlies.

lib/shared/snackbar.dart
hjiangsu: Good call!

I'll switch the approach here - the original intention was to have a way to dismiss the snackbar without swiping (while having an action present). This was because the "restored comment from draft" snackbar was annoying, and blocked the markdown toolbar entirely.

lib/comment/view/create_comment_page.dart
hjiangsu: I think we'll make this a separate PR, so that we can apply the suggestions that @machinaeZER0 mentioned with regards to the select language input!

lib/shared/snackbar.dart
hjiangsu: Done in https://github.com/thunder-app/thunder/pull/1165/commits/c4fb8c4d07363d5a32b336a72481c429ab749e5e!

lib/comment/utils/navigate_comment.dart
```diff
@@ -1,43 +1,95 @@
...
+        return MultiBlocProvider(
+          providers: [
+            BlocProvider<ThunderBloc>.value(value: thunderBloc),
+            BlocProvider<AccountBloc>.value(value: accountBloc),
+          ],
+          child: CreateCommentPage(
+            postViewMedia: postViewMedia,
+            commentView: commentView,
+            parentCommentView: parentCommentView,
+            onCommentSuccess: (CommentView commentView) {
```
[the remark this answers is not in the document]
hjiangsu: Done!

lib/comment/view/create_comment_page.dart
hjiangsu: Done!

hjiangsu:
| Before | After|
|-|-|
|image | image|

@machinaeZER0 does this seem to be a bit better?

machinaeZER0:
> Before 	After
> image 	image
> 
> @machinaeZER0 does this seem to be a bit better?

If it's possible to split the difference I might do that, but if that's literally one pixel's worth of difference then the original spacing is probably fine to keep as-is - probably better to not overly pad, anyway :) thank you for mocking it up, either way!

PR #1172, opened 2024-03-06
Fix CommunitySelector format

micahmo:
## Pull Request Description

This PR fixes an issue where the `CommunitySelector` would not respect the selected community full name syntax. It was using `getLemmyCommunity` which is more for resolving and verifying the identity of a community to a common syntax with which we can navigate. Instead it should have used `generateCommunityFullName` which is purely a textual transformation. This change should also slightly reduce the overhead of calling `getLemmyCommunity` which was `async`.

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu (approved):
LGTM!

Issue #1114, opened 2024-02-05
Feature Request: Show the precise date instead of a relative date for posts and comments 


dtantono:
**Is your feature request related to a problem? Please describe.**
I think it will be great if we can keep tracking our read post accurately on home screen. 

**Describe the solution you'd like**
Is it possible to add a new toggle settings to Show the precise date (i.e: Sun Nov 16 20:14:56 2014 UTC) instead of a relative date (i.e: 7 days ago), for posts and comments?

**Describe alternatives you've considered**
Old Reddit RES

**Additional context**

hjiangsu:
Thanks for the issue! Do you think having an option to long-press on the date to show the precise date works in this scenario? Or would you prefer to show the full time at all times?

It would also be useful to share any screenshots of what you mean to help coordinate the UI (if UI changes are needed)!

dtantono:
a toggle to show fulltime at all times will be nice

hjiangsu:
Update: there is now a PR which adds this setting. In terms of the date format, I've added a way to preselect from a given list of formats.

Attached to the PR #1174 is also a video demo of the feature. Let me know if this is what you were expecting!

PR #1174, opened 2024-03-07
Add option to show full date and select date format

hjiangsu:
## Pull Request Description

This PR adds the ability to show full dates on posts, and select a date format from a pre-selected list. In addition, I added a way to add subtitles to a given PickerItem

## Issue Being Fixed

Issue Number: #1114

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/6633a69a-961a-463a-9fd5-ce19dc219ca9

## Checklist

### DIFF
--- a/lib/settings/pages/post_appearance_settings_page.dart
+++ b/lib/settings/pages/post_appearance_settings_page.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+import 'package:intl/intl.dart';
 import 'package:lemmy_api_client/v3.dart';
 import 'package:expandable/expandable.dart';
 import 'package:collection/collection.dart';
@@ -79,6 +80,22 @@ class _PostAppearanceSettingsPageState extends State<PostAppearanceSettingsPage>
   /// When enabled, posts that have been marked as read will be dimmed
   bool dimReadPosts = true;
 
+  /// When enabled, the full date will be shown
+  bool showFullPostDate = false;
+
+  /// The selected date format. This is only used when `showFullPostDate` is enabled
+  DateFormat? selectedDateFormat;
+
+  /// List of available date formats to select from
+  List<DateFormat> dateFormats = [
+    DateFormat('MMMM dd, yyyy HH:mm:ss'),
+    DateFormat('E, dd MMM yyyy HH:mm:ss Z'),
+    DateFormat('yyyy-MM-dd HH:mm:ss'),
+    DateFormat('dd/MM/yyyy HH:mm:ss'),
+    DateFormat('MM/dd/yyyy HH:mm:ss'),
+    DateFormat('yyyy-MM-ddTHH:mm:ss'),
+  ];
+
   /// When enabled, cross posts will be shown on the post page
   bool showCrossPosts = true;
 
@@ -110,6 +127,8 @@ class _PostAppearanceSettingsPageState extends State<PostAppearanceSettingsPage>
       showPostAuthor = prefs.getBool(LocalSettings.showPostAuthor.name) ?? false;
       useDisplayNames = prefs.getBool(LocalSettings.useDisplayNamesForUsers.name) ?? true;
       dimReadPosts = prefs.getBool(LocalSettings.dimReadPosts.name) ?? true;
+      showFullPostDate = prefs.getBool(LocalSettings.showFullPostDate.name) ?? false;
+      selectedDateFormat = prefs.getString(LocalSettings.dateFormat.name) != null ? DateFormat(prefs.getString(LocalSettings.dateFormat.name)) : DateFormat(DEFAULT_DATE_FORMAT);
 
       // Compact View Settings
       compactPostCardMetadataItems =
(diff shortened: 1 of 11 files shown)


K4LCIFER:
Looks awesome! I would maybe make the date more of a grey color so that it isn't as jarring. Also would it be possible for the user to specify their own date format? It seems unnecessary to try and specify every possible date combination that a user may want. I would use one as a sane default (probably something of the form "Mar 7, 2024 12:23") and then allow the user to specify alternatives if they wish.

lib/settings/pages/post_appearance_settings_page.dart
```diff
@@ -511,6 +540,35 @@ class _PostAppearanceSettingsPageState extends State<PostAppearanceSettingsPage>
...
+              iconDisabled: Icons.date_range_outlined,
+              onToggle: (bool value) => setPreferences(LocalSettings.showFullPostDate, value),
+              highlightKey: settingToHighlight == LocalSettings.showFullPostDate ? settingToHighlightKey : null,
+            ),
+          ),
+          SliverToBoxAdapter(
+            child: ListOption(
+              description: l10n.dateFormat,
+              disabled: !showFullPostDate,
+              value: ListPickerItem(
```
micahmo, line 558: For some reason the initially selected value is not shown as highlighted in the list the first time you open the modal. Once you pick a value, subsequently it is highlighted. So I think you'd have to clear your preferences to repro. But it's visible in your demo.

Maybe it's because the `DEFAULT_DATE_FORMAT` constant is not a reference match for any of the entries in the `dateFormats` list.

micahmo (approved):
This is super cool!

The only thing I could possibly suggest is to have a "system" setting where the date format follows the locale of the device. This is pretty common in date formatting, and it looks like it is possible in Dart by constructing a `DateFormat()` without any parameters.

**EDIT**: I take it back. Constructing a `DateFormat()` with no parameters always falls back to a hard-coded format, `yMMMMd` plus `jms`.

https://github.com/dart-lang/i18n/blob/c6c6c25030c88ca4db7cacc6ad312c8f3d1ba357/pkgs/intl/lib/src/intl/date_format.dart#L724-L727

I will see if there's another way to construct a local format.

**EDIT2:** Ok, sorry for all the edits! We can use the existing `yMMMMd` and `jms` formats while passing the current locale.

``` dart
  final DateTime now = DateTime.now();
  
  print(Intl.systemLocale); // Prints: en_US
  DateFormat dateFormat = DateFormat.yMMMMd(Intl.systemLocale);
  DateFormat timeFormat = DateFormat.jms(Intl.systemLocale);
  print('${dateFormat.format(now)} ${timeFormat.format(now)}'); // Prints: March 7, 2024 3:59:22 PM
```
---
To test that it really does respect the locale...
``` dart
final DateTime now = DateTime.now();
  
  initializeDateFormatting('en_GB');
  DateFormat dateFormat = DateFormat.yMMMMd('en_GB');
  DateFormat timeFormat = DateFormat.jms('en_GB');
  print('${dateFormat.format(now)} ${timeFormat.format(now)}'); // Prints: 7 March 2024 16:04:04
```

I completely agree with @K4LCIFER on having a "sane default", but unfortunately that means different things to different people around the world. 😊

hjiangsu:
> I would maybe make the date more of a grey color so that it isn't as jarring

I can maybe adjust the date colour if using custom dates, otherwise it'll default to what it is now!

> would it be possible for the user to specify their own date format?

This was my plan for a future enhancement! The reason why I opted to show a few pre-selected options for now is to reduce complexity of the feature. To allow for user-specific date formats, there would be some additional requirement to validate the user input, or implement it in a way that's intuitive to use (e.g., maybe something similar to the post card metadata customization with chips for different sections)

hjiangsu:
> The only thing I could possibly suggest is to have a "system" setting where the date format follows the locale of the device.

I like this suggestion! I'll change it so that it does that based on what you mentioned with addling the system locale!

micahmo:
> I'll change it so that it does that based on what you mentioned with addling the system locale!

Sorry for commenting so much on this one. One more thought to add is that, if the system format happens to be a duplicate one of the predefined formats, you could had that so it's only displayed once. You can see what the system format would be by checking `pattern`.

``` dart
'${dateFormat.pattern} ${timeFormat.pattern}'
```

hjiangsu:
Alright, I made some slight changes based on the feedback!
- When the full date is shown, it shows up as slightly dimmer on the feed page
- By default, the full date will use the system's locale to determine the format of the date
- Removed any duplicated formats if it exists (https://github.com/thunder-app/thunder/pull/1174/commits/53184a57e70836ad1f3b3ded8ee38a70fe59ef4d)
- I also removed the seconds from the date. My reason for this is that seconds don't contribute as much information as the rest of the timestamp, but takes up horizontal real-estate which makes it feel "clunky"

Here's an updated video with the changes. Let me know what you think!


https://github.com/thunder-app/thunder/assets/30667958/b4324418-91e3-4eab-bcc5-8ddba5242219

lib/settings/pages/post_appearance_settings_page.dart
```diff
@@ -79,6 +80,15 @@ class _PostAppearanceSettingsPageState extends State<PostAppearanceSettingsPage>
...
   bool dimReadPosts = true;
 
+  /// When enabled, the full date will be shown
+  bool showFullPostDate = false;
+
+  /// The selected date format. This is only used when `showFullPostDate` is enabled
+  DateFormat? selectedDateFormat;
+
+  /// List of available date formats to select from
+  List<DateFormat> dateFormats = [DateFormat.yMMMMd(Intl.systemLocale).add_jm()];
```
micahmo, line 90: Is there any concern that `add_jm()` is not taking a locale here? You can call `DateFormat.jm(<locale>)`, but I couldn't find a great way to combine to formats that both take a locale.

micahmo (approved):
Awesome feature and nice improvement from feedback!

lib/settings/pages/post_appearance_settings_page.dart
hjiangsu: I _believe_ it takes the locale from the date format? `DateFormat.yMMMMd(<locale>)` and `DateFormat.jm(<locale>)` are just constructors which create a DateFormat object with the applied locale.

`add_jm()` is then just a member function which changes the internal DateFormat pattern. When we finally parse the date format, I think that's when the locale is applied. Also, it seems like by default, it uses `Intl.systemLocale` for the locale if none is passed in.

https://api.flutter.dev/flutter/intl/DateFormat/DateFormat.d.html

micahmo: Sweet, thanks for confirming that!

PR #1178, opened 2024-03-08
Improve sort submenu transition

micahmo:
## Pull Request Description

While testing #1156, I discovered that I really liked the submenu transition for the modlog filter better than the existing transition for the sort submenu, so this PR copies that technique.


hjiangsu (approved):
LGTM!

PR #1184, opened 2024-03-11
Add reader mode for in-app browser

micahmo:
## Pull Request Description

This PR adds a reader mode for Android (and iOS) using the in-app browser. Where previously the reader mode option would only be shown on iOS, it is now shown on Android too *if* the in-app browser is selected. It can also be toggled on/off dynamically in the in-app browser, regardless of the default. Note that history isn't shared between the modes.

One potential downside is that the package I'm using for this hasn't been updated for a while, but after a considerable amount of searching (which took a while, since it isn't even on pub.dev), it was not only exactly what I was looking for but seems to work very well. (I would've been happy with a simple web scraper that returns the body contents for us to render, but this has a whole controller architecture and everything.)

Aside from that I think it's pretty easy to sell the usefulness of a cross-platform reader mode. It makes page loads and navigation much faster and scrolling is more responsive.

Note: I also introduced some abstraction to handle the different types of controllers so there aren't a lot of conditionals everywhere.

> Review without whitespace.

## Issue Being Fixed

Issue Number: #760

## Screenshots / Recordings

> Video was too big for GitHub.

https://files.catbox.moe/aetnb4.mp4

## Checklist


hjiangsu:
Hey! Finally getting around to this one - my only worry with this is that the package will be come outdated as time goes on. We can create a fork of this on the thunder repo, but it most likely means we'll have to maintain it. Do you have any thoughts on this?

micahmo:
It's totally your call if you don't mind taking on another package to maintain. Of course I'm willing to help as needed! But I do think this is a pretty neat feature to have to flesh out in the in-app browser, and this package fits in perfectly with our existing browser controller architecture.

micahmo:
@hjiangsu I'm trying to go through my old PRs one-by-one to see what needs to be done to either complete or close them.

For this one, I did just resolve the merge conflicts, so it's working with the latest. However, I know you expressed some concern about introducing a dependency on a package that doesn't seem to have been updated in a while, and that also goes against your general philosophy of trying to slim down the app outlined [here](https://github.com/thunder-app/thunder/pull/1578#issue-2597996589).

So my question is: how do feel about introducing this dependency? If you think it's worth it, I'd say let's merge this now so it doesn't generate any more conflicts, and then we can fork the package later. If you don't think it's worth, I'm totally ok closing this. Just let me know!

hjiangsu (approved):
I'm okay with bringing this in, since the PR was created before #1578 (and it also seems like a fairly useful feature to have)!

> I know you expressed some concern about introducing a dependency on a package that doesn't seem to have been updated in a while

I did a quick skim of the package, and it seems to be fairly straight forward implementation so my initial concerns are lessened (since it'll be easy for us to fork it and fix it up if needed).

I'll go ahead and merge this in!

PR #1187, opened 2024-03-11
More link handling improvements

micahmo:
## Pull Request Description

This PR is similar to #1014 but for users rather than communities (now that they are sharing the same feed).

I discovered a link that Thunder was accidentally identifying as a user, when it was actually a community, which led me to discover that the user error handling is a bit clunky. First, I fixed that so it works a little better. Then I fixed the link parsing to see that it's actually a community link.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Original behavior

https://github.com/thunder-app/thunder/assets/7417301/6da2999f-3516-46e0-b330-f3746d54ba9e

### After user error handling improvements

https://github.com/thunder-app/thunder/assets/7417301/dac014bb-7cc9-4cdb-9a3b-c41e5162f2af

### After link parsing improvements

https://github.com/thunder-app/thunder/assets/7417301/91a725f9-940e-4643-90d0-7ea8659794a0

## Checklist

### DIFF
--- a/lib/feed/bloc/feed_bloc.dart
+++ b/lib/feed/bloc/feed_bloc.dart
@@ -124,7 +124,7 @@ class FeedBloc extends Bloc<FeedEvent, FeedState> {
 
   /// Handles clearing any messages from the state
   Future<void> _onFeedClearMessage(FeedClearMessageEvent event, Emitter<FeedState> emit) async {
-    emit(state.copyWith(status: state.status == FeedStatus.failureLoadingCommunity ? state.status : FeedStatus.success, message: null));
+    emit(state.copyWith(status: state.status == FeedStatus.failureLoadingCommunity || state.status == FeedStatus.failureLoadingUser ? state.status : FeedStatus.success, message: null));
   }
 
   /// Handles post related actions on a given item within the feed
--- a/lib/feed/view/feed_page.dart
+++ b/lib/feed/view/feed_page.dart
@@ -466,16 +466,17 @@ class _FeedViewState extends State<FeedView> {
                             ],
                           ),
                           // Widget representing the bottom of the feed (reached end or loading more posts indicators)
-                          SliverToBoxAdapter(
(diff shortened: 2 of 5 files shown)


hjiangsu (approved):
LGTM! Nice catch!

hjiangsu:
I found one link issue that might be related to all of this: https://lemmy.world/post/13037417 - it tries to find a community but its a link to a mastodon post I believe.

micahmo:
Thanks, I'll take a look! I guess the ideal solution would be to fall back to a web browser if we can't navigate natively. This could be fixed by trying to retrieve the object before navigating, but that would add a delay. The current behavior is nice in that the navigation happens immediately an then we try to load. So we might need to have some way of propagating the error back to the previous page so we can close the community/user feed and then open the browser.

hjiangsu:
Yeah, I agree - I do like that the navigation happens immediately (rather than having a delay) for UX purposes. Would it be possible to do something like this:
- If we detect a user/community link, we can look through our instance list first to see if the instance is available there. If the instance is there, then we know its a proper Lemmy instance and do what we do now (with immediate navigation). If we want to, we can decrease the sensitivity for the instance ci workflow (from `active_users_monthly > 50` to `active_users_monthly > 25`)
- If the instance is not in our list, then we'll try to retrieve the object, but add a snackbar to show that we're doing something so that the user doesn't try to tap on the link again

micahmo:
That seems like a pretty good compromise! Just keep in mind that users/communities from non-Lemmy software will cause a delay, even if the navigation ultimately works.

hjiangsu:
> Just keep in mind that users/communities from non-Lemmy software will cause a delay, even if the navigation ultimately works.

Yeah. Another possible solution is to keep what we have currently, but show a button on the empty navigated page to open the link in the browser if we fail to resolve the user/community! I'll leave it up to you on which method you prefer (I'm good with either or, or an alternative if you have one)

PR #1188, opened 2024-03-11
Fix deep link handling

micahmo:
## Pull Request Description

This PR fixes an issue where the deep link handling would not work more than once if the app is open. (The demo shows one trivial example, but this is a general problem. Because the state was not reset, we could never handle more than one link.)

CC: @ggichure

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/cf5bf941-3f2e-4d86-b846-f41c9b49049e

### After

https://github.com/thunder-app/thunder/assets/7417301/4e3ccad3-38cc-4634-ae2c-8d0456e7a91b

## Checklist


hjiangsu (approved):
LGTM!

PR #1193, opened 2024-03-11
Fix instance explorer

### LINKED ISSUE #1189: Community list looping in 'explore instance'
### Bug Description

When you go and scroll through communities tab in 'explore instance', the community list keeps looping and there is no end to it.

### Expected Behaviour

There is no loop in community list and there is an end to it.

### Steps to Reproduce

1. Go to 'explore instance'
2. Go to community tab
3. Scroll through the list (for easy sake, explore a small instance with a few community)

### Additional Context

_No response_

### App Version

v0.3.0-1

### Device

Samsung Galaxy J6

### OS

Android 10

micahmo:
## Pull Request Description

This PR attempts to fix #1189. I believe the problem is less that the results are cycling endlessly, and more that we are retrieving the same page more than once, causing duplicate results, making it look like a cycle. This seems to be caused by the fact that the scroll trigger can happen multiple times in quick succession, causing the community loading to be re-entrant when we want it to complete (and emit the new page number) before loading more.

## Issue Being Fixed

Issue Number: #1189

### DIFF
--- a/lib/instance/pages/instance_page.dart
+++ b/lib/instance/pages/instance_page.dart
@@ -44,6 +44,7 @@ class InstancePage extends StatefulWidget {
 
 class _InstancePageState extends State<InstancePage> {
   final ScrollController _scrollController = ScrollController(initialScrollOffset: 0);
+  bool _isLoading = false;
 
   bool? isBlocked;
   bool currentlyTogglingBlock = false;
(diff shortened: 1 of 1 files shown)


hjiangsu:
Ahh I see what's happening here. Could you do something similar to how the FeedPage does it? It checks the bloc's state to see if its currently fetching, and only calls the `onScroll` logic when its not in the fetching state:

This is the corresponding logic from the FeedPage:
```dart
if (
  _scrollController.position.pixels > _scrollController.position.maxScrollExtent * 0.7 &&
  context.read<FeedBloc>().state.status != FeedStatus.fetching
) {
  context.read<FeedBloc>().add(FeedFetchedEvent(
    feedTypeSubview: selectedUserOption[0] ? FeedTypeSubview.post : FeedTypeSubview.comment
  ));
}
```

hjiangsu:
Hmm, I see. I guess we can stick to the solution you have for the time being!

hjiangsu (approved):

Issue #1127, opened 2024-02-15
Copy text from post, and select text to copy


Fmstrat:
### Is Your Feature Request Related to a Problem?

Can't copy text in certain areas of the app, which also leads to translation issues (subset of this report at https://github.com/thunder-app/thunder/issues/187). 

The current implementation of copy on comments follows the highly annoying "copy it all" methodology that messaging apps have started using. This is not a good use case for community content, as many content blocks can be large, and a common use case would be to highlight a word or phrase to use in a search later.

There have been other closed issues related to post copying (also a subset), but I believe they were closed accidentally when comment copy was added, see (https://github.com/thunder-app/thunder/issues/162).

### Feature Description

- Allow long-press on posts to then highlight and copy specific text from the post using standard OS copy.
- Add "Select text to copy" to the long-press menu on comments, which would bring the cursor into the comment to use standard OS copy.

### Alternative Solutions

See above.

### Additional Context

N/a

everdred:
I just came to open an issue just like this one. I found a word in a post I wanted to define, and I was going to long-press to select the word and "share" it to my dictionary application.

Issue #1047, opened 2024-01-14
Option to share from local instance


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
Currently, the share-link for a post points at that post's originating instance only.

**Describe the solution you'd like**
It would be helpful to also have an option to share a link that points to the post on the local instance.

**Describe alternatives you've considered**
None.

**Additional context**
None.

micahmo:
+1! The same problem is true of sharing comments.

PR #1199, opened 2024-03-12
Improve post/comment bottom sheets

micahmo:
## Pull Request Description

This PR aims to improve the ever-expanding list of actions for posts and comments by grouping some actions.
1. User-, community-, and instance-related actions are now grouped in respective sub-menus.
2. The options now directly reference the entities by name.
3. The bottom sheet now supports the Android back button to navigate back from sub-menus.
4. For own posts, delete is now moved up to the top row with the other post-related actions.

These changes required a bit of refactoring so that the architecture matches that of other menus like the modlog one. So this is one of those cases where testing may be more fruitful than reviewing. 😊 This refactor should help with future maintainability and usability.

Here are some things that I identified that didn't quite work right before (so this PR doesn't break them), but they should be fixed in the future (and some of them will now be easier!):
 - Remove ability to block self
 - Add ability to unblock users/communities (i.e., check current block state)
 - Immediately update subscribe state rather than having to refresh feed
 - Move comment "Copy Text" to a sub-menu and also provide the option to make the text selectable and show raw ([#1127](https://github.com/thunder-app/thunder/issues/1127))
 - Improve the share sub-menu for both posts and comments to provide options for the current instance and the original instance ([#1047](https://github.com/thunder-app/thunder/issues/1047))

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Posts

https://github.com/thunder-app/thunder/assets/7417301/ac7c82b3-52b2-472d-aea7-dfdad86a6cc0

### Comments

https://github.com/thunder-app/thunder/assets/7417301/58fc0395-92a0-4aae-8799-05b91aee2aa3

## Checklist

### DIFF
--- a/lib/community/utils/post_card_action_helpers.dart
+++ b/lib/community/utils/post_card_action_helpers.dart
@@ -1,4 +1,6 @@
+import 'dart:async';
 import 'dart:io';
+import 'package:back_button_interceptor/back_button_interceptor.dart';
 import 'package:flutter/material.dart';
 
 import 'package:share_plus/share_plus.dart';
@@ -9,6 +11,7 @@ import 'package:thunder/account/bloc/account_bloc.dart';
 
 import 'package:thunder/community/bloc/community_bloc.dart';
 import 'package:thunder/community/enums/community_action.dart';
+import 'package:thunder/core/enums/full_name_separator.dart';
 import 'package:thunder/core/enums/media_type.dart';
 import 'package:thunder/core/models/post_view_media.dart';
 import 'package:thunder/core/singletons/lemmy_client.dart';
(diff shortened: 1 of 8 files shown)


hjiangsu:
The demo videos look great!

I do have one suggestion for the user/community/instance (the main bottom sheet) - do you think its possible to move the entity names to a subheading within the options?
- This allows more horizontal space for the entity name and would look cleaner in my opinion
- `PickerItem` has a `subtitle` parameter that we could use for this (I added it in for the full date format)!

> So this is one of those cases where testing may be more fruitful than reviewing

I agree! I just took a very brief look at the code changes.

hjiangsu (approved):

micahmo:
> do you think its possible to move the entity names to a subheading within the options

Great idea! I just pushed this, how does it look?

| Header | Header |
|--------|--------|
| image | image |

PR #1200, opened 2024-03-13
Even more link handling!

micahmo:
## Pull Request Description

This PR fixes an issue where non-Lemmy links may incorrectly be identified as such. The fix is, after parsing the link, check whether the instance is a known Lemmy instance. If so, navigate immediately. If not (e.g., if it is federated from another platform), retrieve the object before navigating to ensure that it is a valid object. If we cannot retrieve the object, we will fallback to the web browser. Note that this only done in cases when the "link" is a valid URL. If not (i.e., if it's just "Lemmy syntax"), then it would be pointless to fall back to the browser.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/pull/1187#issuecomment-1992241511

### Demo 1: Shows that the original issue is fixed

https://github.com/thunder-app/thunder/assets/7417301/87c3d57c-54ca-4390-8372-ae9fdea69f27

### Demo 2: Shows that valid Lemmy links from non-Lemmy software still work (with a delay)

https://github.com/thunder-app/thunder/assets/7417301/b50ee8a7-fb76-4d4d-a202-a10aab8b99c2

### Demo 3: Shows that valid links from Lemmy software still work (no delay)

https://github.com/thunder-app/thunder/assets/7417301/dedb353d-eaad-49b4-8b68-d6581b287eec

## Checklist

### DIFF
--- a/lib/feed/utils/utils.dart
+++ b/lib/feed/utils/utils.dart
@@ -9,6 +9,7 @@ import 'package:thunder/community/bloc/community_bloc.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
 import 'package:thunder/feed/feed.dart';
 import 'package:thunder/instance/bloc/instance_bloc.dart';
+import 'package:thunder/shared/pages/loading_page.dart';
 import 'package:thunder/shared/sort_picker.dart';
 import 'package:thunder/community/widgets/community_drawer.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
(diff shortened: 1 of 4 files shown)


hjiangsu:
I think that looks good! Do you think we can add a snackbar, or some UX feedback (spinning indicator) when we're attempting to fetch the object?

micahmo:
> Do you think we can add a snackbar, or some UX feedback (spinning indicator) when we're attempting to fetch the object?

Hmm good idea. I tried a snackbar first, but it actually took a little while to appear, so it still feels like there's nothing happening for a bit. I also couldn't find a great way to hide it when the page does eventually load (I think you had the same problem).

My next idea was to show a loading page when we have to retrieve the object so it _looks_ like we've navigated immediately. The trick to making this seamless is to hide the page and show the "real" page without any additional animations. For that I used `pushReplacement` with no `transitionDuration` and it seems to work well! It's not completely seamless when going to the browser (when not using the in-app one), but I think it's good enough!

Here's a before and after so you can see the time differences.

### Before

https://github.com/thunder-app/thunder/assets/7417301/8799cd57-a5eb-4535-b95a-47dd2815811a

### After

https://github.com/thunder-app/thunder/assets/7417301/b0a2f210-aed6-4112-89a7-f181d390fd1b

Here's also what it looks like in the other two browser modes.

### In-app

https://github.com/thunder-app/thunder/assets/7417301/942b9ec0-9a9f-44d8-8cc4-1a30585f9b7e

### External

https://github.com/thunder-app/thunder/assets/7417301/3a450ea6-4647-4dfd-9a2b-f4908a6c0203

I pushed these changes, so this should be good to go if you like it.

hjiangsu:
I think that looks good! Just one more question - does the loading happen for all external URLs (e.g., if we press a link on a post, will the brief loading page show up?)

micahmo:
> does the loading happen for all external URLs (e.g., if we press a link on a post, will the brief loading page show up?)

No, it should only happen when we have to manually load an object, which is only for community/user links when we don't recognize the instance (like the first two examples in the demo).

https://github.com/micahmo/thunder/blob/555d24b8b75c933b19bd6a2d0ca9d9254604a0d4/lib/utils/links.dart#L233

https://github.com/micahmo/thunder/blob/555d24b8b75c933b19bd6a2d0ca9d9254604a0d4/lib/utils/links.dart#L262

hjiangsu:
Ahh okay! In that case, I think this works 😄

hjiangsu (approved):

Issue #1198, opened 2024-03-12
Show user instance name on posts


glyph-se:
### Is Your Feature Request Related to a Problem?

The same username can exists on many instances, therefore it should be possible to always show the instance name together with the username.

### Feature Description

Currently there is a setting in _Appearance -> Comments -> Show User instance_ . I found no way to show the user instance in neither the post list nor in the post+comment page after opening it. There are however two related options, _Show post author_ and _Show User Display Name_.

### Alternative Solutions

Instead of adding a new _Show User instance_, an alternative would be to extend the _General -> Advanced -> User Format_ with new options that are just _username_ and _display name_ . Then that setting could be applied to more than just the comments.

### Additional Context

I understand that this will take up extra space in the post list, but I believe it would be nice to at least have an option for this. Possible the user can be rendered on its line if that would make it look better, but I don't think that is necessary in all cases, the regular wrapping can probably be used.

hjiangsu:
@micahmo I believe #1227 fixes this issue? I'll mark this with the proper tags - let me know if you think otherwise!

micahmo:
Agreed!

PR #1092, opened 2024-01-28
Colorized usernames

ggichure:
## Pull Request Description

This PR adds support for colorized usernames. Using  `colorScheme.secondary` else  `textTheme.bodyLarge?.color`. 

## Issue Being Fixed

Issue Number: #240 

## Screenshots / Recordings

w2

## Checklist

- [x] Did you update CHANGELOG.md?
- [x] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1201, opened 2024-03-13
Add support for user/community styles

micahmo:
## Pull Request Description

This PR adds more configuration options for user and community full name styles.

### Background
In #973 we introduced the ability to choose the style of user and community full names (either `a@b` or `a · b`, and more recently `!` and `@` were added as prefixes). Then in #944, @ggichure added a cool new style for comment usernames where the username is more weighted than the instance name. However, that new format is only available in comments and is not configurable. Then in #1092, @ggichure added colorized usernames, but the same restrictions applied.

This PR aims to make the "style" of the full names configurable universally in the same way that the "format" is currently configurable.

#### Notes
* The `commentUseColorizedUsername` setting has been eliminated in favor of the new username style settings. However, there is a link from the Appearance > Comments page to the new setting (see demo), and there is a migration.
* To ease into this change, and since the special formatting and colorized usernames previously only applied to comments, I am only applying these new changes there as well. However (as should be obvious from the introduction), the goal is to replace all usages of user/community names with these new formatting options, including adding full names where they are not already available (e.g., [#1198](https://github.com/thunder-app/thunder/issues/1198)).

CC: @ggichure

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/276ade74-572f-4fcd-a0e0-5a2d059da251

## Checklist


ggichure (approved):

hjiangsu (approved):
Sorry it took so long for me to get to this PR - I think its really cool and it looks great!

One slight modification I would make (but I'll leave the final decision to you) is to rename "Weight" to "Bold" (e.g., "Bold user name", Bold instance name"). I think "bold" is easier to understand than "weight", but that's just my personal opinion 😅

Let me know what you think! Otherwise, LGTM

micahmo:
> rename "Weight" to "Bold"

I agree, that's a more user-friendly term. I think I'll continue to use `weight` in code, since that corresponds to the text styling.

Should be good now!

| image |
| - |

PR #1192, opened 2024-03-11
Distinguish direct video URLs

hjiangsu:
## Pull Request Description

This PR changes the following:

- Refactored `MediaView` to reduce code complexity.
- Refactored `parsePostView` to reduce code complexity, and added a check for direct video URL extensions
- Added a new `MediaTypeBadgeItem` for video types. It uses a purple background as the base colour.
- Added method to detect video URLs (based off #1081)

Additionally, I moved `lib/utils/image.dart` -> `lib/utils/media/image.dart`. This helps better categorize media util functions (image/video)

## Issue Being Fixed

Issue Number: Partially solves #1177 by showing a badge whenever a video URL is found.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/d5479a24-14b0-4d1c-9eaa-5dcf4039946a

## Checklist

- [ ] Did you update CHANGELOG.md?
- [x] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

Issue #948, opened 2023-12-04
Image dismissing smoother animation 


fedikhatib:
When dismissing an image the action happen very fast that it's like flashing.

https://github.com/thunder-app/thunder/assets/34138112/27876ad9-646e-4a1b-bbf8-80b9af015fab

**Describe the solution you'd like**

Maybe a smoother animation.

**Describe alternatives you've considered**

Here's the sync Lemmy client behavior also the background color is the dominant image color

https://github.com/thunder-app/thunder/assets/34138112/3a31fa0f-6f11-4aaf-845f-12f8fbb4b904

micahmo:
Hey @fedikhatib, if you still use Thunder, can you let us know if the animation looks better to you in the latest version (0.3.0)? Thanks!

PR #1207, opened 2024-03-15
Fix image close transition

micahmo:
## Pull Request Description

This is a small PR which fixes an issue that was "broken" in #1192. I'm not sure if it was intentional, but the `reverseTransitionDuration` parameter was removed from the image viewer page after a refactor. As a result, closing images feels a bit sluggish now. I do agree that the original time (`50`) was probably too low, and was the cause of #948, but the new default of `300` still feels slow and is slower than the opening duration of `100`, so I propose to make both `100`.


hjiangsu:
Ahh I thought removing the `reverseTransitionDuration` would cause it to use the same value as `transitionDuration`! I wanted to make it `100` but might've inadvertently set it to the default value.

hjiangsu (approved):

PR #1208, opened 2024-03-15
Add in-app changelog

micahmo:
## Pull Request Description

This PR adds support for an (optional) in-app message and changelog for when the app is updated.
* This is optional, and while it's on by default, there's a single button which users can press to disable (without having to go into settings).
* No notification is shown the first time the app launches, only when the version is detected to change.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

> [!NOTE]
> In this demo, I have the dialog appearing for every restart of the app so I can demonstrate the settings. In reality, it should only appear when the version has changed.

https://github.com/thunder-app/thunder/assets/7417301/aec3ced7-f4cb-4383-98aa-910226183492

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -93,6 +93,8 @@ enum LocalSettings {
   markPostAsReadOnScroll(name: 'setting_general_mark_post_read_on_scroll', key: 'markPostAsReadOnScroll', category: LocalSettingsCategories.general, subCategory: LocalSettingsSubCategories.feed),
   showInAppUpdateNotification(
       name: 'setting_notifications_show_inapp_update', key: 'showInAppUpdateNotifications', category: LocalSettingsCategories.general, subCategory: LocalSettingsSubCategories.notifications),
+  showChangelogsAfterUpdate(
+      name: 'setting_show_changelogs_after_update', key: 'showChangelogsAfterUpdate', category: LocalSettingsCategories.general, subCategory: LocalSettingsSubCategories.notifications),
   scoreCounters(name: 'setting_score_counters', key: "showScoreCounters", category: LocalSettingsCategories.general, subCategory: LocalSettingsSubCategories.feed),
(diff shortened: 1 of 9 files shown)


hjiangsu (approved):
I think it looks good! Just a couple of minor things:
- For the settings label, we can reword it to "Show Update Changelogs" (the subtitle can be left as is)
- For the bottom sheet title, we can rename it to display "{version} Release Notes" (e.g., v0.3.0-1 Release Notes). I think that's a bit more concise than what its set to right now!
- I'm not sure how I feel about having 3 buttons at the bottom. Ideally, I'd like to reduce it down to two ("Close" and "Disable Release Notes"). I did try to experiment with the changes to see if I could use something like a [DraggableScrollableSheet](https://api.flutter.dev/flutter/widgets/DraggableScrollableSheet-class.html) so that we don't need to have the "Expand" option, but I wasn't able to get it to work just right. I think we can leave it as is for now (but maybe subject to change in the future)

micahmo:
> For the settings label, we can reword it to "Show Update Changelogs" (the subtitle can be left as is)

Done! I went ahead and refactored the whole setting to use this new name.

> For the bottom sheet title, we can rename it to display "{version} Release Notes" (e.g., v0.3.0-1 Release Notes). I think that's a bit more concise than what its set to right now!

Hmm, we can certainly tweak the wording. I did write it intentionally so that it comes across as an announcement. I think it's pretty common for apps to tell you that they updated, and it's a nice reminder for those who use automatic updates (and maybe don't care as much about the changelog).

If you want to save some room, how about something like this?

| image |
| - |

Otherwise, if you feel strongly about your wording, of course I'll be happy to defer.

> I'm not sure how I feel about having 3 buttons at the bottom. Ideally, I'd like to reduce it down to two ("Close" and "Disable Release Notes").

Fair enough! I removed expand/collapse. I did expand the default height a bit as a compromise. Hope that's ok!

| image |
| - |

hjiangsu:
Thanks for the changes! I think all of the changes were reasonable!

micahmo:
Oops, I didn't push the title change. Let me make a new small PR for that.

Issue #961, opened 2023-12-06
Issues with taglines


micahmo:
**Description**  

I discovered a couple of new issues with taglines. I'm not sure if these go back to the feed refactor or something else.

1. Opening the FAB causes the tagline to refresh many times during the animation.
  - I remember having a similar issue when first adding taglines, and the solution was to fetch the random tagline once when loading the feed, rather than during the build. If it's done during the build, then any time the feed rebuilds, the tagline can change. I see now it's fetched directly in the `TagLine` widget, so I'm not sure how best to fix it.
2. Taglines don't work for anonymous instances (note that it will work if you switch from an account to anonymous; you have to restart the app in anonymous mode).

**Screenshots**  

### Demo of first issue

https://github.com/thunder-app/thunder/assets/7417301/58fa66ce-b5a1-4cfe-9278-bc34b4f2d7d5

Niranjan-Dorage:
@micahmo is this issue still open ??
if yes i would like to **give it a try**

micahmo:
Yes it is, go for it!

Niranjan-Dorage:
@micahmo ok

CTalvio:
Can this be closed?

micahmo:
I believe taglines still don't work for anonymous instances.

CTalvio:
Ah, they don't. You could change the title to that specifically now that the other thing is fixed.

PR #1209, opened 2024-03-15
Fix tagline issue

micahmo:
## Pull Request Description

This PR fixes the first issue mentioned in #961 by extracting the tagline selection logic out of the `TagLine` widget so that a new one is not chosen every time the widget builds. The cache has also been removed, so that a tagline is only chosen once initially, and then again when a refresh occurs.

## Issue Being Fixed

Issue Number: #961

## Screenshots / Recordings

See #961 for the **before** video.

https://github.com/thunder-app/thunder/assets/7417301/5fa7f70e-1142-4cd7-992f-3d7cd1099774

## Checklist

### DIFF
--- a/lib/feed/view/feed_page.dart
+++ b/lib/feed/view/feed_page.dart
@@ -185,6 +185,8 @@ class _FeedViewState extends State<FeedView> {
   /// List of post ids to queue for removal. The ids in this list allow us to remove posts in a staggered method
   List<int> queuedForRemoval = [];
 
+  String? tagline;
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM! I believe there was another issue where non-logged in accounts were not able to see Taglines? I think that is a different issue than this but wanted to bring it back up.

I think it's a result of us using  `context.read<AuthBloc>().state.getSiteResponse` which is only filled in when logged in, but I could be wrong about that.

micahmo:
> I believe there was another issue where non-logged in accounts were not able to see Taglines? I think that is a different issue than this but wanted to bring it back up.
> 
> I think it's a result of us using `context.read<AuthBloc>().state.getSiteResponse` which is only filled in when logged in, but I could be wrong about that.

That would make sense! It would explain why they keep working when switching to an anonymous instance (because `getSiteResponse` might not get cleared out), but they stop working when you restart.

I can try to look at that when I get a chance.

PR #1211, opened 2024-03-20
Fix instance snackbar

micahmo:
## Pull Request Description

This PR fixes a strange issue where the snackbar indicating an error navigating to an instance would not appear unless the user interacted with the UI in some other way. I believe this is related to the fact that we're using `addPostFrameCallback`, but there's no other building operation happening. (Most if not all other snackbars are usually related to something that changes/builds at the same time). There may be a better way to fix this, and I'm open to ideas, but for now I just added a condition for whether to use the post-frame or not.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/421fbb8f-8fc2-418f-9b6c-c142a49934b6

### After

https://github.com/thunder-app/thunder/assets/7417301/52fc879e-721c-45a0-8d61-d1dbdd408e85

## Checklist


micahmo:
Note that this might not be an issue now with #1218, but it might not hurt to have this option in the snackbar anyway.

hjiangsu:
I think we can leave this out if its no longer needed! If we end up needing it in the future, we'll merge it in - just helps with keeping the codebase a bit cleaner 😅

micahmo:
Sounds good, I'll close this PR but keep my branch just in case.

PR #1212, opened 2024-03-20
Fix taglines

micahmo:
## Pull Request Description

I believe I broke Thunder for instances without a tagline in #1209, so this fixes that.


hjiangsu (approved):
Ahh yeah I encountered this as well when I was working on some stuff - I'll merge this in and revert my local fix!

PR #1214, opened 2024-03-20
Add development flavor

hjiangsu:
## Pull Request Description

This PR adds flavors to Thunder (mostly dev-facing feature for now). When we run a debug build of Thunder on a physical device, the debug build stops working when we disconnect the debugger. Therefore, to use Thunder outside of development, I have to run a profile/release build after I finish all dev-related tasks. This was a bit of a hassle so I decided to implement flavors.

For now, this PR adds two flavors: production and development.
- The `production` flavor is the general release/nightly builds of Thunder (e.g., profile/release builds)
- The `development` flavor is for any in-development feature builds of Thunder (e.g., debug builds). The only difference is the color of the icon to help differentiate it a bit. Future enhancements to this would be to add a separate splash screen to indicate the flavor.
- We can further extend this in the future to provide a `nightly` flavor for nightly builds.

The benefit of this is that we can now compare between two different versions of Thunder by running different versions on different flavors.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

image

simulator_screenshot_96FA35CF-02F7-4EC4-B6FE-8064E74790ED

## Checklist


micahmo (approved):
Ooh, this looks like a great idea!

By any chance is there a way to show the flavor name in the home screen?

hjiangsu:
> By any chance is there a way to show the flavor name in the home screen?

I was trying this out, but it looks weird on iOS because theres a 12 character limit before iOS removes spaces. So `Thunder Development` would end up being `ThunderDevelopment`. I'm not too sure if Android has the same restrictions too. I was thinking a better way to distinguish them would be to change up the icon a bit more (maybe add a badge or something similar) to visually separate the flavors.

micahmo:
> thinking a better way to distinguish them would be to change up the icon a bit more (maybe add a badge or something similar)

I like that idea!

hjiangsu:
This is the best I was able to come up with - having text with the logo is a bit tough since the area is so small. Hopefully, this is large enough to distinguish the letters

image

image

simulator_screenshot_4E7CD5E7-E8F1-4716-B941-AA0B3C85AF0A

micahmo:
Oh I think that looks awesome!! It's not too small at all.

hjiangsu:
I'll go ahead and merge this in, and we can add more tweaks to the design later if we want to!

micahmo:
The build command I usually use for my phone is

``` powershell
flutter build apk --profile --no-tree-shake-icons
```

It seems to still work, but now spits out

```
Gradle build failed to produce an .apk file. It's likely that this file was generated under .\build, but the tool couldn't find it.
```

Is this because the naming of the .apk changed? Should I change my command?

hjiangsu:
Ahh you have to add `--flavor development` or `--flavor production` to generate the build. It produces `app-development-release.apk` or `app-production-release.apk` because of the flavor.

hjiangsu:
I've also adjusted the VSCode `launch.json` to provide support for launching the different flavors. You might be able to use that as well!

image

micahmo:
> you have to add `--flavor development` or `--flavor production` to generate the build

Thanks, that worked!

PR #1206, opened 2024-03-15
Fix advanced share sheet

micahmo:
## Pull Request Description

This PR fixes an issue where the advanced share sheet would present an option to include an "external link" when there wasn't one (i.e., text posts).

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|--------|--------|
| image | image | 

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1216, opened 2024-03-20
Fix text post issues

micahmo:
## Pull Request Description

This PR fixes a couple of issues related to text previews.
1. Check that the body is not empty (rather than not null) when determining whether to expand/collapse the body.
2. Fix not showing the text preview in collapsed body mode. This seems to be broken because all `PostViewMedia`s have at least one `media`, all of whose properties are empty, but it's still there, meaning that `postViewMedia.media.isNotEmpty` checks don't really work. This was the same issue that caused [#1206](https://github.com/thunder-app/thunder/pull/1206), so I'm not sure what changed, but something did, and I'm probably not addressing the root cause.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/models/media.dart
+++ b/lib/core/models/media.dart
@@ -7,7 +7,7 @@ class Media {
     this.originalUrl,
     this.width,
     this.height,
-    this.mediaType,
+    required this.mediaType,
   });
 
   /// The original URL of the media - this applies if the original URL of the media originates from a external link
@@ -23,7 +23,7 @@ class Media {
   double? height;
 
   /// Indicates the type of media it holds
-  MediaType? mediaType;
+  MediaType mediaType;
 
   @override
   String toString() {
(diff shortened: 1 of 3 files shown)


hjiangsu:
> This seems to be broken because all PostViewMedias have at least one media, all of whose properties are empty, but it's still there, meaning that postViewMedia.media.isNotEmpty checks don't really work. 

Ahh, this was a change related to me adding in the video MediaType. I've made it so that all PostViewMedias have a least a single `media`. The difference here now is that I've also added MediaType.text for posts that are just text.

What we can do is instead of `postViewMedia.media.isNotEmpty`, we can use `postViewMedia.media.first.mediaType != MediaType.text` to exclude text-only posts.

I think eventually, I'd like to have it so that PostViewMedia looks like this:

```dart
import 'package:lemmy_api_client/v3.dart';
import 'package:thunder/core/models/media.dart';

class PostViewMedia {
  PostView postView;
  MediaType mediaType;
  List<Media> media;

  PostViewMedia({
    required this.postView,
    required this.mediaType,
    this.media = const [],
  });
}
```

There's 4 different types I can think of at the moment for the media type: `text`, `link`, `image`, `video`, `gallery` (for posts that contain more than one image)

micahmo:
Ahh ok, sounds good, I will make a change to look for that and probably go back and fix #1206 the same way too. Thanks!

micahmo:
This should be good to go now!

PR #1223, opened 2024-03-21
Add modlog link handling

micahmo:
## Pull Request Description

I've been seeing a lot of direct links to modlogs, so I thought it would be cool if we could parse those as deep links to Thunder! This PR adds support for handling modlog links, both within the app and externally.

It should support all of the potential parameters, except page, since I wasn't sure how to make that work with how we currently support paging in the modlog.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

> Video was too big for GitHub.

https://files.catbox.moe/a5wphw.mp4

## Checklist


hjiangsu (approved):
LGTM!

> It should support all of the potential parameters, except page, since I wasn't sure how to make that work with how we currently support paging in the modlog.

Yeah, this would take some work to support so I think its okay for the time being!

PR #1227, opened 2024-03-22
Use new user/community styles

micahmo:
## Pull Request Description

This PR is the next step of #1201. It adds the following:
* Option to show user instance in post listings
* Option to show user instance in post body
* Option to show community instance in post body
	* Please let me know if you can think of any places I've missed the ability to see the full name of an entity.
* Usage of the new common user/community styling in most of the places where we show the full name, including the following:
	* Create post page
	* Community page
	* Search results
	* Post listing
	* Inbox items
	* Modlog page
	* Post body
	* Community sidebar
	* User sidebar
	* User settings
		* There were a few places that I didn't apply the styling because that area of the UI imposed some other styling which I thought should take precedence.
		* I also couldn't update usages within `TextScroll` and `AutoSizeText` since they just take a `String`. Need to re-evaluate those.

Finally, to avoid inconsistencies and help new users out of the box, I disabled all styling by default. I had previously only enabled bold usernames to preserve the current look of the comments, but the user night not want this throughout the app, and it can looks strange when compared to communities that are unstyled, so I figured it should be an intentional choice.

Next up: I think a next good task would be to extract the "display name" setting from posts and make it part of the styling for both user and community names.

## Issue Being Fixed

Issue Number: #1198

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/9817f81a-cb1c-4a1f-ab76-b530fc63efc2

## Checklist


hjiangsu (commented):
Hmm, now that I'm looking at this, do you think it might be better to convert `generateCommunityFullNameWidget` and `generateUserFullNameWidget` into StatelessWidgets/StatefulWidgets rather than functions that return a widget?

This would allow us to keep Flutter's conventions of using StatelessWidget/StatefulWidget classes, and this might just be more efficient in the long run since Flutter can perform optimizations on them.

Thoughts?

micahmo:
Good call! Let me know what you think of it now.

hjiangsu (approved):
This looks better! Thanks for changing it up 😄

PR #1228, opened 2024-03-23
Load small instances

micahmo:
## Pull Request Description

This PR fixes an issue where the instance viewer would spin indefinitely when it should have "finished" loading.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/69dfba9e-6d41-4106-bf3d-e547f5a63654

### After

https://github.com/thunder-app/thunder/assets/7417301/d3fd9b6b-cd60-43aa-9c23-8e6457f9e29b

## Checklist

### DIFF
--- a/lib/instance/cubit/instance_page_cubit.dart
+++ b/lib/instance/cubit/instance_page_cubit.dart
@@ -11,6 +11,8 @@ import 'package:thunder/utils/error_messages.dart';
 part 'instance_page_state.dart';
 
 class InstancePageCubit extends Cubit<InstancePageState> {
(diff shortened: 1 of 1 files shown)


hjiangsu (commented):
Just one suggestion - could we extract the limit into a constant and use that instead? It'll make it easier to adjust the limit parameter in the future if we need to!

micahmo:
Done!

hjiangsu (approved):
LGTM! Thanks

PR #1232, opened 2024-03-24
Fix text wrapping

micahmo:
## Pull Request Description

This PR fixes an issue where all titles of bottom picker items were limited to one line, but that breaks cases where we want it to wrap. Since that was done to accommodate user/community names in #1199, and since we're now putting those names in the subtitle, I've just reverted that change.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|--------|--------|
| image | image | 

## Checklist


hjiangsu (approved):
LGTM!

PR #1236, opened 2024-03-25
Chore: upgrade gradle version, migrate to flutter_sharing_intent

hjiangsu:
## Pull Request Description

This PR upgrades the Android `gradle` and kotlin language versions. This was brought up when I was attempting to set up `unifiedpush` with Thunder (but ran into some build issues). As a result of this upgrade, I had to change a few things.

Here's an overall summary of the changes (review without whitespaces):

Toolchain
- `kotlin_version`: `1.7.10` -> `1.9.22`
- `android_gradle_plugin`: `7.3.1` -> `8.3.1`
- `gradle`: `7.5` -> `8.7`. This is based on https://docs.gradle.org/current/userguide/compatibility.html#kotlin
- The Java target is now `JavaVersion.VERSION_17` instead of `JavaVersion.VERSION_1_8`.

Packages
- Migrated from `receive_sharing_intent` to `flutter_sharing_intent`. There was an incompatibility with building after upgrading to the latest gradle version which prompted this change.

Because of the migration from `receive_sharing_intent` to `flutter_sharing_intent`, a bunch of refactoring was made. I tested it on an emulator and it seems to be working, but let me know otherwise.

@micahmo if you could try this and let me know if it works properly, that would be great! I would try the following:
- Sharing image/link/text to see if it fills out the appropriate information
- Opening a link in Thunder
- Attempt to build the app on a physical device (both debug and profile) to see if it succeeds on your end

## Issue Being Fixed

Issue Number: N/A, but allows us to set up `unifiedpush` package in the future.

## Checklist


hjiangsu:
Thanks for doing it so quick! I'll go ahead and merge this first since I want to fast-track this a bit (otherwise I might have to make changes to my local dev setup 😅)

Do let me know if you run into any issues!

PR #1242, opened 2024-03-28
Fix missing semantic label on report filter

micahmo:
## Pull Request Description

Fix missing semantic label on report filter.


hjiangsu (approved):
LGTM!

Issue #211, opened 2023-07-04
Comments with markdown footnotes don't render correctly


imbrucy:
**Description**  
Any comments with markdown footers render as large grey blobs.

**How to Reproduce**  
1. Open this comment in Thunder: https://lemmy.one/comment/627186

**Expected Behavior**  
Should render similar to in the web version of Lemmy.

**Screenshots**  
https://pixelfed.social/i/web/post/581637123813379215

**Device & App Version:**  
 - Device: Android: Samsung Galaxy S21
 - thunder v0.2.1+10

hjiangsu:
Oh hmm, I didn't even know there was a feature for markdown footers. I'll have to dig a bit deeper into this and see if this is even supported at the moment, but thanks for bringing it up!

hjiangsu:
Found the issue, and it should be fixed in the upcoming release!

PR #1243, opened 2024-03-28
Chore: Update dependencies

hjiangsu:
## Pull Request Description

This PR upgrades some of our dependencies to their latest versions. I've also removed the dependency override on the markdown package since the package now supports footnotes. For reference, the original issue with markdown is here: https://github.com/thunder-app/thunder/issues/211

@micahmo I looked through the changes for most of the packages, and it seems like a big portion had to do with increasing the `compileSdkVersion` and `gradle` versions for Android. Because of this, do let me know if you encounter any issues running on Android!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - background_fetch (1.2.4):
+  - background_fetch (1.3.2):
     - Flutter
   - device_info_plus (0.0.1):
     - Flutter
@@ -29,6 +29,8 @@ PODS:
     - Flutter
     - FlutterMacOS
   - permission_handler_apple (9.3.0):
+    - Flutter
+  - pointer_interceptor_ios (0.0.1):
     - Flutter
   - share_plus (0.0.1):
     - Flutter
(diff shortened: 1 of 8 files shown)


micahmo (approved):
LGTM! I will keep an eye out for any issues.

micahmo:
@hjiangsu I've noticed a couple things broken so far.

1. It looks like autocomplete suggestions for input dialogs (community/user/etc.) don't work any more. I did notice some code changes in that area.
2. It looks like the Android back button no longer works to navigate back to the feed page from another page, nor does it show the snackbar on the feed page. It now instantly closes the app in both of those scenarios.

micahmo:
@hjiangsu Quick update. I started looking into the back button issue. It looks like it's related to the upgrade of `GoRouter`, since downgrading it back to `^12.1.1` fixes the issue.

I also know `WillPopScope` is being deprecated; however the alternative, `PopScope`, doesn't seem to be playing nicely with `GoRouter` either (see [here](https://github.com/flutter/flutter/issues/138737)). `canPop: false` is ignored and `onPopInvoked` is never called. 😢 (I think it works in other parts of the app because we're using different navigation.)

I'll see if I can find something else that works with the latest version of `GoRouter`!

hjiangsu:
Ahh I was looking at this yesterday! I replaced `WillPopScope` with the `BackButtonInterceptor` that we use on other parts of the app. The changes are in this branch: https://github.com/thunder-app/thunder/tree/fix/android-back-button

I haven't created a PR yet because I was still trying to do some tests, but you can check it out and let me know if this works for you!

micahmo:
Oh yeah, that works perfectly, thanks! 😄

hjiangsu:
Awesome, in that case, I'll make a PR for it

PR #1248, opened 2024-03-29
Fix sidebar overflow

micahmo:
## Pull Request Description

This is a small PR which addresses some sidebar overflow I noticed.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|--------|--------|
| image | image | 

## Checklist

### DIFF
--- a/lib/community/widgets/community_sidebar.dart
+++ b/lib/community/widgets/community_sidebar.dart
@@ -231,13 +231,16 @@ class CommunityModeratorList extends StatelessWidget {
                       mainAxisAlignment: MainAxisAlignment.start,
                       crossAxisAlignment: CrossAxisAlignment.start,
                       children: [
-                        Text(
-                          mods.moderator.displayName ?? mods.moderator.name,
-                          overflow: TextOverflow.ellipsis,
-                          maxLines: 1,
-                          style: const TextStyle(
-                            fontWeight: FontWeight.bold,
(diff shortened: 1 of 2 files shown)


lib/community/widgets/community_sidebar.dart
```diff
@@ -231,13 +231,16 @@ class CommunityModeratorList extends StatelessWidget {
...
                       children: [
-                        Text(
-                          mods.moderator.displayName ?? mods.moderator.name,
-                          overflow: TextOverflow.ellipsis,
-                          maxLines: 1,
-                          style: const TextStyle(
-                            fontWeight: FontWeight.bold,
-                            fontSize: 16,
+                        SizedBox(
+                          width: MediaQuery.of(context).size.width * 0.55,
```
hjiangsu, line 235: Just curious, but where does the 0.55 come from?

hjiangsu (approved):
LGTM! Just one small question.

lib/community/widgets/community_sidebar.dart
micahmo: I just tweaked the percentage until it looked good. I figured since it was a percentage rather than a fixed value it should scale to other devices. What do you think?

I tried `FractionallySizedBox` first, since that's what the whole sidebar uses, but it didn't work.

PR #1249, opened 2024-03-29
Handle search result comments on deleted posts

micahmo:
## Pull Request Description

I noticed that the Lemmy API returns comments from deleted posts. I debated whether we should exclude such comments from search results entirely, but I figured a big change like that should come from the Lemmy side of things. Instead, I made a change in the UI to indicate that the post is deleted and to prevent navigation.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/27ace74c-de4d-4449-8222-9877b61f7a81

### After

https://github.com/thunder-app/thunder/assets/7417301/23de2634-71a7-4385-90ff-6b70c7464df4

## Checklist

### DIFF
--- a/lib/shared/comment_reference.dart
+++ b/lib/shared/comment_reference.dart
@@ -94,7 +94,7 @@ class _CommentReferenceState extends State<CommentReference> {
           ${formatTimeToString(dateTime: (widget.comment.comment.updated ?? widget.comment.comment.published).toIso8601String())}\n
           ${widget.comment.comment.content}""",
       child: InkWell(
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM!

micahmo:
Just as a followup, this has been fixed in the backend in https://github.com/LemmyNet/lemmy/pull/4634. I don't think it hurts to keep our fix though!

hjiangsu:
Yeah, it doesn't hurt to keep this in (especially for instances on older versions that don't have the fix yet)

PR #1250, opened 2024-03-29
Fix jank when using full height images, and other misc media preview imrpovements

hjiangsu:
## Pull Request Description

This PR fixes an issue with layout jank that occurs when using full height images and a few other minor issues.
- Fix layout jank due to `thumbnailUrl` not being present in some occasions
- Added height values to ViewMode enums to remove our scattered usage of constants
- Added a timeout when attempting to fetch image dimensions - this should help with some loading times if the response times of the image URL is slow. This does mean that some images may look squished or stretched if a timeout occurs but it might be worth it for faster loading. Ideally, the Lemmy API should provide the height/width dimensions so that we don't have to manually fetch these dimensions
- Added a background color to the image container when the image has not loaded yet

If any image dimensions look off after this change, let me know!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/enums/view_mode.dart
+++ b/lib/core/enums/view_mode.dart
@@ -1,1 +1,9 @@
-enum ViewMode { compact, comfortable }
+enum ViewMode {
+  compact(75.0),
+  comfortable(150.0);
+
+  /// The height of media previews for the given view mode
+  final double height;
+
+  const ViewMode(this.height);
+}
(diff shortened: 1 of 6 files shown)


micahmo (commented):
Hmm, I tried testing these changes and right off the bat the feed is filled with stretched images. Is it worth the tradeoff of faster loading?

| image |
| - |

hjiangsu:
Nice catch! I didn't test that out. I'll check it out and see - the stretching should only happen on card view

hjiangsu:
Hmm, it looks correct on my device. I'm not sure what's causing the stretching. I'm out at the moment so I can't check on the emulator!

image

hjiangsu:
Just tested it on Android as well and ti looks okay to me. Could you double check if you can reproduce the stretching?

image

micahmo:
Alright I'm not sure what was going on, but I can't repro any more either!

micahmo (approved):

Issue #678, opened 2023-08-23
autoscroll to comment


firefoxlover:
I try to explain it.

1. You are in your notifications view and you want to see what you actually wrote, that a person answered
2. You click on the message and get catapulted into the post comments

But it doesnt autoscroll / exclude other comments automatically for me.

My proposal would be to only show the relevant thread and scroll to the answer of your comment.

Using the latest version from FDroid currently.

Awesome app, the best for Lemmy!

micahmo:
Thanks for the proposal @firefoxlover!

> My proposal would be to only show the relevant thread

This is already done.

> scroll to the answer of your comment

Yes, this needs to be added. Thanks again!

K4LCIFER:
> You are in your notifications view

One thing to note is that this also should apply to comments from a users comment history, and not just for comments from notifications.

shortwavesurfer2009:
I suggest a slightly different tact. How about only loading the comment it is a direct reply of and the reply itself. Say you are replying to an L8 comment, instead of loading L1 through L7 as well only load L7 and L8. Your reply then becokes L9

K4LCIFER:
> I suggest a slightly different tact. How about only loading the comment it is a direct reply of and the reply itself. Say you are replying to an L8 comment, instead of loading L1 through L7 as well only load L7 and L8. Your reply then becokes L9

This is what the Lemmy UI does, so, at least for the sake of continuity, this could be a good option.

micahmo:
I just want to link this issue with PR #1363, which laid the groundwork for this, as well as issue #1413, which contains follow ups to that PR.

PR #1252, opened 2024-03-29
Feature/improve comment context

micahmo:
## Pull Request Description

This PR is a _small_ improvement to comment context. It's _not_ quite the full implementation that we need to add ([described here](https://github.com/thunder-app/thunder/issues/678#issuecomment-1928020130)) but it helps a little. Currently, the selected comment is highlighted. However, if there is a long post, you may have to scroll down quite a bit before you can see the comment. This PR adds the ability to scroll to the first comment. Now it still may not show the highlighted comment immediately (if it's deep down in the thread) but at least it will put you in the right area. I also fixed an issue where using the navigation buttons would lose the comment context.

> [!NOTE]
> Previously the "View all comments" button was built as part the `1`st indexed widget in the list. As a result, scrolling to index `1` would go to the top of the button instead of the top of the comment. I moved the button into the `0`th widget (i.e., the post body), which fixed the scroll issue, but it meant that the button couldn't participate in the `SlideTransition`, so I added a new animation for the button to disappear. It arguably looks a little better now. 😊

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/ad7eb3ed-a083-4efe-b9c9-ddcc8329c5d8

### After

https://github.com/thunder-app/thunder/assets/7417301/12f3f0f5-857a-4845-9441-7226f5e68bac

## Checklist

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -617,9 +617,21 @@ class PostBloc extends Bloc<PostEvent, PostState> {
 
   Future<void> _navigateCommentEvent(NavigateCommentEvent event, Emitter<PostState> emit) async {
     if (event.direction == NavigateCommentDirection.up) {
-      return emit(state.copyWith(status: PostStatus.success, navigateCommentIndex: max(0, event.targetIndex), navigateCommentId: state.navigateCommentId + 1));
+      return emit(state.copyWith(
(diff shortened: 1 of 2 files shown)


hjiangsu (approved):
LGTM!

PR #1261, opened 2024-04-01
Fix tagline overflow

micahmo:
## Pull Request Description

This PR fixes an issue with taglines. Currently there is some logic to condense the tagline if it's very long. However, this logic performs a very simple string truncation, which can break the tagline if it truncates in the middle of markdown.

I fixed this issue by using a similar technique to the post body expander, so that the it is done by height, rather than characters. I also introduced a measurement, so the truncation only happens if the tagline is taller than a specified height. We might consider introducing this mechanism back into the post body (because I've seen cases where the body is short and the expander does almost nothing).

> Review without whitespace.

### Before

https://github.com/thunder-app/thunder/assets/7417301/ec0a20e6-c62b-4dbb-9012-4c243afc18f6

### After

https://github.com/thunder-app/thunder/assets/7417301/6fda4168-7bb3-4cb2-8f73-f90cb5fa028e

### After -- short tagline still has no expander

| tagline_short |
| - |

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/feed/view/feed_page.dart
+++ b/lib/feed/view/feed_page.dart
@@ -3,6 +3,7 @@ import 'dart:math';
 
 import 'package:back_button_interceptor/back_button_interceptor.dart';
 import 'package:expandable/expandable.dart';
+import 'package:flex_color_scheme/flex_color_scheme.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


hjiangsu (approved):
LGTM!

PR #1263, opened 2024-04-01
Apply name styling to auto-scroll

micahmo:
## Pull Request Description

This PR is a followup to #1227 wherein it was mentioned that the new name styling could not be applied to `TextScroll` widgets. This PR fixes that issue that by swapping `TextScroll` for a more generic `Marquee`, which can take any widget (not just a `String`) -- in this case, the styled name widgets. I believe it could support a similar scrolling mechanism as the old widget (infinite scrolling from RTL with a pause), but I actually like its default behavior better (scrolling RTL, then back LTR, with no duplication), so I left it like that. It should also not scroll at all if the name is short enough.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/ba1dfdf5-25e6-42e8-9fd3-4e0a1c7e7169

### After

https://github.com/thunder-app/thunder/assets/7417301/4a1f030b-b502-4002-98c3-9748e39708e7

## Checklist


hjiangsu:
Nice! I do like the feeling of the new `Marquee` widget. Looking at the [pub](https://pub.dev/packages/marquee_widget) page, it seems like it's a fairly simple implementation. I'm wondering if we can create our own widget to use for the Marquee rather than add a package (which seems to no longer be maintained)

For reference, this is the source: https://github.com/yousifk/marquee_widget/blob/master/lib/marquee_widget.dart. The license is MIT which should allow us to use it!

micahmo:
Good point! I copied over the file (and applied code fixups as suggested by the linter). I also added a section to our LICENSE file to reference the MIT-licensed file and its original copyright. Let me know if that looks good to you.

hjiangsu:
Looks good! Thanks for doing this - also appreciate adding the addition to the LICENSE 😊

hjiangsu (approved):

PR #1265, opened 2024-04-02
Add ability to view raw markdown

micahmo:
## Pull Request Description

This PR adds the ability to view the raw markdown for posts and comments. For the latter, I created a new subsection of options in the comment long-press menu which now includes both "Copy Text" and the new option. I chose to use the term "source" because that's what Lemmy UI uses.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/576c06ce-5778-4ce3-aa35-873bfdd02f41

## Checklist

### DIFF
--- a/lib/shared/thunder_popup_menu_item.dart
+++ b/lib/shared/thunder_popup_menu_item.dart
@@ -4,14 +4,22 @@ import 'package:flutter/material.dart';
 class ThunderPopupMenuItem extends PopupMenuItem {
   final IconData icon;
   final String title;
+  final bool? checkboxValue;
+  final void Function()? onCheckboxValueToggled;
 
-  ThunderPopupMenuItem({super.key, required super.onTap, required this.icon, required this.title})
-      : super(
+  ThunderPopupMenuItem({
+    super.key,
+    required super.onTap,
+    required this.icon,
+    required this.title,
+    this.checkboxValue,
+    this.onCheckboxValueToggled,
+  }) : super(
           child: ListTile(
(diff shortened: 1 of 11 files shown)


lib/shared/thunder_popup_menu_item.dart
```diff
@@ -4,14 +4,22 @@ import 'package:flutter/material.dart';
...
-            dense: true,
-            horizontalTitleGap: 5,
-            leading: Icon(icon, size: 20),
-            title: Text(title),
-          ),
+              dense: true,
+              horizontalTitleGap: 5,
+              leading: Icon(icon, size: 20),
+              title: Text(title),
+              trailing: checkboxValue == null ? null : Checkbox(value: checkboxValue, onChanged: (_) => onCheckboxValueToggled?.call())),
```
hjiangsu, line 23: I think it might be better to pass in a `trailing` widget rather than `checkboxValue` and `onCheckboxValueToggled` into ThunderPopupMenuItem.

Doing this will allow us to place any arbitrary widget (e.g., icons, buttons, etc.) as part of the trailing end of this ListTile!

hjiangsu (approved):
Just one comment!

lib/shared/thunder_popup_menu_item.dart
micahmo: Good idea, done! In addition, I removed the usage of `Checkbox` and just passed an `Icon`.

Issue #322, opened 2023-07-11
[Feature Tracker]: Offline reading/action queue support


Fmstrat:
Since 2-column tablet mode was easier than expected, I'll be moving on (slowly, in my spare time) into the next big feature on my bucket list, offline synchronization for reading. This issue is for tracking that progress and for conversation on the implementation methodology. As it stands now, this will come in the form of many step-wise PRs. I'm sure there are things I'm missing, and this issue will be updated accordingly as we go. If you'd like to help support this long-term initiative, let me know, as I may be able to reorganize things in a less linear fashion.

# PR1: Online detection
Before anything else, we need a reliable way to detect if the user is online or not. This should occur in two stages:
- **Android/iOS reporting**: The first is via the standard OS report to determine if the device is online
- **Server polling**: The second is with an array of status booleans to indicate if the server for each account is accessible. We will need to take into account new installs with no account added yet, too.
- **Error messages**: For now, any clicks on buttons/etc should go to a central file, and from there push out a message that the system is offline, or conduct the action. This way we are prepped to handle a future queue of actions.

# PR2: Action queue
Next up will be the action queue. Right now, when a comment is made, an upvote occurs, or any other action is taken, it is submitted to the API immediately. To support offline use, this will need to be migrated in a few ways.
- **The queue**: The queue itself should be an array of actions against an array of servers, this way the queue can operate independently for each logged in account.
- **Background job**: A background job to monitor the queue will be required to actually make the API calls. This will need a way to report back to the UI if there are issues, and should have retry logic that pauses the queue processing for a particular server if there is an issue, which would take over for the above error messages.
- **UI use of queue**: Once those are in place, the UI can be migrated button-by-button until the queue is fully leveraged.

# PR3: In-memory storage
Use of in-memory storage for loaded data can happen before the official offline support. This would mean the system would need to store pulls from the API as objects in memory (including images), so they can later be migrated to local storage.
- **API response cache**: API responses should be stored in memory variables in a centralized file, so that this can be the staging ground for offline local-storage.
- **Image and URL cache**: In addition to the JSON data, image data will need a location in these central in-memory variables, too. An option to consider could also be caching the web pages via a browser call, too. This will likely morph as we get into it, since memory will be at a premium. It might need to wait until local storage support is implemented.

# PR4: Offline support
The final step is actual offline support. This requires multiple elements as well
[...]

PR #1266, opened 2024-04-02
Core: Migrate from `sqflite` to `drift` packages

hjiangsu:
## Pull Request Description

This PR migrates our use of `sqflite` to [drift](https://pub.dev/packages/drift). In short, Drift is a library that builds on top of SQLite to provide some really useful features, including:
- Convenience methods for writing database queries using dart and raw SQL
- Code gen which generates classes from tables
- and more as mentioned in the pub description

This is part of a bigger picture which will allow us to provide an offline experience (#322) and other improvements. While using the existing `sqflite` would work for these purposes, it's much more difficult to accomplish due to the fact that we have to write raw SQL queries, and create classes for each of the associated tables. Having to do this hinders the DX and makes it more tedious to accomplish.

I'm hoping that this solution will help mitigate those issues.

This migration will likely happen in multiple stages:
1. Remove usages of sqflite, and migrate all existing information to sqlite using drift. I'm not entirely sure if sqflite and sqlite are compatible with each other, so I opted to create a whole new database, and manually perform data migration. This is what this PR accomplishes.
2. Reduce code complexity. This means removing `Account`, `Favorite`, and `LocalCommunity`models in favour of the autogen classes provided by drift, and refactoring parts of the code to use the new autogenerated classes.
3. Once the databases have been migrated to sqlite, remove usages of sqflite from our codebase. This includes removing the sqflite dependencies. This step will likely occur in the future after a couple of versions
4. Work on additional improvements once we have this framework up. This could include things such as caching API calls into the database for offline use, improve the anonymous account experience, etc.

Update: I've created some tests to check proper migration

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/database/database.g.dart
+++ b/lib/core/database/database.g.dart
@@ -10,83 +10,52 @@ class $AccountsTable extends Accounts with TableInfo<$AccountsTable, Account> {
   $AccountsTable(this.attachedDatabase, [this._alias]);
   static const VerificationMeta _idMeta = const VerificationMeta('id');
   @override
-  late final GeneratedColumn<int> id = GeneratedColumn<int>(
-      'id', aliasedName, false,
-      hasAutoIncrement: true,
-      type: DriftSqlType.int,
-      requiredDuringInsert: false,
-      defaultConstraints:
-          GeneratedColumn.constraintIsAlways('PRIMARY KEY AUTOINCREMENT'));
-  static const VerificationMeta _usernameMeta =
-      const VerificationMeta('username');
+  late final GeneratedColumn<int> id = GeneratedColumn<int>('id', aliasedName, false,
+      hasAutoIncrement: true, type: DriftSqlType.int, requiredDuringInsert: false, defaultConstraints: GeneratedColumn.constraintIsAlways('PRIMARY KEY AUTOINCREMENT'));
+  static const VerificationMeta _usernameMeta = const VerificationMeta('username');
   @override
-  late final GeneratedColumn<String> username = GeneratedColumn<String>(
-      'username', aliasedName, true,
-      type: DriftSqlType.string, requiredDuringInsert: false);
+  late final GeneratedColumn<String> username = GeneratedColumn<String>('username', aliasedName, true, type: DriftSqlType.string, requiredDuringInsert: false);
   static const VerificationMeta _jwtMeta = const VerificationMeta('jwt');
   @override
-  late final GeneratedColumn<String> jwt = GeneratedColumn<String>(
-      'jwt', aliasedName, true,
-      type: DriftSqlType.string, requiredDuringInsert: false);
-  static const VerificationMeta _instanceMeta =
-      const VerificationMeta('instance');
+  late final GeneratedColumn<String> jwt = GeneratedColumn<String>('jwt', aliasedName, true, type: DriftSqlType.string, requiredDuringInsert: false);
+  static const VerificationMeta _instanceMeta = const VerificationMeta('instance');
   @override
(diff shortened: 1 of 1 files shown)


hjiangsu:
Alright, this should be ready now! I believe I've tested as many cases as I could (both on emulator and on my physical device).

Everything seems to look okay from what I can see. For now, I've opted to not remove the original database just in case it needs to be restored. The "Delete Local Database" action in Settings only deletes the new database (for now) as a way to test migration.

@micahmo If you are able to, could you try this out and let me know if you encounter any issues on your end? Thanks!

micahmo (approved):
The code LGTM! I did a very brief test (basically just running the app) and things seem to be working ok!

hjiangsu:
I'll go ahead and merge this in. I'll generate a nightly build with the current changes, and give a week for any last minute fixes before I finalize 0.3.0 for general release!

micahmo:
@hjiangsu I noticed one issue so far. If you use the temporary account switcher, it seems to lose the actual account afterwards on the account page. The drawer and switcher are still correct.

https://github.com/thunder-app/thunder/assets/7417301/24fea0d3-d33d-404a-9b83-210f0bb37522

Let me know if you want me to look at this (since I worked on the temp account switcher) or you (since you added the new db).

hjiangsu:
Hmm, I can take a quick look at this!

hjiangsu:
Update: I don't think this is directly related to the DB changes. It seems like it's an issue with the `reload` parameter whenever we switch accounts temporarily.

When we switch accounts, this line triggers in `on<SwitchAccount>`:
```dart
emit(state.copyWith(status: AuthStatus.loading, isLoggedIn: false));
```

Since `reload` isnt set here, `reload = true` by default and the account page changes to the non-logged in status (because of `isLoggedIn: false`)

When we restore the original user, `reload` is still set to `false` so the account page doesn't update. Note that the account page updates properly if we open the drawer before navigating to the account page

hjiangsu:
I believe the easiest solution for now would be to remove the `if (!state.reload) return;` checks on account_page.dart. This will cause some loading screens if we switch to the account page fast enough, but would ensure that the account page always reflects the correct user

micahmo:
Thanks so much for taking a look at this! If it's not related to the db upgrade then I can look further as well.

I think I had added `if (!state.reload) return;` to the account page because you can temporarily switch accounts from the account settings page, which is directly above the account page, meaning that navigating back would always show some weird switching.

Here's what it looks like now.

https://github.com/thunder-app/thunder/assets/7417301/362ebf2f-ae74-4fec-8718-b7a949464f76

And here's what it looks like with those lines. Not the end of the world, but not super clean either.

https://github.com/thunder-app/thunder/assets/7417301/cbcb114f-132f-485d-865f-329e47594608

Do you think it's worth my time to do any further investigation, or should we just remove those lines?

hjiangsu:
You can give it a shot if you'd like to find an alternate solution! I don't think its a particularly large issue since this use-case won't be triggered often, but having smoother transitions is always good.

micahmo:
Quick update! First, you are right that this doesn't seem to be related to the new db. I was able to repro in a commit before that change. It's just a very specific sequence of steps that I might have never tried before.

1. Launch app fresh.
2. Navigate to account page.
3. Navigate back to main feed.
4. Create post.
5. Change user posting as.
6. Close post page.
7. Navigate to account page.

Lastly, I think you were right that the initial `emit` for `refreshing` is probably what's causing problems. It seems like adding the `reload` parameter helps that case. I'll open a PR for this change!

PR #1268, opened 2024-04-03
Add timeout when loading image dimensions for image viewer

hjiangsu:
## Pull Request Description

> Review without whitespace

This PR adds a two second timeout for retrieving image dimensions on the image viewer. I've noticed that there are often times where some images don't load, or take a long time to load.

This makes it so that the image still loads if retrieving image dimensions fails (given its a proper image). The logic for retrieving image dimensions only affects the maximum zoom level, which is not a critical condition to meet. Thus, if an image fails to load their dimensions, we'll default to a max zoom level of 3.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/shared/image_viewer.dart
+++ b/lib/shared/image_viewer.dart
@@ -115,15 +115,23 @@ class _ImageViewerState extends State<ImageViewer> with TickerProviderStateMixin
   }
 
   Future<void> getImageSize() async {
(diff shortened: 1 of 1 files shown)


hjiangsu:
> I also wonder if there's some way for us to retrieve/calculate the dimensions after the fact and adjust the zoom level then.

I was also thinking about this, and was going to explore it further when we refactor the image viewer (most likely in 0.4.0)! For now, it's just a quick fix for 0.3.0 since I want to release a general build soon

PR #1269, opened 2024-04-03
Fix another back button issue

micahmo:
## Pull Request Description

This PR fixes another broken issue from #1251 and related to #1260. In the latter PR, I fixed some issues where the top-level back button handler tries to take precedence even when it's not at the top of the navigation stack. However, there's another case -- the profile page for the logged-in user -- where the back button handler technically is at the top of the navigation stack, but it still needs to defer to the back button handler on the user page. It looks like the easiest solution here is to disable the top-level handler when we're on the account page and let the account page handle scrolling to the main page via a delegate. Let me know if you agree!

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/af520c4b-503d-4f2a-9c3f-69284a2cecda

### After

https://github.com/thunder-app/thunder/assets/7417301/25cb604a-3a7e-4f6d-b4cb-3e7ab8053514

## Checklist

### DIFF
--- a/lib/account/pages/account_page.dart
+++ b/lib/account/pages/account_page.dart
@@ -9,7 +9,9 @@ import 'package:thunder/shared/primitive_wrapper.dart';
 import 'package:thunder/user/pages/user_page.dart';
 
 class AccountPage extends StatefulWidget {
-  const AccountPage({super.key});
+  final bool Function(int page) scrollToPage;
+
+  const AccountPage({super.key, required this.scrollToPage});
 
   @override
   State<AccountPage> createState() => _AccountPageState();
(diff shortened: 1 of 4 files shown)


hjiangsu:
I think there's a simpler way to achieve this behaviour without having to pass in callback functions.

We can use `stopDefaultButtonEvent` to determine if we should run a given BackButtonInterceptor handler.

```dart
// thunder_page.dart
FutureOr<bool> _handleBackButtonPress(bool stopDefaultButtonEvent, RouteInfo info) async {
  final bool topOfNavigationStack = ModalRoute.of(context)?.isCurrent ?? false;

  if (!topOfNavigationStack) return false;
  if (stopDefaultButtonEvent) return false; // Add this line

  // ... rest of code
}
```

Adding this seems to give the desired behaviour. Let me know if this change works! I made a quick PR with this change: https://github.com/thunder-app/thunder/pull/1293

hjiangsu:
If it works, I'll go ahead and close this PR in favour of the other one!

PR #1271, opened 2024-04-03
Slightly improve mentions

micahmo:
## Pull Request Description

This PR slightly improves mentions. There's probably more work that needs to go in there to make them more like replies (maybe just using `CommentReference`), but at least now the originating username follows standards from through the app.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|--------|--------|
| image | image | 

## Checklist


lib/l10n/app_en.arb
```diff
@@ -713,6 +713,10 @@
   },
   "inReplyTo": "In reply to {post} in {community}",
   "@inReplyTo": {},
+  "in_": "in",
```
hjiangsu, line 716: This is going to be a weird one, but I don't know if it makes sense to have this as part of a localization string. This is mainly due to language/grammar differences. For example, English may have connecting words like "in", "to", etc. but this may not apply to all languages.

I think the usual convention is to pass in the full sentence or context so that the proper grammar can be derived from that. (e.g., passing in localization for "{person} to community") but this doesnt necessarily work in this case since we have separate widgets.

I thought about this with the localization for "{user} to {community}" in the post body metadata but wasn't sure how to best approach this. 😅

micahmo: I totally hear what you mean. I've run into issues in the past where parts of sentences were localized, and the translator couldn't really make it work because the sentence structure would be totally different in their language.

In this case, we're always going to have the username and the community name, and I don't think creating a string with both parts, like `{username} in {communityname} would work well (even if it was a single widget) because we don't really want the translator to totally rearrange things or add a bunch of extra words.

If we localize just `in`, then the translator can try to come up with a word or phrase that makes the two parts connect. And if there's nothing really analogous in their language, they can either leave `in` or maybe even leave it blank. I think both of those approaches would be preferable to _not_ localizing this word, especially since it will work fine in many languages.

That was my thought process at least, but I'm curious to hear yours!

hjiangsu: Hmm, yeah - I'm not too sure on the best approach to take here. The idea of leaving the field blank is definitely something that I didn't think of, so maybe that could work here! Maybe in this case, we can try out the approach that you have and see how that goes 😄

micahmo: There are a couple other things to consider.

1. I believe this might be the only place we use `in`, and other places we use `to`, so maybe we want to agree on the terminology.
2. We could possibly use something like an [arrow](https://api.flutter.dev/flutter/material/Icons/arrow_right_rounded-constant.html) instead of a word. Or we could omit the word entirely.

I'm open to ideas!

hjiangsu: I think we can keep the original approach you had, and if anything, we can adjust it in the future!

CTalvio:
> This PR slightly improves mentions. There's probably more work that needs to go in there to make them more like replies (maybe just using CommentReference), but at least now the originating username follows standards from through the app.

Lemmy doesn't actually work with post body mentions, we could absolutely go ahead and just make mentions use the exact widgets that replies do.

PR #1274, opened 2024-04-04
Fix login issue

micahmo:
## Pull Request Description

This PR fixes an issue with switching to a new account after logging in, which is related to the recent db migration (#1266). It looks like the problem is that, in `on<LoginAttempt>` we still used the old `accountId` format (which was a `Uuid`) as the `active_profile_id` in the `sharedPreferences`. The fix is to grab the newly insert id from `insertAccount` and put that in `active_profile_id`.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/d0c4ef4e-3c5b-4ce8-b3f4-11e8895df7c0

### After

https://github.com/thunder-app/thunder/assets/7417301/f2f1fcb9-e845-4eb7-81a2-9dee107cd1bc

## Checklist

### DIFF
--- a/lib/account/models/account.dart
+++ b/lib/account/models/account.dart
@@ -20,13 +20,26 @@ class Account {
     this.userId,
   });
 
-  static Future<void> insertAccount(Account account) async {
+  Account copyWith({String? id}) => Account(
+        id: id ?? this.id,
+        username: username,
+        jwt: jwt,
+        instance: instance,
+        userId: userId,
+      );
+
+  static Future<Account?> insertAccount(Account account) async {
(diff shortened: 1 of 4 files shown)


hjiangsu:
Ahh, I can't believe I missed this during testing 🤦. You're completely right! I made changes to the primary key `id` to be an auto-increment integer rather than a UUID string.

I think this might also potentially affect adding new favorites then since that was previously using UUID strings for the primary key as well.

hjiangsu (approved):

PR #1280, opened 2024-04-06
Fix notification issue

micahmo:
## Pull Request Description

This PR fixes an issue with background notifications that was broken after the db upgrade. It seems the problem is related to the fact that the global `database` field is not initialized when we're invoked via the background task, because the `main` method doesn't run. I added a getter with a backing field to do lazy initialization. However, it might also be safe to initialize the field directly, so I'm open to suggestions here. Either way, this does seem to fix the issue.

| image |
| - |

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -57,7 +57,7 @@ void main() async {
   SharedPreferences prefs = (await UserPreferences.instance).sharedPreferences;
 
   File dbFile = File(join((await getApplicationDocumentsDirectory()).path, 'thunder.sqlite'));
-  
+
   if (!await dbFile.exists()) await migrateToSQLite(database);
 
   // Clear image cache
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -43,7 +43,8 @@ import 'package:flutter/foundation.dart';
 import 'package:thunder/utils/notifications.dart';
 import 'package:thunder/utils/preferences.dart';
 
-late AppDatabase database;
+AppDatabase get database => _database ??= AppDatabase();
+AppDatabase? _database;
 
 void main() async {
   WidgetsBinding widgetsBinding = WidgetsFlutterBinding.ensureInitialized();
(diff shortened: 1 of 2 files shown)


lib/main.dart
```diff
@@ -43,7 +43,8 @@ import 'package:flutter/foundation.dart';
 import 'package:thunder/utils/notifications.dart';
 import 'package:thunder/utils/preferences.dart';
 
-late AppDatabase database;
+AppDatabase get database => _database ??= AppDatabase();
```
hjiangsu, line 46: This might pose an issue if a user is upgrading to the new db. This is because the main function checks if the new db file exists and runs the migration if it doesn't detect the new db.

`File dbFile = File(join((await getApplicationDocumentsDirectory()).path, 'thunder.sqlite'));`

If we move the initialization here, it might create an empty db before the check runs which would skip the migration step.

micahmo: Good point! And I guess that could happen if they update the app but don't run it, and then get a notification.

Any other ideas? Maybe we can skip notification checks if the db isn't upgraded, but we still need a way to initialize it, if it is.

micahmo: Alright, I pushed another change. It reverts these changes and adds a separate method that anyone can call to ensure proper database initialization and migration. If that has already occurred (i.e., in `main`) it's a no-op. The notifications code can call this, knowing `main` might not have happened. Thoughts?

My main concern is that this is sort of a big code change at this point in the release. But I do think we need to fix the notifications issue.

hjiangsu: Just took a look at the code, and it seems to look okay! I haven't tested it out myself so I'l have to do that. We should probably also test going from 0.2.9 stable to make sure everything works!

hjiangsu (approved):

hjiangsu:
Just tested it and it seems to be okay - I'll merge this in and make a new nightly build to test this out further!

Issue #1283, opened 2024-04-08
Pull to refresh not always working


Offerel:
### Bug Description

Currently im in vacation and internet connection is really poor and unstable. Often I can't reach the lemmy instance directly after starting the app. If this is the case I get a warning that the choosed instance isn't available,  which is expected. 

But I can't refresh now,  since pull to refresh seems not to be implemented in this rare usecase. 

It would be nice if you could enable either pull to refresh here to or implement a simple button to retry.

### Expected Behaviour

Refresh possible when I can't resolve instance 

### Steps to Reproduce

1. Cut down internet 
2. Open the app
3. Host can't be resolved 
4. Reconnect internet 
5. Try to refresh 

### App Version

3.0.6

### Device

S23

### OS

Android 14

PR #1284, opened 2024-04-08
Fix initial load issue

micahmo:
## Pull Request Description

This PR fixes an issue where there's not an easy way to retry loading the feed if it times out due to a network issue. Note that in addition to changing the `ErrorMessage` action from "Account Settings" to "Retry", I also made the action button disabled and show a spinner when it is pressed. I did this because the main page doesn't react to `AuthStatus.loading`, so there was no other indication that any action had been taken. I'm not sure exactly why that was, but I have a feeling it had something to do with the main page indicator and the feed page's indicator not quite aligning together, and most of the time only the latter is needed.

> Review without whitespace.

## Issue Being Fixed

Issue Number: #1283

## Screenshots / Recordings

The demo shows first an unsuccessful retry attempt, and then a successful one.

https://github.com/thunder-app/thunder/assets/7417301/754a0835-30a1-4b1a-b173-6f5e5f8ed04d

## Checklist

### DIFF
--- a/lib/shared/error_message.dart
+++ b/lib/shared/error_message.dart
@@ -39,24 +39,42 @@ class ErrorMessage extends StatelessWidget {
             ),
             const SizedBox(height: 32.0),
             if (actions?.isNotEmpty == true)
-              Row(
+              Column(
                 children: [
-                  for (var action in actions!) ...[
-                    if (actions!.indexOf(action) > 0) const SizedBox(width: 10),
-                    Expanded(
-                      child: ElevatedButton(
-                        style: ElevatedButton.styleFrom(minimumSize: const Size.fromHeight(50)),
-                        onPressed: action.loading == true ? null : () => action.action.call(),
-                        child: action.loading == true
-                            ? const SizedBox(
-                                width: 20,
-                                height: 20,
-                                child: CircularProgressIndicator(),
-                              )
-                            : Text(action.text),
-                      ),
+                  ElevatedButton(
+                    style: ElevatedButton.styleFrom(minimumSize: const Size.fromHeight(50)),
+                    onPressed: actions![0].loading == true ? null : () => actions![0].action.call(),
+                    child: actions![0].loading == true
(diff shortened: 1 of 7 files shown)


micahmo:
Good call! I've made `ErrorMessage` more generic so it can take a list of actions. Let me know what you think!

https://github.com/thunder-app/thunder/assets/7417301/f61ebd18-79d5-469e-bd62-4fb221cd7283

hjiangsu:
I just have one small UI tweak - can we make it so that the buttons are in separate rows? Also, I think making one button the "primary" action and the other a "secondary" action would help with overall UI/UX!

For the primary/secondary, I'd suggest the following:
- Primary button uses the current look (elevated) - this can be the "Retry" option
- Secondary button uses a `TextButton` - this one can be the "Account Settings" option

micahmo:
> can we make it so that the buttons are in separate rows?

I actually tried this first and liked side-by-side better, but I'll put it back!

> Secondary button uses a TextButton - this one can be the "Account Settings" option

Alright, I went ahead and pushed this. It looks below. Is that ok?

| image |
| - |

lib/shared/error_message.dart
```diff
@@ -4,10 +4,14 @@ import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 class ErrorMessage extends StatelessWidget {
   final String? title;
   final String? message;
-  final String? actionText;
-  final VoidCallback? action;
+  final List<({String text, void Function() action, bool loading})>? actions;
```
hjiangsu, line 7: Interesting use of records!

hjiangsu (approved):
LGTM! Just double checking, this is ready to be merged?

micahmo:
> this is ready to be merged?

If the last screenshot looks good to you, yes!

lib/shared/error_message.dart
micahmo: Yeah it's close to needing a class. 🫣

PR #1286, opened 2024-04-08
Improve spoiler appearance

micahmo:
## Pull Request Description

This PR aims to improve the appearance of spoilers. While the functionality has been working perfectly, I've noticed many times that it's easy to miss a spoiler, or for it to get lost in the larger comment/post body. The only difference in the text is that it's bold (which is not unusual in posts). And since the expander icon aligned to the right, it's very easy to miss (at least for me). Finally, when the spoiler is open, it can be hard to tell which text is actually part of the spoiler as opposed to the rest of the post (this is a problem in the Lemmy web UI too). You can see in the before screenshot that spoilers have a tendency to blend in with the rest of the text, and it's hard to correlate the icon on the right with the specific heading.

Here is my proposed change:
* Move the expander icon to the left so that it is immediately visible.
* Highlight the spoiler heading (and body, once expanded) with a faint background color.
* Tweak the icons to more closely match Lemmy UI.

One complaint about this change may be that the spoiler heading no longer nicely aligns with the rest of the body.

I'm totally open to feedback here, and I'd be curious to hear @machinaeZER0 chime in as well!

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/69b2244c-d9d9-4079-b1b2-9e7bc1d2e888

### After

https://github.com/thunder-app/thunder/assets/7417301/51f1d4b1-1769-407e-a6a4-58e25365a71f

## Checklist

### DIFF
--- a/lib/community/widgets/community_sidebar.dart
+++ b/lib/community/widgets/community_sidebar.dart
@@ -112,7 +112,6 @@ class _CommunitySidebarState extends State<CommunitySidebar> {
                             child: CommonMarkdownBody(
                               body: communityView.community.description ?? '',
                               imageMaxWidth: (kSidebarWidthFactor - 0.1) * MediaQuery.of(context).size.width,
-                              allowHorizontalTranslation: false,
                             ),
                           ),
                           const SidebarSectionHeader(value: "Stats"),
--- a/lib/instance/widgets/instance_view.dart
+++ b/lib/instance/widgets/instance_view.dart
@@ -68,7 +68,6 @@ class InstanceView extends StatelessWidget {
         const Divider(),
         CommonMarkdownBody(
           body: site.sidebar ?? '',
-          allowHorizontalTranslation: false,
         ),
       ],
     );
--- a/lib/shared/common_markdown_body.dart
+++ b/lib/shared/common_markdown_body.dart
@@ -35,8 +35,6 @@ class CommonMarkdownBody extends StatelessWidget {
 
   final double? imageMaxWidth;
 
-  final bool allowHorizontalTranslation;
-
   const CommonMarkdownBody({
     super.key,
     required this.body,
@@ -44,7 +42,6 @@ class CommonMarkdownBody extends StatelessWidget {
     this.isSelectableText = false,
     this.isComment,
     this.imageMaxWidth,
-    this.allowHorizontalTranslation = true,
   });
 
   @override
@@ -121,7 +118,7 @@ class CommonMarkdownBody extends StatelessWidget {
       extensionSet: customExtensionSet,
       inlineSyntaxes: [LemmyLinkSyntax(), SubscriptInlineSyntax(), SuperscriptInlineSyntax()],
       builders: {
-        'spoiler': SpoilerElementBuilder(allowHorizontalTranslation: allowHorizontalTranslation),
+        'spoiler': SpoilerElementBuilder(),
         'sub': SubscriptElementBuilder(),
         'sup': SuperscriptElementBuilder(),
       },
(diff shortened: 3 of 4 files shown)


machinaeZER0:
Oh wow, that's much nicer! I really, really like that :D 

Some clarifying questions:
- We had formatted spoilers in a specific way (not just blacking out the text) in part because people were using it in their instance/community sidebars, right? Or was that a different scenario that I'm mixing up? If I am thinking of the right thing, will this change affect that use case, and does the implementation look good to you?
- Can spoiler tags be used in post titles (in Lemmy in general)? If so, how is that handled currently/is it affected in any way by this tweak?
- How do we currently handle spoilers that are not broken onto a separate line (a few words spoiler tagged within a sentence, for example)? If it's handled through this functionality, how does it look with this new implementation?

The above bullets would be the things that immediately jump out to me as things we would want to take into account, if any of them are covered by this tweak. But again, in the example as shown above I think you smashed this out of the park on every front - better in every regard :)

micahmo:
> Oh wow, that's much nicer! I really, really like that :D

Thanks for the positive feedback!

> We had formatted spoilers in a specific way (not just blacking out the text) in part because people were using it in their instance/community sidebars, right? Or was that a different scenario that I'm mixing up? If I am thinking of the right thing, will this change affect that use case, and does the implementation look good to you?

Yes you are thinking of the right thing. Community and instance sidebars were a major consideration in the original design. I suppose the sidebars are arguably a little slightly less clean (mostly because there are often multiple spoilers consecutively, not mixed with other text) but I still think we get the same benefits. Thoughts?

| image |
| - |

> Can spoiler tags be used in post titles (in Lemmy in general)? If so, how is that handled currently/is it affected in any way by this tweak?

I don't believe they are allowed because a title cannot be multi-line, so I think we are all set! (Although some markdown is allowed in titles, and I don't think we handle it properly, but that's a different issue.)

> How do we currently handle spoilers that are not broken onto a separate line (a few words spoiler tagged within a sentence, for example)? If it's handled through this functionality, how does it look with this new implementation?

I do not believe inline spoilers are part of the Lemmy markdown syntax. I think we had our own version of line spoilers at one point, but abandoned them because they wouldn't recognized by any other app/UI. @hjiangsu can refresh my memory there.

Thanks again!

micahmo:
Just FYI, I pushed another commit which removes the horizontal adjustment and adds internal padding. If we go with this approach, it won't matter nearly as much if the text left-aligns with the rest of the post body, since it is intended to be set apart anyway. Here's what it looks like.

| Post | Sidebar |
|--------|--------|
| image | image |

machinaeZER0:
I like that tweak a lot! @hjiangsu any callouts on your end?

hjiangsu:
> Can spoiler tags be used in post titles (in Lemmy in general)? If so, how is that handled currently/is it affected in any way by this tweak?

Yup, can confirm that spoilers do not work on post titles!

> How do we currently handle spoilers that are not broken onto a separate line (a few words spoiler tagged within a sentence, for example)? If it's handled through this functionality, how does it look with this new implementation?

Yeah, as @micahmo mentioned, inline spoilers are not supported by Lemmy. I previously implemented an inline spoiler, but removed it promptly because we want to ensure compatibility with Lemmy syntax.

> Just FYI, I pushed another commit which removes the horizontal adjustment and adds internal padding

I like this change! It looks much better and definitely helps distinguish the spoiler a bit more 😄

hjiangsu (approved):
LGTM! @micahmo let me know if this is ready to be merged.

micahmo:
This is good to go!

Issue #744, opened 2023-09-15
Ability to block posts and communities that are set to a different language


hjiangsu:
**Is your feature request related to a problem? Please describe.**
Its a QoL feature which reduces the number of posts/communities which you are most likely not going to interact with because they are in a different language.

**Describe the solution you'd like**
My thoughts on how this would be implemented:
- The languages will be pulled from your account settings (whatever you set up on your instance)
- When fetching posts (e.g., on feeds), we can check for the post language
- When fetching communities (e.g., on searches), we can check for the community language

This should be an option that we select in the Settings page.

**Describe alternatives you've considered**
N/A

**Additional context**
Related: https://lemmy.world/post/5068121

iparks99:
> When fetching communities (e.g., on searches), we can check for the community language

Is this possible? I don't see a language id property on [communities](https://lemmy.readme.io/reference/get_community).

I think filtering posts by languages is useful in itself, but if you have an idea of how to filter communities by language please let me know!

> The languages will be pulled from your account settings (whatever you set up on your instance)

Any tips on this for anonymous users? Or would this feature just not be available to them? I can't seem to find a list of languages for the language input modal that doesn't come from the instance _and_ has language ids.

hjiangsu:
> Is this possible? I don't see a language id property on [communities](https://lemmy.readme.io/reference/get_community).

Communities have a property called `discussion_languages` which holds an array of integers corresponding to languages. I think we can use this to determine the allowed languages for a given community. However, I'm not too sure if the API automatically filters out communities if we set the discussion languages in the user settings - we would have to do a test to see what the behaviour is here!

> Any tips on this for anonymous users? Or would this feature just not be available to them?

I mentioned it in here: https://github.com/thunder-app/thunder/pull/1289#issuecomment-2047929386 but we can limit the scope of this to logged in users for now! It would be possible to do it for anonymous accounts, but it might be a bit more complicated (and might be more suited as a separate PR)

hjiangsu:
I'll assign this to you since you're working on it but do let me know if you'd like me to remove the assignment!

PR #1289, opened 2024-04-09
Support for discussion languages

iparks99:
## Pull Request Description

I added an additional section to the Account Settings -> General for discussion languages. Posts that are not of the filtered languages will not be displayed in the user's feed.

This is my first PR, and one of my first projects in Dart/Flutter. Any advice and tips are greatly appreciated!

## Issue Being Fixed

As described [here](https://github.com/thunder-app/thunder/issues/744), we'd ideally be able to filter out posts/communities in languages we don't understand.

Issue Number: #744

## Screenshots / Recordings

account_settings
add_language
remove_language
remove_undetermined

## Checklist

- [x] Pull languages from account settings
- [x] Ability to set discussion languages
- [x] Filter posts in feed by language

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -60,6 +60,10 @@ enum LocalSettingsSubCategories {
 }
 
 enum LocalSettings {
+  /// -------------------------- Account Settings --------------------------
+  // Discussion Languages
+  discussionLanguages(name: 'account_discussion_languages', key: 'discussionLanguages', category: LocalSettingsCategories.account),
+
   /// -------------------------- Feed Related Settings --------------------------
   // Default Listing/Sort Settings
 
@@ -408,6 +412,7 @@ extension LocalizationExt on AppLocalizations {
       'feedTypeAndSorts': feedTypeAndSorts,
       'profiles': profiles,
       'animations': animations,
+      'discussionLanguages': discussionLanguages,
     };
 
     if (localizationMap.containsKey(key)) {
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -473,7 +473,7 @@
   "@deleteLocalPreferencesDescription": {
     "description": "Description for confirmation action to delete local preferences"
   },
-  "deselectUndeterminedWarning": "Warning: If you deselect Undetermined, you will not see most content.",
+  "deselectUndeterminedWarning": "If you deselect Undetermined, you will not see most content.",
   "@deselectUndeterminedWarning": {},
   "detailedReason": "Reason: {reason}",
   "@detailedReason": {
@@ -491,6 +491,8 @@
   "@discussionLanguages": {
     "description": "Only load posts and communities in your language(s)"
   },
+  "discussionLanguagesTooltip": "Content is filtered to the selected languages.",
+  "@discussionLanguagesTooltip": {},
   "dismissRead": "Dismiss Read",
   "@dismissRead": {},
   "displayUserScore": "Display User Scores (Karma).",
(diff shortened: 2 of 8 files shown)


iparks99:
Thanks for the feedback @hjiangsu! Hopefully just finished getting this working based on your comments. Let me know what you think when you have a moment.

iparks99:
Sorry for last minute force pushed rebases and all the associated actions!

micahmo (commented):
The code LGTM, nice work, especially for a first contribution!!!

I had some thoughts regarding the UI. Note that these are just my opinions and you're not obligated to implement any of them. 😊
1. I was wondering if the "Discussion Languages" heading should more closely align with the Block Instances/Users/Communities, since the UX is otherwise very similar. Along the same lines, the placeholder text "No discussion languages" looks slightly different than the placeholder text for the other sections, and the same is true of the list of languages. Maybe they could have an X on the right to quickly remove?
2. Alongside the placeholder text, would it be useful to say something like, "If no languages are selected, content from all languages will be displayed"? Maybe that's a bit verbose, but it could help users understand how the feature works.
3. When I added and then removed "Undetermined", I got the message "Warning: If you deselect Undetermined, you will not see most content." I realize that the wording comes from the Lemmy UI, but it's a bit misleading when you have no other languages selected, because now you will see all content. Maybe you could only show that warning if there is at least one other language selected?
4. I thought it might be neat to have a quick link in the Filters page that can bring you right to the Discussion Languages setting. See Appearance > Comments > "Looking for username formatting?" as an example.

BTW, we usually squash PRs, so you don't have to worry quite as much about your branch history looking good. I force push a lot when I'm working locally, but once I make a PR I try to just commit so others can see the progression of the PR.

iparks99:
@micahmo okay, I think all the feedback has been addressed! Let me know what you think!

micahmo:
@iparks99 It looks great!! Thank you so much for addressing all the feedback! 🎉

I only noticed one small glitch. After acknowledging the "Undetermined" warning, the dialog doesn't seem to disappear.

https://github.com/thunder-app/thunder/assets/7417301/89609669-07c1-4056-be76-e2fac0531546

iparks99:
@micahmo Fixed! Thanks for pointing that out, I must've missed the Navigator pop when moving that code block around.

micahmo (approved):
Nice work!

hjiangsu:
Sorry for getting back to this a bit late! I think it looks really good - thanks for doing this @iparks99 😄

I tested it out, and it seems to be working pretty well! Just had a couple of interesting observations:
- If I add a single language to the discussion languages, it always adds Undetermined. I was trying to figure out why this was the case, but it seems to be related to a Lemmy issue: https://github.com/LemmyNet/lemmy/pull/2851

https://github.com/thunder-app/thunder/assets/30667958/6fdfc4a2-7b1e-4004-b468-936a7051f331

- If I have both _Undetermined_ and another language (e.g., English), I can't remove both of them at once. I must remove English first, refresh the page, and then remove Undetermined in order to clear the list. Again, I don't think this is related to us, but its more-so a Lemmy issue.

https://github.com/thunder-app/thunder/assets/30667958/3134c211-0df6-4146-aa14-f229cbc86bd5

- Within the language dropdown, there is an option for "No language", which isn't actually a valid option. Would it be possible to hide that entry from the list?


---

This being said, I think the only actionable thing we can do here is hide the "No language" entry. Aside from that, I think this is good to go!
- For this, I would suggest adding a parameter to `showLanguageInputDialog` to exclude languages by passing in an array of language ids
```dart
showLanguageInputDialog(
  context,
  title: "title",
  excludedLanguageIds: [-1], // Pass in array of language ids to filter out from the resulting list
  onLanguageSelected: () {},
)
```

iparks99:
@hjiangsu Just added the parameter :)

hjiangsu (approved):

hjiangsu:
Looks good! Thanks again 😄

PR #1292, opened 2024-04-10
Apply suggested fixups

micahmo:
## Pull Request Description

When integrating the i18n plugin, I also added the following option to VS Code's `settings.json` file.

``` json
"editor.codeActionsOnSave": {
    "source.fixAll": "always",
}
```

This means that any suggested fixups will be applied when a file is saved. I think this is a really nice time-saver. Unfortunately, it means that if any existing files have suggested fixups, they will be fixed every time you save, even if you don't want to include those changes with your current work.

This PR takes care of a few instances that I've been running into a lot. With these fixups applied, there shouldn't be further unwanted changes in these files upon saving.

### DIFF
--- a/lib/inbox/bloc/inbox_bloc.dart
+++ b/lib/inbox/bloc/inbox_bloc.dart
@@ -65,7 +65,7 @@ class InboxBloc extends Bloc<InboxEvent, InboxState> {
     int limit = 20;
 
     try {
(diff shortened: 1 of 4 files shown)


hjiangsu (approved):
Looks good! I was encountering this issue as well so thanks for doing this 😄

PR #1277, opened 2024-04-05
chore: configure `i18n.ally` plugin

ipcjs:
## Pull Request Description

Configuration `i18n.ally` plugin, support writing the multi-language characters in the form of `l10n.key`

## Issue Being Fixed

Close https://github.com/ipcjs/i18n-ally/pull/3

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1294, opened 2024-04-10
Developer QOL: Integrate Run on Save.

micahmo:
## Pull Request Description

This PR is a followup to #1264 and #1277. Using the [Run on Save](https://marketplace.visualstudio.com/items?itemName=emeraldwalk.RunOnSave) extension, we can essentially close the loop with regards to quickly and easily adding new translations. Run on Save allows us to execute an arbitrary script upon saving a matching set of files. I have it set up to match `.arb` files and perform the following steps.
* Activate `arb_utils` package (should be a no-op if already installed)
* Run `arb_utils generate-meta` (ensures that all keys have the metadata key, which is *not* added automatically by `i18n-ally`)
* Run `arb_utils sort` (self-explanatory)
* Run `flutter gen-l10n`

If there are errors with any of these steps (like the recent issue with activating `arb_utils`) it should not affect the developer. If needed, the logs can be found in the Output window.

As before, these new entries in `settings.json` only affect developers using VS Code and only those with the Run on Save extension installed.

P.S. Can you verify that these commands run successfully on Mac as well?

## Screenshots / Recordings

Here is a full demo of the end-to-end workflow. You can see how easy things are now! :blush:

https://github.com/thunder-app/thunder/assets/7417301/fe051c12-ca0a-4ef3-a175-3fcc4641a35e

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,6 @@
 {
     "recommendations": [
-        "ipcjs.i18n-ally-ipcjs"
+        "ipcjs.i18n-ally-ipcjs",
+        "emeraldwalk.runonsave"
     ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,5 +36,14 @@
                 "l10n.{key}({args})"
             ]
         }
-    ]
+    ],
+    "emeraldwalk.runonsave": {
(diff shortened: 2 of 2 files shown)


hjiangsu (approved):
LGTM! Tested it on Mac and it seems to be running well

Issue #1002, opened 2023-12-23
Add "Scaled" to the "Default Sort" setting


K4LCIFER:
**Is your feature request related to a problem? Please describe.**
The default sort type setting does not contained the option for "scaled"
![Screenshot_20231222-160236](https://github.com/thunder-app/thunder/assets/7829231/d0895ba7-d072-4f46-9d48-df3434118fd0)

**Describe the solution you'd like**
Add the "Scaled" sort-type to the default sort type option.

**Describe alternatives you've considered**
None.

**Additional context**
None.

micahmo:
Previous discussion for reference! Thanks!

https://github.com/thunder-app/thunder/pull/829#discussion_r1365879707

hjiangsu:
I'll just add this in here: https://fedidb.org/software/lemmy

I believe we should wait a bit longer until the majority of instances have migrated to 0.19.x before adding the new sorts as options. Once the top 30~40 instances have migrated to 0.19.x, we can enable these options in the settings. This should cover the majority of users.

K4LCIFER:
> I'll just add this in here: https://fedidb.org/software/lemmy
> 
> I believe we should wait a bit longer until the majority of instances have migrated to 0.19.x before adding the new sorts as options. Once the top 30~40 instances have migrated to 0.19.x, we can enable these options in the settings. This should cover the majority of users.

Would it be worth it to just have Thunder check what version the server is running, or check if the server offers a scaled sort, and then dynamically change the UI depending on what the server offers?

hjiangsu:
> Would it be worth it to just have Thunder check what version the server is running, or check if the server offers a scaled sort, and then dynamically change the UI depending on what the server offers?

The one issue here is that the sort type is global across all instances/servers, and it is not synced per account. This means that if a user chooses one of the new sort types, there should be a secondary option to fallback to if the instance does not support the new sort type. Doing this would take a bit more work on our part and it would be a short term solution (that would be removed once most instances have migrated to 0.19.x)

I believe the better approach here (and feel free to let me know if you think otherwise) is to make it so that the default sort type is synced with your account (Lemmy) settings since that is already part of Lemmy itself. Then, we could re-use the existing sort type setting to be used for anonymous (non-logged in) accounts.

orangeNgreen:
Now that most instances have upgraded to 0.19.x, it might be a good time to implement Scaled sort as a default setting.

daviessm:
[...]

PR #1298, opened 2024-04-11
Add support for new sort types as defaults

micahmo:
## Pull Request Description

The PR adds "Scaled" and "Controversial" to the possible default feed sorting options, and "Controversial" to the possible comment sorting options. If connected to an instance that doesn't support these types, Thunder's built-in defaults will be applied (Hot and Top, respectively). Along with these changes I added several new helpers to the `LemmyClient` class, such as a getter for the current version, and the ability to compare an arbitrary version against a feature. This turned out to be a bigger effort than I was expecting, so I think now (early in the release cycle) is a great time to get it in and tested!

## Issue Being Fixed

Issue Number: #1002

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/e83831e2-31ae-4581-969c-42ee822286a1

## Checklist

### DIFF
--- a/lib/community/widgets/community_drawer.dart
+++ b/lib/community/widgets/community_drawer.dart
@@ -177,7 +177,7 @@ class FeedDrawerItems extends StatelessWidget {
                 isSelected: destination.listingType == feedState.postListingType,
                 onTap: () {
                   Navigator.of(context).pop();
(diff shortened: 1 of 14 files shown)


hjiangsu (approved):
LGTM! This should hopefully make version-restricted features easier to implement!

Issue #1221, opened 2024-03-21
Option to show user profile pictures in post view


Fxsch:
### Feature Description

There's already the option to show the community profile picture on posts in your feed but when I click on a post, I would like to be able to see the profile pictures of the post author and the commenters.

iparks99:
Great idea, I'll start working on this!

glyph-se:
Is this a duplicate of https://github.com/thunder-app/thunder/issues/600 ?

Issue #600, opened 2023-08-09
Option to display User Avatar


MetalZealot:
**Is your feature request related to a problem? Please describe.**
Sometimes it's easier to discern a user by their Avatar, especially when scanning for small text and without any colour differentiation. Displaying their Avatar to the left of their user name would help with this

**Describe the solution you'd like**
implement a togglable option to Display User Avatar

Toldry:
Bump.

@MetalZealot perhaps add the "enhancement" tag to this issue?

PR #1281, opened 2024-04-06
Refactor post metadata and user/community chips

hjiangsu:
## Pull Request Description

This PR refactors the post view's metadata (comment count + date/time), and the post view username/community information.
- Removes the old `PostCardMetaData` widget
- Refactors `PostViewMetaData` to use the same widgets as the post card (for comment count, date/time)
- Refactored the logic for determining special users and colours.
- New widgets have been created for user/community chips which apply special colours and icons (similar to how comment header and post page show a user's special status)
- Fixed an issue where the full name widgets were not applying the proper instance colour - separate PR for this #1282 

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

Issue #675, opened 2023-08-21
Data saving mode


sombresu:
**Is your feature request related to a problem? Please describe.**
Limited data plan on a third world country. Also as soon as the app is opened, It uses so much data.

**Describe the solution you'd like**
Reduce image preview 
Not auto buffering videos an gifs
Reduce video resolution

CTalvio:
This will be difficult. Lemmy does not currently provide lower quality versions of media for a lot of stuff, as it is often not even hosted on the instance. As such, its not possible to just "reduce resolution" of content to save data use.

hjiangsu:
Yeah, as @CTalvio mentioned, it would be a bit difficult to reduce data usage down at this point in time because there's no option to change the quality of the images/media that is retrieved from lemmy.

One way to reduce data consumption is to reduce the amount of posts that get loaded in, or to do some further optimizations on our caching and network calls. However, this will take some time to do and this will only continue to get better over time as we continue working on Thunder! Reducing the amount of posts that get loaded in is also fairly minimal in terms of data usage so it would not improve data consumption much.

An extreme alternative would be to host our own servers which downscale the images and act as a proxy between the app and the lemmy instance. However, this brings in a whole new level of complexity and issues which would be best to avoid at this point in time.

CTalvio:
It might be worth adding a setting to not load images for use in thumbnails at all.

hjiangsu:
That's a good idea, that would definitely help alleviate data usage. Maybe this can be grouped as a setting for "Data Saving Mode" along with any other tweaks such as loading less posts/comments/etc.

@micahmo @ajsosa thoughts on other suggestions for lowering data usage?

CTalvio:
I have discovered, that at least with images hosted on a lemmy instance, we in fact can request lower resolution images.

This image hosted on sopuli.xyz, can be requested in a variety of resolutions by using `?thumbnail=512&format=webp` as a query string.

![](https://sopuli.xyz/pictrs/image/cd319103-2136-48d5-a326-fb0afed527b0.png?thumbnail=64&format=webp)

![](https://sopuli.xyz/pictrs/image/cd319103-2136-48d5-a326-fb0afed527b0.png?thumbnail=128&format=webp)

![](https://sopuli.xyz/pictrs/image/cd319103-2136-48d5-a326-fb0afed527b0.png?thumbnail=256&format=webp)

[...]

PR #1300, opened 2024-04-12
User avatar support for comments and post views

iparks99:
## Pull Request Description

Added toggle options to Settings -> Appearance -> Comments -> Show User Avatar and to Settings -> Appearance -> Posts -> Post Body Settings -> Show User Avatar.

If these settings are toggled on, then when a user views a post, the poster and commenters will have their avatars appear to the left of their display names.

## Issue Being Fixed

Enhancement to display user avatars in post view.

Issue Number: #1221, #600

## Screenshots / Recordings

image
image
Screenshot from 2024-04-12 16-04-25

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -135,6 +135,7 @@ enum LocalSettings {
   // Post body settings
   showCrossPosts(name: 'setting_show_cross_posts', key: 'showCrossPosts', category: LocalSettingsCategories.posts, subCategory: LocalSettingsSubCategories.general),
   postBodyViewType(name: 'setting_general_post_body_view_type', key: 'postBodyViewType', category: LocalSettingsCategories.posts, subCategory: LocalSettingsSubCategories.general),
+  postBodyShowUserAvatar(name: 'setting_general_post_body_show_user_avatar', key: 'postBodyShowUserAvatar', category: LocalSettingsCategories.posts, subCategory: LocalSettingsSubCategories.general),
   postBodyShowUserInstance(
       name: 'setting_general_post_body_show_user_instance', key: 'postBodyShowUserInstance', category: LocalSettingsCategories.posts, subCategory: LocalSettingsSubCategories.general),
   postBodyShowCommunityInstance(
@@ -156,6 +157,7 @@ enum LocalSettings {
   showCommentActionButtons(
       name: 'setting_general_show_comment_button_actions', key: 'showCommentActionButtons', category: LocalSettingsCategories.comments, subCategory: LocalSettingsSubCategories.general),
   commentShowUserInstance(name: 'settings_comment_show_user_instance', key: 'showUserInstance', category: LocalSettingsCategories.comments, subCategory: LocalSettingsSubCategories.comments),
+  commentShowUserAvatar(name: 'settings_comment_show_user_avatar', key: 'showUserAvatar', category: LocalSettingsCategories.comments, subCategory: LocalSettingsSubCategories.comments),
   combineCommentScores(name: 'setting_general_combine_comment_scores', key: 'combineCommentScores', category: LocalSettingsCategories.comments, subCategory: LocalSettingsSubCategories.comments),
   nestedCommentIndicatorStyle(
       name: 'setting_general_nested_comment_indicator_style', key: 'nestedCommentIndicatorStyle', category: LocalSettingsCategories.comments, subCategory: LocalSettingsSubCategories.comments),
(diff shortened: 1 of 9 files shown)


iparks99:
@micahmo is there a better way to get the user avatars than I did here? Perhaps it's unfounded, but I'm worried I would be excessively straining the instance by pulling so many user avatars for bigger comment trees. Is it possible to fetch smaller sized avatar files from the instance?

hjiangsu:
Thanks for taking on this!

> Is it possible to fetch smaller sized avatar files from the instance?

There is a way to query for a resized image, given that the image is hosted on the instance. To do so, I believe you have to append a query string at the end of the URL. For example:

```
https://sopuli.xyz/pictrs/image/cd319103-2136-48d5-a326-fb0afed527b0.png?thumbnail=256&format=png
```



In this case, the relevant parameters are `thumbnail=512&format=webp`
- `256` corresponds to the re-sized width/height in pixels. I'm not entirely sure if it corresponds to the width, or the height so we might have to test that out.
- `format=png` corresponds to the format of the image that we want to get back. This parameter isn't necessary, but its useful if we want to enforce a specific file type

More information about this is present in https://github.com/thunder-app/thunder/issues/675. Aside from that, `UserAvatar` uses `CachedNetworkImage` which should help reduce requests if we're requesting the same images!

---

For this to work, I'd imagine that we would have to do the following:
- Add a parameter to `UserAvatar` indicating the size of the image/thumbnail. This parameter should be nullable so that we can retrieve the full size image if necessary
- If that parameter is passed in, check to see if the URL is a `pictrs/image` endpoint. This might be as simple as checking if the URL contains `/pictrs/image/`

iparks99:
Thanks for the detailed info @hjiangsu! I've added two optional parameters to `UserAvatar`: `int? thumbnailSize` and `String? format`. The code block I added only adds these parameters to the query if they are present and the avatar URI contains `/pictrs/image/`. I tested some of the URIs it output and they're definitely thumbnail sized :)

micahmo (approved):
Hey, sorry for taking so long to get back to this! I just wanted to say that this is an awesome implementation. The only super minor UI feedback I have is that there seems to be a bit of space to the left of the avatar. This is especially obvious when it's a "special" user, or when you see the inkwell. Maybe that could be shrunk down to about the same amount as between the avatar and the text.

| image |
| - |

P.S. I would love to see this implementation optionally extended to a few other places, like...
* community name in post body
* user and communities names in feed in both Card and Compact modes
	* Note that we already have Theming > Posts > Show Community Icons for Card View, but I think your implementation (smaller icons, and integrated with the name chip) is even better. :blush:

iparks99:
How do we feel about me baking the avatars into the full name widgets then? That way the avatars show up wherever the usernames or community names appear.

The full name widgets could return a row with two children: an optional (user/community) avatar, and the textspan it currently returns. The full name widgets could then either take in an optional avatar parameter, or just the community and grab all the requisite fields from that object.

Then we could have two global toggleable settings: show user avatars and show community icons.

e.g. both widgets in `full_name_widgets.dart` would return something like:
```dart
return Row(
  mainAxisSize: MainAxisSize.min,
  crossAxisAlignment: CrossAxisAlignment.center,
  children: [
    if (avatar != null)
      Padding(
        padding: const EdgeInsets.only(right: 3.0),
        child: avatar,
      ),
    autoSize
        ? AutoSizeText.rich(
            softWrap: false,
            maxLines: 1,
            overflow: TextOverflow.fade,
            style: textStyle,
            textSpan,
          )
        : Text.rich(
            softWrap: false,
            overflow: TextOverflow.fade,
            style: textStyle,
            textScaler: TextScaler.noScaling,
            textSpan,
          )
  ],
);
```

Thoughts @micahmo?

micahmo:
> How do we feel about me baking the avatars into the full name widgets then?

I think that makes a lot of sense! We just have to make sure either (a) the avatars look good everywhere the full name widget is used, or (b) there's an option to hide it (even when the setting is enabled).

hjiangsu:
> How do we feel about me baking the avatars into the full name widgets then? That way the avatars show up wherever the usernames or community names appear.

I think this is a good idea! Although, I would like to make a slight tweak here - we can place the user avatars within `UserChip` rather than the full name widgets. For reference, UserChip was recently added in https://github.com/thunder-app/thunder/pull/1281

This should alleviate the concern that @micahmo had since the UserChip only shows up in PostView and CommentHeader at the moment!
> the avatars look good everywhere the full name widget is used

iparks99:
Sounds good! I think I'll just close this PR and open a new one.

PR #1305, opened 2024-04-15
Fix issue marking multiple replies as read

micahmo:
## Pull Request Description

This PR fixes an issue with marking multiple replies as read in quick succession, especially on a slow network.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/23746e69-43c8-4b8e-9a14-37408d9aa0aa

### After

https://github.com/thunder-app/thunder/assets/7417301/b496712c-fde4-43a8-91b7-36548499cf58

## Checklist


hjiangsu (approved):
LGTM! I would like to eventually refactor this so that it matches the conventions we have for other parts of the app (e.g., optimistic actions).

micahmo:
Fixed merge conflicts!

> I would like to eventually refactor this so that it matches the conventions we have for other parts of the app (e.g., optimistic actions).

I've been thinking about this as well and wondering if we should do the opposite (show a loading indicator when performing an action until it completes successfully). I guess the problem with optimistic updating is that there's not a great way to tell the user if it failed. For example, if they upvote a post and scroll on, they might never know that it reverted. Anyway, something to think about.

hjiangsu:
> I guess the problem with optimistic updating is that there's not a great way to tell the user if it failed.

You do have a point here - I'd like to eventually add an automatic retry feature (using exponential backoff perhaps) if an action does fail to hopefully reduce these situations. Additionally, we could also add a "system" notification area which shows any errors that occurred with regards to user actions, etc.

The main concern with showing loading indicators is that it makes everything feel "slow" because you have to wait for the network request to complete before updating the status.

PR #1311, opened 2024-04-17
Use loading page for posts/comments

micahmo:
## Pull Request Description

This PR adds the loading page feature for post and comment links. The loading page was originally added in #1200, and it was to handle the case where we have an ambiguous link for a person or community, so we load the full object to verify whether it really exists before navigating. The loading page helps to bridge the gap while we fetch the object so the user knows something is happening.

I've been noticing that navigating to posts and comments has a similar problem where there seems to be no activity for a period of time (especially on a slower network). I believe this is because we similarly load the whole post/comment object before navigating, and unlike users/communities, we *always* load posts/comments, making the problem even more obvious. Furthermore, we often have to do a `ResolveObject` if the entity is on a different instance.

> Review without whitespace.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/8b1ebad3-0107-4c87-980d-7993c31c4fba

### After

https://github.com/thunder-app/thunder/assets/7417301/31990072-40ff-4398-8db8-b00527502971

## Checklist

### DIFF
--- a/lib/shared/pages/loading_page.dart
+++ b/lib/shared/pages/loading_page.dart
@@ -31,7 +31,7 @@ class LoadingPage extends StatelessWidget {
                             semanticLabel: MaterialLocalizations.of(context).backButtonTooltip,
                           )
                         : Icon(Icons.arrow_back_rounded, semanticLabel: MaterialLocalizations.of(context).backButtonTooltip),
-                    onPressed: () => Navigator.of(context).maybePop(),
+                    onPressed: null,
                   )),
               const SliverFillRemaining(
                 child: Center(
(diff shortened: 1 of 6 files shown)


lib/shared/pages/loading_page.dart
```diff
@@ -31,7 +31,7 @@ class LoadingPage extends StatelessWidget {
                             semanticLabel: MaterialLocalizations.of(context).backButtonTooltip,
                           )
                         : Icon(Icons.arrow_back_rounded, semanticLabel: MaterialLocalizations.of(context).backButtonTooltip),
-                    onPressed: () => Navigator.of(context).maybePop(),
+                    onPressed: null,
```
hjiangsu, line 34: Just curious, but why was this changed?

hjiangsu (approved):
LGTM! Just one comment!

ggichure (approved):

lib/shared/pages/loading_page.dart
micahmo: I disabled back navigation from the loading page because it can be confusing. As it was, you could close the loading page, but then the "real" page (which was always still loading in the background) would pop in eventually, which can be confusing. In other words, the ability to close the loading page gave the false sense of canceling the navigation when it really didn't do anything.

Of course, if we could cancel the request upon closing the loading page, that would be a better solution. But that would be a lot more work. I thought this would lead to the least confusion as things are now.

hjiangsu: Ahh okay, that makes sense! Thanks for the clarification

PR #1316, opened 2024-04-18
Adjust language selector styling

hjiangsu:
## Pull Request Description

This PR is a follow up to some of the discussions from #1165. In particular:
- I've moved the LanguageSelector widget to its own shared file (to be used by both post and comment creation pages)
- Adjusted the styling of the language selector to match more closely to Lemmy UI

@micahmo @machinaeZER0 let me know your thoughts on the style change!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|-|-|
|simulator_screenshot_3C2AE542-32D7-40BA-B7E0-B946193FC90B |simulator_screenshot_3729947E-10DC-4453-A27B-2C27C91777CC|
| simulator_screenshot_2D067C5A-2A75-4B5B-AA09-413112C911FC|simulator_screenshot_1B84157D-9E6F-46FE-8EFE-A21F43275BFF|
|simulator_screenshot_B6FA6B3B-741C-42ED-9803-E81C6D3CBA54| simulator_screenshot_EFFFD9BF-3E10-46E9-A35A-D1CAEC3C3748|
|simulator_screenshot_9E8841C9-FCC2-400C-B83F-F63D58B65D03|simulator_screenshot_AD42AF6C-E3FB-4DE9-9162-51593C382FA8|

## Checklist


micahmo (approved):
I think it looks pretty slick! The only feedback I might have is to use a right-facing chevron. On the web, the down arrow implies a literal dropdown, but we're showing a modal. I think we use the right chevron in other places like the account/community selector to imply a modal popup of some sort. Thoughts?

machinaeZER0:
I stylistically prefer the down one, but I agree that since it won't actually be a drop-down in our case that it makes less sense here. Either way I think this is great! 

Side question: If the language name is long enough/the font size is big enough, does it interact weirdly with the NSFW toggle?

micahmo:
One more tiny note: is it possible for the inkwell to only be as wide as the text? Right now it takes up the entire available width.

ggichure (approved):

hjiangsu:
> The only feedback I might have is to use a right-facing chevron.

Done!

> If the language name is long enough/the font size is big enough, does it interact weirdly with the NSFW toggle?

It seems like it wraps around the language:

simulator_screenshot_1393DEE6-363D-40DA-A641-DE7E5831FDD6

> is it possible for the inkwell to only be as wide as the text?

Done!

PR #1319, opened 2024-04-19
Improve the comment context jumping

micahmo:
## Pull Request Description

This PR is a followup to #1252. There I added automatic scrolling to the comments when viewing comment context. Unfortunately it didn't account for slow loading, so unless the comments loaded immediately, it wouldn't work that well. Now I've adjusted the logic to ensure that the scrolling does not occur until the comments are loaded. I also added a little extra room at the bottom so that, in case the comment chain isn't long, it can still be scrolled to the top. Essentially the bottom spacer will now occupy enough room (the height of the list minus the height of the last comment) so that the last comment can always be scrolled to the top.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/thunder-app/thunder/assets/7417301/b4cd16ed-a65a-413f-a06c-a0c694466cbc

### After

https://github.com/thunder-app/thunder/assets/7417301/95331a1d-fea5-491e-9253-ec2cdbfce27f

## Checklist


lib/post/widgets/comment_view.dart
```diff
@@ -96,19 +103,42 @@ class _CommentSubviewState extends State<CommentSubview> with SingleTickerProvid
...
-        Future.delayed(const Duration(milliseconds: 250), () {
-          widget.itemScrollController.scrollTo(index: 1, duration: const Duration(milliseconds: 250));
-        });
+    // The following logic helps us to set the size of the bottom spacer so that the user can scroll the last comment
+    // to the top of the viewport but no further.
+    // This must be run some time after the layout has been rendered so we can measure everything.
+    // It also must be run after there is something to scroll, and the easiest way to do this is to do it in a scroll listener.
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      widget.itemScrollController.primaryScrollController?.addListener(() {
+        double lastCommentTopOffset = widget.itemPositionsListener.itemPositions.value.toList().reversed.skip(2).first.itemLeadingEdge;
```
hjiangsu, line 112: I looked through this and I believe we can remove this `lastCommentTopOffset` logic altogether, and just simply add `_lastCommentKey.currentContext != null)` to the conditional statement below and it should still work?

e.g.,

```dart
if (_bottomSpacerHeight == null && lastCommentTopOffset > 0) {
```

becomes 
```dart
if (_bottomSpacerHeight == null && _lastCommentKey.currentContext != null) {
```

This would remove our dependency on `itemPositionsListener` and allow us to merge this and #1361! Let me know if this works!

micahmo: I seem to recall there was an issue where this was getting calculated too soon, hence the check to ensure that the `lastCommentTopOffset` was a reasonable value (I recall seeing it negative, which was weird). But it seems to work fine the way you suggested, so let's go with it!

hjiangsu: I'll keep a lookout for any inconsistencies here!

hjiangsu (approved):

PR #1323, opened 2024-04-23
Add support for managing uploaded media

micahmo:
## Pull Request Description

This PR adds the ability to delete uploaded images. There is a new section under account settings for managing media. Navigating to this new page will display all uploaded images with the option to delete any of them. This is only supported for instances running >= 0.19.4.

See: thunder-app/lemmy_api_client#21.

Issue: #1447

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/2e49593d-12d1-424e-9c6b-a64aa1ac438d

---

P.S. While working on this feature, I discovered something really interesting! I've always been annoyed at how seemingly quickly our image caches reset despite having the new aggressive image caching setting. It turns out that this is less about image caching and more about widget caching. The default behavior of `SliverList`s is to pretty aggressive dispose and rebuild widgets as they're scrolled off/on screen, and this naturally causes some loading artifacts, even for cached images. However, this trick seems to make things much better.

Pass these paramters to the `SliverList`.

```
addSemanticIndexes: false,
addAutomaticKeepAlives: false,
addRepaintBoundaries: false,
```

And then wrap whatever is the root widget in the list with a `KeepAlive(keepAlive: true)`.

For now I've only done it here, but maybe we could consider doing this in other places of the app (respecting the aggressive image caching option). Here's a before/after of just that difference.

### Before

https://github.com/thunder-app/thunder/assets/7417301/f15ba94e-8a12-4934-a224-f45d80fde45d

### After

https://github.com/thunder-app/thunder/assets/7417301/f686bbc2-a1e0-4976-a4b5-d0b4747661f1


micahmo:
@hjiangsu Would you be able to give me a hand figuring out why the tests are failing here? They're saying something about not finding `freezed`. But I didn't remove that package from the pubspec here, nor in the corresponding change in the API repo. So I'm a bit stuck! 😊

hjiangsu:
Hmm, not entirely sure what's causing the failures. I glanced at the CI logs and did see this: https://github.com/thunder-app/thunder/actions/runs/9664871877/job/26661495345?pr=1323#step:5:170

I'm not sure if this is what's causing the subsequent failures, but it could be something!

image

micahmo:
I saw that too, which is what made me think I removed it by accident, but I didn't. 😆 Maybe I should just add it?

micahmo:
> Maybe I should just add it?

Ok that seems to have worked! It really just changed the reference type in the pubspec.lock from `transitive` to `direct main`.

Now it's failing at the linting step, despite apparently not showing any errors (only info/warning). Lol!

micahmo:
Finally got this building! 😌

micahmo:
@hjiangsu Alrighty, this is my next PR based on age. I know this is a big one in terms of changes, but it's one that I'm very excited to have. I think it's especially useful since it's related to the new (well, not so new any more 😆) Lemmy version, so folks might be expecting this behavior to be available once their instance is upgraded. Let me know what you think!

hjiangsu:
@micahmo Sorry for the delay on this one again - would it be possible for you to fix the merge conflicts here?

After the merge conflicts are fixed, I'll go ahead and merge this one in!

hjiangsu (approved):
Thanks again for implementing this and for fixing the merge conflicts!

micahmo:
Just FYI, this isn't building for me locally and I'm not sure why (so I wasn't going to ask for it to be merged yet). But the CI passed, so I'm not too worried. Does it build for you?

hjiangsu:
Ahh, you know what - I forgot to update the CI to use 3.27, so it builds properly in 3.24 but it doesn't in 3.27. This is likely due to pubspec.lock not being updated properly.

Try running `flutter pub upgrade` and build again! I'll create a PR for this

micahmo:
Ahh yep, that was it! I was able to build and re-test that everything works after the merge. I'll wait for your PR to officially update the `pubspec.lock`. Thanks!!

PR #1324, opened 2024-04-24
Fix empty notifications

micahmo:
## Pull Request Description

This PR fixes an issue with getting empty notifications generated by background checks. If it makes you better, I believe this was a regression that I introduced and was not related to the notifications refactor! :blush: I also snuck in another fix for generating usernames.

> Followup to #1237.

### DIFF
--- a/lib/comment/utils/comment.dart
+++ b/lib/comment/utils/comment.dart
@@ -187,14 +187,24 @@ bool updateModifiedComment(List<CommentViewTree> commentTrees, CommentView comme
 }
 
 String cleanCommentContent(Comment comment) {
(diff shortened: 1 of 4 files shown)


micahmo:
I decided to tack on another fix for duplicate channels in OS settings by assigning the same channel ID for the same channel name. Note that this does not affect message grouping since it is done by group key instead.

micahmo:
I added one more tiny fix related to the payload of the message where we were adding the wrong ID (of the comment instead of the reply). Now I'm working on the account issue, but that will be separate.

ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1325, opened 2024-04-24
Use stronger typing for notifications

micahmo:
## Pull Request Description

This PR continues work on the notifications refactor. This is all in service of supporting multiple accounts better, but it's just a stepping stone. I decided to make our notifications payloads and keys more strongly typed to avoid the weird parsing that we have to do now. That should make things much more reliable. I also decided to group things by both account ID and by inbox message type, so there should be a much clearer delineation of what's what (it also helps to have separate notification channels in the OS settings). I didn't change anything related to APN or UnifiedPush as it seems like there might be some server-side changes required to support the strongly typed payloads. Let me know your thoughts on this.

> Hide whitespace.

### DIFF
--- a/lib/notification/enums/notification_type.dart
+++ b/lib/notification/enums/notification_type.dart
@@ -26,3 +26,10 @@ enum NotificationType {
     }
   }
 }
+
+/// Denotes the different kinds of inbox messages
+enum NotificationInboxType {
+  reply,
+  mention,
+  message,
+}
--- a/lib/notification/notifications.dart
+++ b/lib/notification/notifications.dart
@@ -1,5 +1,6 @@
 // Dart imports
 import 'dart:async';
+import 'dart:convert';
 
 // Flutter imports
 import 'package:flutter/foundation.dart';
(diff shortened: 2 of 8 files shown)


hjiangsu (approved):
LGTM! For UnifiedPush, it sends back a JSON of the notification information, so you can treat it as a response from the Lemmy API essentially.

```
{
  reply: CommentReplyView,
  mention: PersonMentionView,
  message: PrivateMessageView
}
```

I'll handle the changes on the APN side of things 😄

micahmo:
> For UnifiedPush, it sends back a JSON of the notification information, so you can treat it as a response from the Lemmy API essentially.

Ok, very cool! I haven't dug into that side of things yet, but it seems like it bypasses all of the local notifications logic.

> I'll handle the changes on the APN side of things

Thanks! I hope this didn't cause too much extra work.

> Let me know if this PR is good to go!

Should be!

hjiangsu:
> I haven't dug into that side of things yet, but it seems like it bypasses all of the local notifications logic.

Pretty much - the only thing that local and unified push notifications share is the `showAndroidNotification` (which is why I pulled the logic out of the local notifications logic in the refactor!)

PR #1327, opened 2024-04-25
Add selectable post/comment text

micahmo:
## Pull Request Description

Alright, here's take 2 of allowing posts/comments to be selectable. This can replace #1272. Now, when choosing the option to select text, a bottom modal is displayed. There are also some quick actions for viewing raw, selecting all, and copying the selection.

P.S. I removed the checkmark related to the "View source" feature, and simply changed the text to "View original" when in source mode. This feels a little cleaner to me.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/7f1e429e-7877-4e16-a8f0-5af6af46e518

## Checklist

### DIFF
--- a/lib/shared/text/selectable_text_modal.dart
+++ b/lib/shared/text/selectable_text_modal.dart
@@ -23,6 +23,8 @@ void showSelectableTextModal(BuildContext context, {String? title, required Stri
   final FocusNode focusNode = FocusNode();
   final GlobalKey selectableRegionKey = GlobalKey();
 
+  bool isAnythingSelected = false;
+
   showModalBottomSheet(
     context: context,
     showDragHandle: true,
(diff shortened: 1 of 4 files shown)


lib/shared/text/selectable_text_modal.dart
```diff
@@ -0,0 +1,176 @@
+import 'dart:io';
+
+//import 'package:drift/drift.dart';
```
hjiangsu, line 3: Remove commented import!

hjiangsu (approved):
LGTM! Just a few minor tweaks:
- Could we move the text actions so that the copy actions are together? We can have it in this order as an example:
  - Select text
  - Copy comment (renamed from copy text) to indicate that we just want to copy the whole comment. Alternatively, we could just remove this option if you feel like it might be redundant
  - View original
- When we are in the bottom sheet, "Copy selected" should be disabled if the user has not selected any text to copy

lib/shared/text/selectable_text_modal.dart
micahmo: Good catch, fixed!

micahmo:
> * Could we move the text actions so that the copy actions are together? We can have it in this order as an example:
>     * Select text
>     * Copy comment (renamed from copy text) to indicate that we just want to copy the whole comment. Alternatively, we could just remove this option if you feel like it might be redundant
>     * View original

Of course! How does this look?

| image |
| - |

> When we are in the bottom sheet, "Copy selected" should be disabled if the user has not selected any text to copy

Hmm, this is a good idea. I can handle most things through `onSelectionChanged`. However, it seems like tapping elsewhere on the text, which also dismisses the selection, does not trigger this event. I tried a bunch of other stuff, like wrapping everything in a `GestureDetector`, listening to changes from the `FocusNode`, and handling the `onTapText` from `CommonMarkdownBody`, but none of them seem to give me a way to know when the selection was dismissed due to tapping on the text.

At the very least, I could make the button disable initially and only enabled once the first selection is made. Thoughts?

hjiangsu:
> Of course! How does this look?

It looks great. Thanks!

> I could make the button disable initially and only enabled once the first selection is made. Thoughts?

I think that works for now. Thanks for trying this out!

micahmo:
> I think that works for now. Thanks for trying this out!

No problem! Here's what it'll look like.

https://github.com/thunder-app/thunder/assets/7417301/2a8cb674-30ef-4e92-b3da-78cec2d2fc07

PR #1322, opened 2024-04-21
Add score to cross-posts

micahmo:
## Pull Request Description

This PR adds scores to cross-posts. This is something I've been wanting for a while because I often come across a post with cross-posts and I want to view the one with the most engagement. The cross-posts are sorted by votes, but I can't tell if the one I'm looking at or one of the ones listed in the cross-posts area has more votes. Now that is possible to see, and it can help to inform whether I am interested in viewing a certain cross-post.

Note: I'm not sure whether any other Lemmy UI/client does this, but there is precedent from the Mbin UI.

| image |
| - |

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/c266edc7-8df6-4c7f-8eca-6c71cac45da4

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1304, opened 2024-04-15
Add support for customizable action colors

micahmo:
## Pull Request Description

The PR adds the ability to customize the colors for the following actions in Thuder.
* Upvote
* Downvote
* Save
* Mark as read/unread
* Reply/Edit

These colors apply to swipe gestures, inkwells, status indicators, etc. Note that we can pretty easily add more colors in the future. It should also be possible for us to use a similar mechanism to customize other colors in the app (e.g., nested comment indicators or media badge indicators).

Finally, there is a new settings hint in the gestures page which points to this new setting.

As this was quite a big change, please be on the lookout for unusual colors!

## Issue Being Fixed

Issue Number: #1301 and #25

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/59093405-8eec-4647-a481-7c72e18d5010

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1332, opened 2024-04-28
Improved cross-post UI

CTalvio:
## Pull Request Description

Improves cross posting UI

## Issue Being Fixed

Cosmetics and metadata visibility. Provides similar and additional functionality to #1322

## Screenshots / Recordings

Screenshot_20240428_231932
Screenshot_20240428_231924

## Checklist


micahmo (commented):
Hey @CTalvio, we've missed your contributions! 😊

I love the idea of this, even more than what I was trying to add in #1322. I think all of this extra metadata is critical to deciding whether you're interested in viewing a cross-post.

Here are some of my observations.
* All of the inkwells have hard corners (we usually use something like a radius of 5 or 10), although an alternative here would be to make the widget edge-to-edge and then it wouldn't need to be rounded).
* The CrossPosts widget has an inkwell effect even when there's only one cross-post (i.e., it looks tappable, but tapping does nothing).
* The inkwell around the community name seems to have some space at the end.

And then I just had a couple thoughts to potentially increase the usability of the feature.
* What do you think about the default view just being a single statement like "Cross-posted to 1 community"/"Cross-posted to 2 communities" with a down arrow that you have to expand. I think that would alleviate some concerns about the metadata being confused with the main post. You would have to expand to see the metadata, even for a single cross-post.
* What do you think about having the whole row tappable instead of just the community name, so it's more clear that you're navigating to the post and not the community (I think that would be an improvement over the current UX).

Nice work, I think this looks awesome, and I would definitely like to pursue this path over the one I started. 😊

CTalvio:
I glad you approve!

You bring up several thoughts I've had myself since yesterday.

I'll fiddle with this further 👍

CTalvio:
https://github.com/thunder-app/thunder/assets/4365015/cb907853-a01e-4150-a3a6-bead55e73c88

Now looks like this.

CTalvio:
Hang on. Gonna check how it works in post creation.

CTalvio:
Done

micahmo:
Just FYI, I briefly get an error when loading the create post screen. Then it works fine once the results have come in.

https://github.com/thunder-app/thunder/assets/7417301/b62d0b72-4c82-414b-b210-d309dcf2afe9

CTalvio:
Must have been happening with old crosspost info too, then, but thats an easy fix

lib/community/widgets/post_card_metadata.dart
```diff
@@ -353,6 +353,78 @@ class UrlPostCardMetaData extends StatelessWidget {
   }
 }
 
+class CrossPostMetaData extends StatelessWidget {
+  final int unreadComments;
```
hjiangsu, line 357: I would add some additional comments to each parameter for conventions

lib/community/widgets/post_card_metadata.dart
```diff
@@ -353,6 +353,78 @@ class UrlPostCardMetaData extends StatelessWidget {
...
+    super.key,
+    required this.unreadComments,
+    required this.comments,
+    required this.hasBeenEdited,
+    required this.published,
+    required this.saved,
+    required this.score,
+  });
+
+  final MaterialColor upVoteColor = Colors.orange;
```
hjiangsu, line 374: Add a TODO to use the customizable colours that will be introduced in this PR: #1304

lib/community/widgets/post_card_metadata.dart
```diff
@@ -353,6 +353,78 @@ class UrlPostCardMetaData extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return BlocBuilder<ThunderBloc, ThunderState>(
+      builder: (context, state) {
+        return Row(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            IconText(
```
hjiangsu, line 387: Are we able to replace this to use `ScorePostCardMetaData` from `post_card_metadata.dart`? This would help reduce duplicate logic.

If we need to apply a dimmer look to the content, we could possibly set `hasBeenRead` to true when using `ScorePostCardMetaData`. It might not be exactly the same but could be close enough. Alternatively, we could introduce a new `opacity` parameter to `ScorePostCardMetaData` which overrides the existing opacity.

lib/community/widgets/post_card_metadata.dart
```diff
@@ -353,6 +353,78 @@ class UrlPostCardMetaData extends StatelessWidget {
...
+                Icons.arrow_upward,
+                size: 17.0,
+                color: theme.textTheme.titleSmall?.color?.withOpacity(0.4),
+              ),
+              text: formatNumberToK(score),
+              textColor: theme.textTheme.titleSmall?.color?.withOpacity(0.5),
+              padding: 5.0,
+            ),
+            const SizedBox(width: 10.0),
+            IconText(
```
hjiangsu, line 399: Similarly, we can potentially use `CommentCountPostCardMetaData` here

lib/community/widgets/post_card_metadata.dart
```diff
@@ -353,6 +353,78 @@ class UrlPostCardMetaData extends StatelessWidget {
...
+                Icons.chat,
+                size: 17.0,
+                color: theme.textTheme.titleSmall?.color?.withOpacity(0.4),
+              ),
+              text: formatNumberToK(comments),
+              textColor: theme.textTheme.titleSmall?.color?.withOpacity(0.5),
+              padding: 5.0,
+            ),
+            const SizedBox(width: 10.0),
+            IconText(
```
hjiangsu, line 411: This one could possibly use `DateTimePostCardMetaData`!

lib/shared/cross_posts.dart
```diff
@@ -10,6 +13,8 @@ import 'package:thunder/utils/instance.dart';
 import 'package:thunder/post/utils/navigate_create_post.dart';
 import 'package:thunder/post/utils/navigate_post.dart';
 
+import '../community/widgets/post_card_metadata.dart';
```
hjiangsu, line 16: Use absolute imports like the other imports for conventions!

lib/shared/cross_posts.dart
```diff
@@ -47,157 +52,117 @@ class _CrossPostsState extends State<CrossPosts> with SingleTickerProviderStateM
...
-                          itemBuilder: (context, index) {
-                            return Padding(
-                              padding: const EdgeInsets.only(left: 5),
+                    ListView.builder(
+                      shrinkWrap: true,
+                      physics: const NeverScrollableScrollPhysics(),
+                      itemBuilder: (context, index) {
+                        return Column(
+                          children: [
+                            /*const Divider(),*/
```
hjiangsu, line 84: I would remove this commented code to keep the code base clean

hjiangsu (changes requested):
Thanks again for this @CTalvio. I haven't tested this physically but from the demos, it seems to be good to me!

Just have a few comments with regards to the PR itself, but otherwise, looks good to me!

lib/community/widgets/post_card_metadata.dart
CTalvio: This is no longer used

lib/community/widgets/post_card_metadata.dart
CTalvio: Done

lib/community/widgets/post_card_metadata.dart
CTalvio: Good idea

lib/shared/cross_posts.dart
CTalvio: This one was probably automatically added by android studio.

lib/shared/cross_posts.dart
CTalvio: Yep, accidental.

hjiangsu (approved):
LGTM! Thanks for addressing the changes.

PR #1336, opened 2024-04-29
Add some basic documentation on notifications

micahmo:
## Pull Request Description

This PR adds some very basic documentation for notifications. You can see there are several TODOs that we need to flesh out, specifically with regards to self-hosting. But I wanted to get something out there to start. I think eventually we can link to this page from directly in Thunder (or even load up the contents and render it).

BTW, so far, I haven't found a way to request a specific endpoint from the UnifiedPush connector package. Therefore, I'm documenting how to use a custom server with ntfy by setting it up in the app first.


micahmo:
Sounds good!

PR #1338, opened 2024-04-30
Allow creating posts/comments with null language

micahmo:
## Pull Request Description

This PR fixes an issue where creating posts/comments would generate complaints about using an unsupported language. This is because we used to always set the language to `null` (by not passing it) which allowed the server to make whatever determination it wanted (usually using the default language). However, now we fall back to `0` (which I believe is "No language"), which most communities don't support.

This PR fixes that issue by allowing `null` to be passed, meaning "I don't want to specify a language, let the server decide". This fixes the regression.


hjiangsu (approved):
LGTM!

PR #1339, opened 2024-05-01
View notifications as correct user

micahmo:
## Pull Request Description

This PR adds support for switching to the context of a different account when tapping on a notification to an account that is not the currently selected one. This uses a technique similar to the "post as other user" feature, where the account is switched without reloading certain parts of the app, and when navigating back, the account is reset, so that whatever you were looking at previously is left untouched. This also works if you were previously browsing anonymously.

> I also fixed a small issue related to startup where we would skip the notification check if we weren't started by a notification.

> Hide whitespace.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/5c9fd595-19d3-4a42-baf2-1926f84dbfa8

## Checklist


hjiangsu (approved):
LGTM!

PR #1343, opened 2024-05-02
Add notification debugging section

micahmo:
## Pull Request Description

This PR adds a new section to the Debug page that can help us to troubleshoot issues with notifications.

Status
- It will display the currently selected notification mode.
- It will display whether notifications are allowed by the OS.
- It will display the connected UnifiedPush app (if any) and how many are available.
- There is a link to and from the general notification settings.

Testing
* When in local notifications mode, there is an option to send a test notification, both while the app is running, and also to queue up a background notification and kill the app.
* When in UnifiedPush mode, there is an option to request that the server send us a notification while the app is running, and an option to ask the notification server to send us a delayed notification and kill the app. The implementation for these settings is not yet there, as the server doesn't have corresponding endpoints. But these options are also hidden behind an experimental flag for now anyway.

All of these things should be extendable to the iOS side as well!

## Screenshots / Recordings

> Note: The local background notification takes about 2 minutes to arrive, so you may wish to fast-forward to 2:43 at that point. 😊

https://github.com/thunder-app/thunder/assets/7417301/252d9637-80ca-4e6c-82a0-d4e0bdb08cd0

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -7,6 +7,7 @@
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
     <uses-permission android:name="android.permission.VIBRATE" />
+    <uses-permission android:minSdkVersion="34" android:name="android.permission.USE_EXACT_ALARM" />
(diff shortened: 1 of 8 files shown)


hjiangsu (approved):
LGTM! As mentioned in #1335, I'm fast-tracking notification PRs to help with troubleshooting/fixing the issues with notifications. As such, I've only taken a brief look over the changes 😅

PR #1344, opened 2024-05-02
Fix issue with profile as drawer

micahmo:
## Pull Request Description

This PR is a followup to #1313 which fixes an issue where the profile picture may appear where the back arrow should appear instead.


hjiangsu (approved):
LGTM!

PR #1348, opened 2024-05-03
Add ability to create links to settings.

micahmo:
## Pull Request Description

This PR adds a famous feature from Sync where we can provide deep links directly to settings in Thunder. I've been noticing in the Thunder community that people often reference a setting here or there, so I thought it would be super cool to be able to link directly.

* To copy a link to a setting, simply long-press it. It will populate the clipboard with a markdown-compatible version of the link. This will work for any settings which do not already have a long-press action (which is most).
* Then paste wherever you want (Thunder/Lemmy or elsewhere).
* It should work on Android from within Thunder or any other app as well.
* I believe it can be made to work on iOS as well, since it supports the `thunder://` prefix scheme (maybe with minimal or no changes).

Note: As part of these changes, I added the ability for settings widgets to know what setting they represent. That should assist with the highlighting feature, as well as the ability for them to generate the sharing link, and possibly other things in the future.

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/75b53895-e76b-4135-adec-aae5636f43eb


CTalvio:
That is wild

ggichure (approved):

hjiangsu (approved):
First, this is super cool and thanks for doing this! Sorry it took so long to get to this 😅 I just did a very quick review of this, but it LGTM!

PR #1349, opened 2024-05-03
Fix issue popping notification page

micahmo:
## Pull Request Description

This PR fixes an issue where dismissing the last notification from the replies page can pop not only the replies page but also the page underneath it.

### Before

https://github.com/thunder-app/thunder/assets/7417301/e68c3ea7-6b55-493b-a00a-521c45f8d4f0

### After

https://github.com/thunder-app/thunder/assets/7417301/1f92c7d4-c16a-49ac-a1ac-14ac7e5ea45a


hjiangsu (approved):
LGTM!

Issue #1340, opened 2024-05-01
Add an option to disable Swipe Gesture to go back


Bingo90:
### Is Your Feature Request Related to a Problem?

Thunder's Swipe Gesture interferes with Android's system-wide back gesture, making it feel odd to go back. There currently doesn't seem to be a way to disable this.

### Feature Description

Add a toggle to disable the gestures. They're not needed when using Android's gesture navigation.

micahmo:
Hey @Bingo90, thanks for the issue! Just to clarify, do you mean the ability to go back by swiping from anywhere left-to-right on the screen (as opposed to just the edge)?

If so, you're right that there isn't a way to disable that gesture specifically, although you can essentially "override" it by having a swipe gesture enabled for posts and comments. When those are enabled, swiping from left-to-right on the screen activates actions for posts/comments and does not swipe the page back.

Of course, if you don't want any of the action gestures enabled either, then we might have a case for adding a setting to completely disable the full-page swipe.

P.S. Not to discredit your experience at all, but I have never really experienced issues with the Android gesture conflicting with the full-page swipe. If I swipe from the edge, the Android gesture takes precedence. If I swipe from the middle, the app will handle the navigation. If there's something I'm missing, a video demonstrating the issue would go a long way to helping us understand!

Bingo90:
The left to right back gesture, yeah. I noticed it mostly in the settings screen, here's a video:
https://github.com/thunder-app/thunder/assets/28258958/d56142ae-0f62-4016-b479-f56a5b83e415

micahmo:
Hey @Bingo90, apologies but apparently this setting already exists and it totally slipped my mind. Check out Settings > Gestures > Fullscreen Swipe Gestures > disable. If that works for you, we can close this issue!

Bingo90:
> Hey @Bingo90, apologies but apparently this setting already exists and it totally slipped my mind. Check out Settings > Gestures > Fullscreen Swipe Gestures > disable. If that works for you, we can close this issue!

I already had this disabled - I guess this doesn't apply to the settings pages?

micahmo:
> I guess this doesn't apply to the settings pages?

Ah, maybe not. That could be a miss on our part. We'll keep this issue open to track that!

hjiangsu:
Just chiming in on this - I believe that setting only disables the full-screen swipe gesture, but keeps the edge swipe to go back gesture which is why it's not working.

[...]

PR #1350, opened 2024-05-03
Disable Thunder's edge swipe to go back on Android

hjiangsu:
## Pull Request Description

This PR disables Thunder's edge swipe to go back gesture when `Full Swipe Gestures` is disabled. This does not affect Android's system navigation gestures. For iOS, the edge swipe to go back will always be enabled regardless of the setting toggle because that's the primary way to navigate back on iOS devices.

@micahmo if you could test this out on a physical device and let me know, that would be great! I tested it on an emulator and it seems to be working as expected there.

## Issue Being Fixed

Issue Number: #1340

## Checklist

### DIFF
--- a/lib/account/widgets/profile_modal_body.dart
+++ b/lib/account/widgets/profile_modal_body.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:dart_ping/dart_ping.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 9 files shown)


micahmo (approved):
Tested on a physical device and it seems fine!

Although I'm still a bit unsure as to the cause of this bug because the page showed in the bug demo wasn't created with SwipeablePageRoute (and I verified that I can't use full-page swipe on my device, regardless of the setting).

hjiangsu:
> I'm still a bit unsure as to the cause of this bug because the page showed in the bug demo wasn't created with SwipeablePageRoute

Ahh yeah you're right - I wasn't able to reproduce that issue either. I think this is still a step in the right direction!

PR #1352, opened 2024-05-04
Allow commenting as other user

micahmo:
## Pull Request Description

The PR adds the ability to create comments as a different user than the currently selected one. It should work whether responding directly to a post or to a comment. As with #1159, there is object resolution in place to look up the corresponding post and (if necessary) comment on the new instance, and there is error handling if those objects are not found. The original user is restored upon navigating back while the rest of the app is left untouched. Since we can't know if the new comment has federated yet, we can't do the nifty thing where the new comment is immediately shown in the thread, but by refreshing it can be seen once it federates.

> Hide whitespace.

## Screenshots / Recordings

> Apologies in the demo that the comments take a little while to federate.

https://github.com/thunder-app/thunder/assets/7417301/d82db6dc-a9ee-4305-935a-241712b7c102


lib/user/widgets/user_selector.dart
```diff
@@ -71,6 +87,44 @@ class _UserSelectorState extends State<UserSelector> {
...
               }
+
+              // If there is a selected post, see if we can resolve it to the new user's instance.
+              if (widget.postActorId?.isNotEmpty == true && widget.onPostChanged != null) {
+                PostView? resolvedPost;
+                try {
+                  final ResolveObjectResponse resolveObjectResponse = await LemmyApiV3(newUser.instance!).run(ResolveObject(q: widget.postActorId!));
+                  resolvedPost = resolveObjectResponse.post;
+                  if (resolvedPost != null) {
+                    widget.onPostChanged!(resolvedPost);
```
hjiangsu, line 98: Return early here so that we don't have to go through the rest of the logic if it succeeds!

hjiangsu (approved):
This is super cool! Thanks for adding this feature in. I haven't tested it yet but the code LGTM!

Just one minor nitpick and it should be good to go.

lib/user/widgets/user_selector.dart
micahmo: I'm not sure I completely understand what you want me to change here. Do you mean, if we're able to successfully resolve the post, for example, to skip the rest of the object resolution?

I guess that could work in this context, but maybe `UserSelector` could be used in a different context where we'd want multiple objects resolved, so we can't know (internally) that we're done after one succeeds. What do you think?

hjiangsu: Oops, I forgot to respond back to this! I was under the impression that we only had to resolve a single object here, but you do make a good point that we might need to resolve multiple objects in the future!

I'll resolve this issue for now.

PR #1355, opened 2024-05-07
A few notifications-related improvements

micahmo:
## Pull Request Description

This PR includes a couple of miscellaneous changes related to notifications. (Don't feel the need to fast-track this as it shouldn't be related to any bugs.)
* Use the LoadingPage while setting up the notifications page after responding to a tap (since it can take a while to load everything).
* Ensure that the optional profile picture on the main feed doesn't reload for temporary accounts by respecting the `reload` flag.
* Enlarge the thunder icon displayed in the notification bar to match other apps' sizes.

### Icon Size (look at the top)

| Before | After |
|--------|--------|
| image | image |


hjiangsu (approved):
LGTM! I'm not too sure if it affects things, but have you tested this change out with a profile/release build?

The main thing to look out for is if the icon appears properly on those modes since we adjusted the icon location (the build processes might not include the new icon assets in release)

micahmo:
I just tested both debug and profile as well as the production and development flavors, and everything looks good!

(I can't make a release build due to a signing error and I never looked into that. Probably I'd have to supply my own signing key. But I've been able to get away with profile. 😊)

hjiangsu:
If it works on profile mode, then I believe it should work on release builds as well!

PR #1357, opened 2024-05-07
Fix fab transparency and exit fade

CTalvio:
## Pull Request Description

Removes redundant ModalBarrier and fixes the FAB exit animation.

Both ModalBarrier would be present in the main feed, stacking their colors. Using the ModalBarrier under AnimatedOpacity also breaks the FAB exit animation, so i separated them out in the stack and used a Container under an IgnorePointer to accomplish the darkening.

## Issue Being Fixed

Solid color FAB background only in main feed, instant exit fade as the instantly removed ModalBarrier cannot animate.

## Checklist


hjiangsu (approved):
LGTM! I haven't ran it physically yet but from what you mentioned on Matrix, seems to address the initial concerns

PR #1359, opened 2024-05-08
Improve creating posts/comments

micahmo:
## Pull Request Description

This PR includes a few changes related to creating posts/comments.
* Improve selectability of posts/comments when replying.
	* I found that the `MarkdownWidget`'s `selectable` flag doesn't support selecting across multiple paragraphs, so I wrapped it in a `SelectableRegion` which is more flexible.
	* I added the ability to view the raw contents of a post/comment when replying. This gives you the ability to quote the original poster while preserving their markdown formatting within your quote.
		* The next step in my plan is to automatically populate the post/comment selection into the comment editor when pressing the "quote" markdown option. This will make it even easier to quote something (without copying/pasting) and to optionally preserve the markdown formatting.
* Made the new post preview look more like the new comment preview.

> Hide whitespace

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/e2a96c05-33a0-42e9-8a13-275019789318


ggichure (approved):

hjiangsu:
> I found that the MarkdownWidget's selectable flag doesn't support selecting across multiple paragraphs, so I wrapped it in a SelectableRegion which is more flexible.

This is an aside, but we might be able to completely remove the `selectable` parameter in MarkdownWidget if that's the case! I believe we initially used it in the old comment page only (but check to see if there are any existing references to it)

I'm also wondering if we can adjust the positioning of the view source button! From the demo, it feels like it's lumped up with the metadata information which might reduce discovery of the feature.

Some ideas:
- We can play around with padding to make it feel like it's not part of the metadata
- Make a new section under the preview for additional actions (we could even potentially add another action to just copy the whole contents)

These are just some initial points. I haven't taken a look at the implementation yet but will do so soon! 😄

micahmo:
> This is an aside, but we might be able to completely remove the `selectable` parameter in MarkdownWidget if that's the case! I believe we initially used it in the old comment page only (but check to see if there are any existing references to it)

Good idea! I've removed it from `CommonMarkdownBody`, but not from `ExtendedMarkdownBody` as that maps to `ExtendedMarkdownWidget` which maps to `MarkdownWidget` which comes from a package, and I figured there's no need to hide legitimate parameters via our wrapper widgets.

> I'm also wondering if we can adjust the positioning of the view source button!

Sure, I'll play around with it!

hjiangsu:
> Alright, I added a divider, and also a copy option

I think that looks better! Thanks for making the change 😄

lib/shared/divider.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:flutter/material.dart';
+import 'package:thunder/shared/conditional_parent_widget.dart';
+
+class ThunderDivider extends StatelessWidget {
+  final bool sliver;
```
hjiangsu, line 5: I would add some comments here to indicate the parameters and what they mean! (especially `sliver` since that might not be obvious)

lib/comment/view/create_comment_page.dart
```diff
@@ -93,6 +94,10 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
   /// Used for restoring and saving drafts
   SharedPreferences? sharedPreferences;
 
+  /// Whether to view source for posts or comments
+  bool postViewSource = false;
```
hjiangsu, line 98: I believe we can only show either post or comment at the same time (they're mutually exclusive since you can only reply to a post or comment at one time). We could get away with just using a single parameter `viewSource` here!

lib/post/widgets/post_view.dart
```diff
@@ -228,6 +255,49 @@ class _PostSubviewState extends State<PostSubview> with SingleTickerProviderStat
                     hasBeenEdited: postViewMedia.postView.post.updated != null ? true : false,
                     url: postViewMedia.media.firstOrNull != null ? postViewMedia.media.first.originalUrl : null,
                   ),
+                  if (widget.showReplyEditorButtons && widget.postViewMedia.postView.post.body?.isNotEmpty == true) ...[
```
hjiangsu, line 258: It seems like this logic and the comment reply action logic is the same - we can move this to its own widget

lib/post/widgets/post_view.dart
```diff
@@ -173,16 +184,32 @@ class _PostSubviewState extends State<PostSubview> with SingleTickerProviderStat
...
-                      ? ScalableText(
-                          post.body ?? '',
-                          style: theme.textTheme.bodySmall?.copyWith(fontFamily: 'monospace'),
-                          fontScale: thunderState.contentFontSizeScale,
-                        )
-                      : CommonMarkdownBody(
-                          body: post.body ?? '',
-                          isSelectableText: widget.selectable,
-                        ),
+                  child: ConditionalParentWidget(
```
hjiangsu, line 187: Do you think we need to apply the `SelectableRegion` here, or would it be more suited as part of the create comment page? (same thing with the comment SelectableRegion)

I guess the main discussion point is whether or not we're expecting to make the post selectable in other areas outside of the create comment page. We already have a way to copy post/comments from #1327!

lib/community/pages/create_post_page.dart
```diff
@@ -515,26 +516,31 @@ class _CreatePostPageState extends State<CreatePostPage> {
...
-                                : MarkdownTextInputField(
-                                    controller: _bodyTextController,
-                                    focusNode: _bodyFocusNode,
-                                    label: l10n.postBody,
-                                    minLines: 8,
-                                    maxLines: null,
-                                    textStyle: theme.textTheme.bodyLarge,
+                            AnimatedCrossFade(
+                              firstChild: Padding(
+                                padding: const EdgeInsets.only(top: 8.0),
```
hjiangsu, line 521: We might be able to get rid of this `padding` and apply it to the `Container`'s margin instead!

hjiangsu (commented):
Just did a quick review of this and had some points of discussion!

lib/shared/divider.dart
micahmo: Done!

lib/comment/view/create_comment_page.dart
micahmo: Good point! Done!

lib/post/widgets/post_view.dart
micahmo: I originally put the `SelectableRegion` around the whole `PostSubview` and `CommentContent` widgets, but that had a couple of weird side-effects. For one, I couldn't see the drag handles or the selection highlight color. But the bigger issue is that it allows everything in those widgets to be selectable, including the metadata, which I thought was weird! It seemed to be behave much better when wrapped directly around the `CommonMarkdownBody`, and I don't think it causes any harm, even if this is the only place we use it. Thoughts?

lib/community/pages/create_post_page.dart
micahmo: Good idea, done!

lib/post/widgets/post_view.dart
micahmo: Sure! I left the divider there, since it feels separate to me, but I moved the row with the buttons.

micahmo:
This should be ready to go, depending on your thoughts on the unresolved comment!

lib/post/widgets/post_view.dart
hjiangsu: Ahh okay, you make a good point about allowing metadata to be selected! We'll keep it as it!

Issue #588, opened 2023-08-08
There is a weird quick in comment navigation


micahmo:
**Description**  
There is one scenario in which the new comment navigation feature doesn't quite work properly.

**How to Reproduce**  
1. Use comment navigation to navigate to a comment.
2. Collapse the comment.
3. Without scrolling, navigate to the next comment.

It will jump to where the next comment _used_ to be before collapsing.

**Expected Behavior**  

It should jump correctly to the next comment.

**Screenshots**  

https://github.com/thunder-app/thunder/assets/7417301/f0429098-7a9d-42d2-92e7-5952e18a47b8

**Device & App Version:**  
N/A

**Additional Context**  
N/A

PR #1361, opened 2024-05-09
Replace scrollable_positioned_list with super_sliver_list

hjiangsu:
## Pull Request Description

For some background, I'm in the process of refactoring the comment list to align more with our current conventions of using Slivers for performance reasons. However, I encountered one issue where `scrollable_positioned_list` does not have a Sliver version.

Because of this, I started to look into other packages and stumbled upon `super_sliver_list`. This package allows us to scroll to a given index, and from my testing, seems to work fairly well. Additionally, it exposes a few more additional functions that could be useful (including the ability to specify the alignment when scrolling to a given index)

This PR essentially replaces the usage of `scrollable_positioned_list` with `super_sliver_list`. Additionally, I've adjusted the logic for the comment navigation FAB. It now keeps track of the index internally which hopefully simplifies some logic.

@micahmo I'm not sure if the changes made to the comment navigation FAB breaks anything, but do let me know (since you were mainly the one who implemented the scroll-to-index functionality)!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
This looks great! I had stumbled upon `super_sliver_list` too and it totally makes sense as the next logical step to use the slivers for the comments.

In addition, I think keeping track of the index internally rather than through a bloc event is a huge upgrade, because there were a number of times that the state wouldn't be right after jumping (because it wouldn't emit everything correctly).

The best part of this change, which might not have even been intentional, is that it seems to fix #588!!

> Note that this PR will have some conflicts with #1319.

hjiangsu:
> In addition, I think keeping track of the index internally rather than through a bloc event is a huge upgrade, because there were a number of times that the state wouldn't be right after jumping (because it wouldn't emit everything correctly).

There are some situations where we might need to override the index (such as when we jump to a specific comment, or when performing search) which is why I kept `navigateCommentIndex` in post bloc. However, for everything else, I do agree that storing it locally simplifies a lot of the logic!

> Note that this PR will have some conflicts with https://github.com/thunder-app/thunder/pull/1319.

I'll see what I can do here to mimic the behaviour from #1319, since we don't have access to `itemScrollController` anymore!

micahmo:
> I'll see what I can do here to mimic the behaviour from #1319, since we don't have access to `itemScrollController` anymore!

Thanks!! The main things are to delay the scrolling until the comments are built, and to calculate the height of the bottom spacer based on the height of the last comment so it can always be scrolled to the very top even if short.

hjiangsu:
I'll go ahead and merge this in now that #1319 is merged in. Let me know if you encounter any issues running with this change!

hjiangsu:
Interesting, I didn't know that behaviour existed! There might be a way to do this, but I would have to look further into how `super_sliver_list` works. Thanks for finding this 😅

micahmo:
Thanks!! I'm pretty sure it's how most people will expect the jump feature to work. Jump down or up to the next top-level comment. (It occurred to me that the same problem also occurs when jumping upward. It needs to be relative to where we've scrolled, not absolute.)

PR #1366, opened 2024-05-15
Fix partial comment navigation

hjiangsu:
## Pull Request Description

This PR adds some additional logic that attempts to take into account the user's currently visible comments, and perform comment navigation based on the visible contents. I also increased the animation duration since it felt a bit too quick in the previous implementation!

When the user has partially navigated down, tapping on the next comment will bring them to the next comment. Likewise, tapping on the previous comment will bring the user up to the current comment (if it is partially obscured). If the current comment is not partially obscured, it will bring the user to the previous comment.

Navigating down:

https://github.com/thunder-app/thunder/assets/30667958/59560ab5-d8bd-4455-a545-0d67a0fcf356

Navigating up:

https://github.com/thunder-app/thunder/assets/30667958/2548ba30-4830-467b-8610-40ce27eb3d53

Note: There is currently an issue where if you tap the up comment nav while you're in a partially obscure comment, and repeat the action with the same comment, it goes up twice. Here's a video showing the issue:

https://github.com/thunder-app/thunder/assets/30667958/cb01eee3-8e62-4412-b4f3-1ff655324575

I haven't found an easy way to fix this yet, so if you have any ideas, please let me know!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/shared/comment_navigator_fab.dart
+++ b/lib/shared/comment_navigator_fab.dart
@@ -114,34 +114,44 @@ class _CommentNavigatorFabState extends State<CommentNavigatorFab> {
   }
 
   void navigateUp() {
-    if (currentIndex == 0) return;
+    var unobstructedVisibleRange = widget.listController.unobstructedVisibleRange;
+
+    int nextIndex = currentIndex - 1;
+    if (unobstructedVisibleRange?.$1 != null && unobstructedVisibleRange!.$1 != currentIndex) {
+      nextIndex = unobstructedVisibleRange.$1;
+    } else if (currentIndex != 0) {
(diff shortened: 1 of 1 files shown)


micahmo (approved):
Nice find, thank you so much for fixing this!!

hjiangsu:
>Note: There is currently an issue where if you tap the up comment nav while you're in a partially obscure comment, and repeat the action with the same comment, it goes up twice.

I'll go ahead and merge this in first, and work on this issue at a later time (I don't believe it's a big deal because this situation shouldn't happen too often, and can be fixed by tapping on the next comment button)

PR #1081, opened 2024-01-24
Experimental support for video player

ggichure:
## Pull Request Description

 -  [youtube_player_flutter](https://pub.dev/packages/youtube_player_flutter) to play  yt videos on android and ios
 -  [youtube_player_iframe](https://pub.dev/packages/youtube_player_iframe) to play  videos on web , etc
 -  [river_player](https://pub.dev/packages/river_player) to play any file with a standard codec `mp4,mov` etc 
 

## Issue Being Fixed

Issue Number: #383

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/31619962/8e5ecf34-72bb-44ee-b31d-626bbe36310e

## Checklist

- [x] Did you update CHANGELOG.md?
- [x] Did you use localized strings where applicable?
- [x] Did you add `semanticLabel`s where applicable for accessibility?

Issue #1009, opened 2023-12-26
Embedded Peertube videos


K4LCIFER:
**Is your feature request related to a problem? Please describe.**

Currently, Peertube videos are displayed as a link ([for example](https://sh.itjust.works/post/11657970)).

**Describe the solution you'd like**

It would be great if Peertube videos were embedded directly into Thunder.

**Describe alternatives you've considered**
None.

**Additional context**
- Not sure if this is an exact duplicate of #383 
- @Nutomic [mentioned] that embeds are supported by the Lemmy backend, so this should be able to be implemented into Thunder.

Nutomic:
FYI the embed url is automatically put into `post.embed_video_url` for Peertube/Youtube links.

PR #1369, opened 2024-05-16
Image and video player improvements

hjiangsu:
## Pull Request Description

> Review without whitespace

This PR builds upon the great video player work that @ggichure implemented in #1081 and also adds some simplification to the way we determine image dimensions.

To summarize, these are the high level changes:
- Added back buttons to the video players (mainly to allow iOS to navigate back to the previous page).
- Added a snackbar which appears if a video fails to load. Tapping on the snackbar will open the video link in a browser as a fallback.
- Fixed an issue where some youtube links were not being detected properly (and thus, attempted to use the normal video player)
- Adds a separate widget that displays in the feed if it detects that the url is a video. This also allows any existing thumbnail images to show up on the feed along with a play button that opens up the video player
- Added logic in the `parsePostView` function to handle embedded video links `postView.post.embedVideoUrl`. In these cases, the `postView.post.url` may not directly point to a video link, but `postView.post.embedVideoUrl` may. We want these links to also open up the video player (e.g., embedded peertube video links) 
  - For reference: https://github.com/thunder-app/thunder/issues/1009
- This is not directly related to videos, but I've improved the handling of determining image dimensions. It turns out there is an existing Flutter implementation that helps decode images, so fetching image dimensions should be more robust. This should reduce instances where images fail to load properly, or have the incorrect image dimensions.

I've also adjusted `Media` to have a `thumbnailUrl` parameter. This is to allow us to account for both images/videos
- `mediaUrl` should now point to the url of the media (image or video urls)
- `thumbnailUrl` should point to a valid image url for displaying on the feed
- `originalUrl` is the original url of the post

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/devtools_options.yaml
+++ b/devtools_options.yaml
@@ -1,0 +1,1 @@
+extensions:
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1,6 +1,10 @@
 PODS:
   - background_fetch (1.3.2):
     - Flutter
+  - Cache (6.0.0)
+  - connectivity_plus (0.0.1):
+    - Flutter
+    - FlutterMacOS
   - device_info_plus (0.0.1):
     - Flutter
   - Flutter (1.0.0)
@@ -10,6 +14,13 @@ PODS:
     - Flutter
   - flutter_icmp_ping (0.0.1):
     - Flutter
+  - flutter_inappwebview (0.0.1):
+    - Flutter
+    - flutter_inappwebview/Core (= 0.0.1)
+    - OrderedSet (~> 5.0)
+  - flutter_inappwebview/Core (0.0.1):
+    - Flutter
+    - OrderedSet (~> 5.0)
   - flutter_keyboard_visibility (0.0.1):
     - Flutter
   - flutter_local_notifications (0.0.1):
@@ -21,8 +32,15 @@ PODS:
   - gal (1.0.0):
     - Flutter
     - FlutterMacOS
+  - GCDWebServer (3.5.4):
+    - GCDWebServer/Core (= 3.5.4)
+  - GCDWebServer/Core (3.5.4)
+  - HLSCachingReverseProxyServer (0.1.0):
+    - GCDWebServer (~> 3.5)
+    - PINCache (>= 3.0.1-beta.3)
   - image_picker_ios (0.0.1):
     - Flutter
+  - OrderedSet (5.0.0)
   - package_info_plus (0.4.5):
     - Flutter
   - path_provider_foundation (0.0.1):
@@ -30,10 +48,24 @@ PODS:
     - FlutterMacOS
   - permission_handler_apple (9.3.0):
     - Flutter
+  - PINCache (3.0.4):
+    - PINCache/Arc-exception-safe (= 3.0.4)
+    - PINCache/Core (= 3.0.4)
+  - PINCache/Arc-exception-safe (3.0.4):
+    - PINCache/Core
+  - PINCache/Core (3.0.4):
+    - PINOperation (~> 1.2.3)
+  - PINOperation (1.2.3)
   - pointer_interceptor_ios (0.0.1):
     - Flutter
   - push_ios (0.0.1):
     - Flutter
+  - river_player (0.0.1):
+    - Cache (~> 6.0.0)
+    - Flutter
+    - GCDWebServer
+    - HLSCachingReverseProxyServer
+    - PINCache
   - share_plus (0.0.1):
     - Flutter
   - shared_preferences_foundation (0.0.1):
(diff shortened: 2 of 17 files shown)


hjiangsu:
> Do we need the play icon on top of the thumbnail when we already have the media type indicator?

I can get rid of it! It just felt weird to not have anything in the box when there's no thumbnail

> Any idea why videos without thumbnails sometimes look different?

Hmm, could you provide the community/links so that I can check this out?

> I noticed that sharing a post with a video doesn't provide the option to share the video link, only the thumbnail link.

I think we can make this into a separate PR!

micahmo:
> > Any idea why videos without thumbnails sometimes look different?
> 
> Hmm, could you provide the community/links so that I can check this out?

Here's an example I found in youtube@lemmy.fmhy.ml (I know that instance is down, but the community is available via other instances).

| image |
| - |

Another thing I see in this screenshot (and that I've been noticing a lot while browsing lately) is that URLs detected as non-video links with no thumbnail have the play icon instead of the web browser icon as the placeholder. I think that's a regression (maybe not in this PR though).

hjiangsu:
Thanks, I'll check it out.

> URLs detected as non-video links with no thumbnail have the play icon instead of the web browser icon as the placeholder

I noticed that too - it does seem to be a regression from the original PR (it'll be fixed in this PR though!)

micahmo:
Interesting, good catch!! I think I've seen that error placeholder for articles as well. Do we just need to supply our own widget for error cases (which can be the solid gray color, as though there's image at all)?

hjiangsu:
A few updates:
- I removed the extra play icon that appears when there's no thumbnail, it should be the default solid colour background
- I added a way to open in browser when opening the video, in case the video does not load properly (for both youtube and the regular video player)
- Fixed the issue where links would show the play icon
- Played around with the UI for both compact/card mode

I think it's now at a place where you can try out the new changes and let me know if you see any new issues!

ggichure:
The youtube player doesn't have an unmute which means should you want to enable audio you have to go to settings to disable auto mute.

ggichure (approved):

micahmo:
> The youtube player doesn't have an unmute which means should you want to enable audio you have to go to settings to disable auto mute.

I noticed this as well, and I like your suggestion!

P.S. One more thing I noticed on Android is that, once you enter fullscreen mode, Thunder stays in fullscreen mode even after exiting the video player. This means that the navigation and notification bars are hidden and have to be "swiped in". @ggichure Any ideas on that?

hjiangsu:
Thanks for the feedback! I'll add in the mute button for the youtube player. It would be nice if we could combine both players into one to make everything more consistent but that can be explored in the future.

> One more thing I noticed on Android is that, once you enter fullscreen mode, Thunder stays in fullscreen mode even after exiting the video player. 

This also happens on iOS and I'm not too sure why - it's probably related to the internal logic of the video players.

ggichure:
> One more thing I noticed on Android is that, once you enter fullscreen mode, Thunder stays in fullscreen mode even after exiting the video player.

We can play around with this to achieve desired result.

For BetterPlayer , although this seems to be working correctly without this.

```dart
 List<SystemUiOverlay> systemOverlaysAfterFullScreen = SystemUiOverlay.values,
  List<DeviceOrientation> deviceOrientationsAfterFullScreen = const [DeviceOrientation.portraitUp, DeviceOrientation.portraitDown, DeviceOrientation.landscapeLeft, DeviceOrientation.landscapeRight],
```

```Dart
 BetterPlayerConfiguration betterPlayerConfiguration = BetterPlayerConfiguration(
      aspectRatio: 16 / 10,
      fit: BoxFit.cover,
      autoPlay: autoPlayVideo(thunderBloc),
      fullScreenByDefault: thunderBloc.videoAutoFullscreen,
      looping: thunderBloc.videoAutoLoop,
      autoDetectFullscreenAspectRatio: true,
      useRootNavigator: true,
      systemOverlaysAfterFullScreen: [ //New
        SystemUiOverlay.bottom, //New
        SystemUiOverlay.top, //New
      ],
```

Youtube Player

```Dart
child: ypf.YoutubePlayerBuilder(
                  onExitFullScreen: () {
                    SystemChrome.setEnabledSystemUIMode(SystemUiMode.edgeToEdge);
                  },
```


https://github.com/thunder-app/thunder/assets/31619962/f7a5b016-b5b4-4258-95fa-3ec86638d193

micahmo:
I think the best way to review this is to test it. 😊 Since @ggichure approved, you can merge any time, but I will also build this and run it for a little while if that's ok. 😊

micahmo:
One super small thing I just noticed is that playing a video doesn't respect the "Mark Read After Viewing Media" setting. Otherwise so far so good...

ggichure:
> One super small thing I just noticed is that playing a video doesn't respect the "Mark Read After Viewing Media" setting. Otherwise so far so good...

Working on this on a different branch

PR #1374, opened 2024-05-21
Allow quoting selected text from post/comment reply

micahmo:
## Pull Request Description

This PR allows selecting and quoting text from the post/comment that you are replying to. This is especially useful when the original poster has used some markdown that you want to preserve when quoting them. You can view the raw text, then select, then press the quote button, and their text will be used as the basis of your quote.

See: https://github.com/thunder-app/markdown-editor/pull/10

https://github.com/thunder-app/markdown-editor/assets/7417301/84145d2a-81d0-439d-8c07-77e9ae4c7020


pubspec.yaml
```diff
@@ -22,8 +22,8 @@ dependencies:
   path: ^1.8.3
   markdown_editor:
     git:
-      url: https://github.com/thunder-app/markdown-editor.git
-      ref: 3196471d8e1b26873aecaf04254043cf9817bbe3
+      url: https://github.com/micahmo/markdown-editor.git
```
micahmo, line 25: Must fix this before merging

micahmo: Done

hjiangsu (approved):
LGTM!

PR #1376, opened 2024-05-22
Add user/community display name setting

micahmo:
## Pull Request Description

Previously, the option to show a use "display names" was only available for user names within the feed. This PR makes that setting part of the global name formatting options, so it will apply anywhere the full name widgets are used (with some intentional exceptions). I've also added a similar option for communities.

Note that this setting is ignored in places where we already show the display name (e.g., profile pages, autocomplete results, etc.).

https://github.com/thunder-app/thunder/assets/7417301/e0481955-ab3a-418a-a99f-86ce00aa0222


lib/shared/chips/community_chip.dart
```diff
@@ -15,10 +15,11 @@ import 'package:thunder/utils/instance.dart';
 class CommunityChip extends StatelessWidget {
   const CommunityChip({
     super.key,
-    this.communityId,
-    this.communityAvatar,
-    this.communityName,
-    this.communityUrl,
+    required this.communityId,
```
hjiangsu, line 18: If we're marking these as required, then we should make the parameter's non-nullable below!

hjiangsu (approved):
Thanks for doing this! Just have one minor comment otherwise LGTM!

lib/shared/chips/community_chip.dart
micahmo: Sure thing, just changed! (I wouldn't say `required` fields should _always_ be non-nullable; sometimes it's valid to pass `null`, but in this case everything works, so we're good.)

PR #1388, opened 2024-05-28
Fix account page infinite loading

hjiangsu:
## Pull Request Description

> Hide whitespace when reviewing

This PR attempts to fix an issue that I noticed where the account page would load indefinitely until the refresh button was manually triggered. I'm not too sure what the root cause of the issue is, but wrapping the widget with `BlocBuilder` seems to resolve the issue. It's difficult to accurately reproduce the issue.

This is most likely a regression from the recent account page changes that were made.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/account/pages/account_page.dart
+++ b/lib/account/pages/account_page.dart
@@ -41,14 +41,18 @@ class _AccountPageState extends State<AccountPage> with AutomaticKeepAliveClient
           },
         ),
       ],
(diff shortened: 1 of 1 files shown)


micahmo (approved):
I think I've seen this too. We'll have to keep our eyes out for weird quirks like this, especially when doing temporary account switches. LGTM

PR #1389, opened 2024-05-28
Defer loading image dimensions

micahmo:
## Pull Request Description

This is a small PR which defers the loading of image dimensions in the image viewer. Previously we would wait for the dimensions to load and the image. Since first call is not cached, this combination can make images take longer than desired to load, especially on a slow network.

It seems to work fine to change `maxZoomLevel` after the image has initially loaded.

> Hide whitespace.


hjiangsu (approved):
LGTM!

PR #1390, opened 2024-05-28
A few comment/image improvements

micahmo:
## Pull Request Description

This PR makes a couple of changes to comments and images.
1. Comment contents are now in an `AnimatedSize`. This means when anything changes with respect to the height of the comments (an example is viewing the raw markdown), it will more nicely animate.
2. Add a solid color placeholder for images previews in markdown. This should make it more obvious that an image is loading instead of there just being a blank space in comments.
3. Defer SVG checking, which previously would cause a big delay (and a big jump in comment size, which is annoying when scrolling) before the image placeholder could take its place. Now we assume it's a regular image and only show an SVG if we find otherwise.
4. Remove the "unable to load" message (especially since this looked bad in the feed) and replace with a more generic placeholder for a "failed" image. (This also makes the SVG loading look less bad than an error message.)

Hopefully that all makes sense!

> Hide whitespace.

### Bad Image Placeholder

| Before | After |
|--------|--------|
| image | image | 

#### Before

https://github.com/thunder-app/thunder/assets/7417301/6116d7f7-6ea6-4f85-ad61-689a7b49143c

#### After

https://github.com/thunder-app/thunder/assets/7417301/835610d9-40a9-4482-9968-42e01f389e0e

### DIFF
--- a/lib/shared/image_preview.dart
+++ b/lib/shared/image_preview.dart
@@ -8,6 +8,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:thunder/core/enums/image_caching_mode.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
+import 'package:thunder/utils/colors.dart';
 
 import 'package:thunder/utils/media/image.dart';
 
@@ -25,6 +26,7 @@ class ImagePreview extends StatefulWidget {
   final void Function()? navigateToPost;
   final bool? isComment;
   final bool? read;
+  final bool showPlaceholder;
 
   const ImagePreview({
     super.key,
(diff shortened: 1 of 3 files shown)


lib/shared/image_preview.dart
```diff
@@ -25,6 +26,7 @@ class ImagePreview extends StatefulWidget {
   final void Function()? navigateToPost;
   final bool? isComment;
   final bool? read;
+  final bool showPlaceholder;
```
hjiangsu, line 29: I think we should always show a placeholder for images - we might not need to add this as an optional parameter. Thoughts?

hjiangsu (approved):
Thanks for doing this - the comment section jumping was also somewhat annoying for me, but I never had the chance to look into it further!

LGTM - just one comment!

lib/shared/image_preview.dart
micahmo: Sure! I wasn't sure if it would have a negative impact on any other places that `ImagePreview` is used, but I agree always having a placeholder makes sense.

PR #1363, opened 2024-05-11
Refactor Post/Comment page

hjiangsu:
## Pull Request Description

This is a draft PR which refactors the existing post page (and comments). This is still very early in development and does not have all the features that the current post page has. Since this is a large re-architecture/refactor, I've opted to do the following:
- I introduced a new toggle in the Debug settings page to enable experimental mode. Thunder will need to be restarted in order to use the experimental features.
- When experimental features are enabled, navigating to the post page will use the new post page. Otherwise, it will continue to use to current post page.
- I would like to slowly bring the refactor up to parity with the current post page. I'm thinking we can do this with a series of PRs.
  - The benefit with hiding this under an experimental toggle is that we can merge the post page when its still in an incomplete stage without affecting the existing implementation

The current state of the post page allows you to:
- View the post information immediately upon opening the post
- View the comments associated with the post
- Refresh the comments
- Switch the sort type for the comments

Current TODOs before this PR is ready:
- Add back the additional top app bar actions (creating cross-post, viewing post source, copying post content
- Fix interactions (comment actions, post actions such as voting, saving, replying, etc)

Future TODOs (to come in separate PRs):
- Bring back post page FAB
- Bring back comment navigation
- Allow comment navigation to navigate between top level comments, and also child comments (which should be capable now with the re-architecture)
- Bring back comment search - with the ability to scroll to specific child comments

Because of this, any future PRs (enhancement or new features) related to post/comment page **should** be targeting the new post page rather than the old one. PRs that fix existing issues with the current post page will still be accepted.

Any thoughts on this approach @micahmo?

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1392, opened 2024-05-29
Fix optimistic voting

micahmo:
## Pull Request Description

This PR fixes a tiny issue with post and comment upvoting where separate counts would not update optimistically, but combined counts (and colors) would.


hjiangsu (approved):
That's funny - I also noticed this issue [when I was working on the post page refactor](https://github.com/thunder-app/thunder/pull/1363/files#diff-e2d256b63eba0948b145acd18b4a27c38fc7aeb55012a913468b2eb79bd24453) and made the changes there (although I did miss lib/post/utils/post.dart) 😅

Anyways, LGTM!

micahmo:
Ah, I see we had very similar fixes! Not sure how much longer your refactor will be but it doesn't hurt to get this in now for existing usage. 😊

PR #1394, opened 2024-05-29
Do not allow changing account while editing

micahmo:
## Pull Request Description

This is a super small PR which prevents changing the account to post as when you're editing an existing post (same for comments).

> Hide whitespace.

### DIFF
--- a/lib/user/widgets/user_selector.dart
+++ b/lib/user/widgets/user_selector.dart
@@ -32,7 +32,7 @@ class UserSelector extends StatefulWidget {
 
   /// Whether the user is allowed to change the active account
   /// (e.g., it should not be allowed during edit)
-  final bool allowChange;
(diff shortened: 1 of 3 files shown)


lib/user/widgets/user_selector.dart
```diff
@@ -30,6 +30,10 @@ class UserSelector extends StatefulWidget {
   final String? parentCommentActorId;
   final void Function(CommentView)? onParentCommentChanged;
 
+  /// Whether the user is allowed to change the active account
+  /// (e.g., it should not be allowed during edit)
+  final bool allowChange;
```
hjiangsu, line 35: Just one nitpick on the naming of this: what do you think of `enableAccountSwitching`?

micahmo: Done!

PR #1395, opened 2024-05-29
Fix superscript/subscript parsing

micahmo:
## Pull Request Description

This is a small PR which fixes a markdown parsing issue I noticed where superscript/subscript would be rendered inside of inline code block (when they shouldn't be).

> As a nice side-effect, I believe this also fixes the problem that had been commented in code as not supported.
> `This is a text^superscript^ and this is a text~subscript~`

I also re-tested with [this comment](https://lemmy.ca/comment/7678098), and everything still looks good!

### Demo

Found in https://programming.dev/comment/9864839

| Before | After |
|--------|--------|
| image | image |


hjiangsu (approved):
LGTM!

PR #1396, opened 2024-05-29
Do not render images too wide in spoiler body

micahmo:
## Pull Request Description

This is a super small PR which fixes an issue where images would be rendered too wide in spoiler blocks.

| Before | After |
|--------|--------|
| image | image |


hjiangsu (approved):
LGTM!

PR #1400, opened 2024-05-30
A few client-side notifications improvements

micahmo:
## Pull Request Description

This PR fixes a couple of minor client-side issues related to notifications.
* Fixed an issue where the group notification was not generated for inbox notifications from UnifiedPush
* Improve the account lookup for UnifiedPush notifications so that it can't be confused by identical usernames on different instances
* Remove images from contents (which aren't formatted properly) and replace with placeholder text


hjiangsu (approved):
LGTM

PR #1404, opened 2024-05-31
Use experimental flag for push notifications

micahmo:
## Pull Request Description

This is a super small PR which puts the new push notifications feature behind the experimental flag rather than debug mode. This allows us to do testing in the field without using a debug build.


hjiangsu (approved):
LGTM!

PR #1405, opened 2024-05-31
Small thumbnail preview improvements

micahmo:
## Pull Request Description

This is a small PR which adds a placeholder icon for videos without a thumbnail, and also adds standard fallback icons in case any image fails to load.

I also aligned the placeholder colors a little bit.

https://github.com/thunder-app/thunder/assets/7417301/482ee099-8fd6-40b4-90ac-14adbe1b1b68


hjiangsu (approved):
LGTM! Just one small comment

lib/shared/media_view.dart
```diff
@@ -278,6 +278,11 @@ class _MediaViewState extends State<MediaView> with SingleTickerProviderStateMix
           fit: widget.allowUnconstrainedImageHeight ? StackFit.loose : StackFit.expand,
           alignment: Alignment.bottomLeft,
           children: [
+            if (widget.postViewMedia.media.first.thumbnailUrl?.isNotEmpty != true)
```
hjiangsu, line 281: This could potentially clash with the warning icon for NSFW posts - we can add a quick check for that here:
```suggestion
            if (!blurNSFWPreviews && widget.postViewMedia.media.first.thumbnailUrl?.isNotEmpty != true)
```

micahmo: Good point! I committed your change!

Do you think the same could be true for the other change I made here below?

hjiangsu: Yeah, that might also be possible! I would also apply the checks when the image fails to load

micahmo: I did some testing and found that, regardless of the "blur" setting, if we don't have a thumbnail, we'll still display the warning icon, due to this line.

https://github.com/thunder-app/thunder/blob/768819324a39c2b7e00249b2f5df491e128c75be/lib/shared/media_view.dart#L292

So I think I will not display any other icon if the post is NSFW.

Let me know if this looks good to you now.

PR #1408, opened 2024-05-31
Show post/comment language

micahmo:
## Pull Request Description

* This PR adds the ability to show the post language in the post metadata. If the language is `Undetermined`, nothing is displayed (matching the web UI).
* There is also the ability to see the post/comment and languages in the bottom sheet (and we could possibly extend this area to include more metadata in the future, so people can see more info about posts/comments without cluttering up the default view).

https://github.com/thunder-app/thunder/assets/7417301/c04dff24-f973-4996-8d86-ac31ac7f3301


hjiangsu:
I think this looks good! Just a quick UI comment - for the language metadata in the bottom sheets, can we make it so that the background is transparent, rather than the dark grey?

I feel like the dark grey background brings a bit too much attention to the metadata. Let me know your thoughts!

micahmo:
> I think this looks good! Just a quick UI comment - for the language metadata in the bottom sheets, can we make it so that the background is transparent, rather than the dark grey?
> 
> I feel like the dark grey background brings a bit too much attention to the metadata. Let me know your thoughts!

Of course! I simplified the bottom sheet to use the same `LanguagePostCardMetaData` as the post listing itself. Let me know if this looks good.

| image |
| - |

hjiangsu (approved):
LGTM!

PR #1409, opened 2024-06-01
Ability to view entire post as a different account

micahmo:
## Pull Request Description

This PR adds the ability to view a post as a different account. This allows you to perform all post-related actions, like voting/saving/etc. As soon as you navigate back, your original account will be restored. If the post is not found on the instance you switch to, a warning will be shown and the account will be reverted.

> Hide whitespace.

Note to self: newly created comments aren't getting added automatically. I think this should be allowed, if the create comment page's user matches the post page's user, so it's just something to look into, but not a big deal.

## Issue Being Fixed

Issue Number: #341

## Screenshots / Recordings

Apologies for the delay in the recording where I'm waiting for federation. 😊

https://github.com/thunder-app/thunder/assets/7417301/f26764af-67ae-441e-ba23-81a968961370


hjiangsu (approved):
I just have to say, this is super cool (and equally surprised that it seems like there weren't too many changes required to implement this)!

Looked through the code and it LGTM. Have not tested it physically but we can merge it in first!

micahmo:
> surprised that it seems like there weren't too many changes required to implement this

Yeah, the temporary account switching framework has come in very handy and makes these kinds of features pretty easy to implement now, and that combined with how the new Post Page reacts to the bloc event, which can emit the entire post object, made this pretty straightforward. 😊

Let me know if you can think of any other places where it would make sense to implement a feature like this!

Issue #1413, opened 2024-06-02
New Post Page Followups


micahmo:
### Bug Description

This issue is for tracking additional work that needs to be done to the new post/comment page as a followup to #1363 in order to bring it in parity with the old one. The new one is still behind an experimental flag, so these things aren't super urgent, but should be done before releasing it for general availability.

#### Known Issues (these are things intentionally deferred from the initial implementation)
* [x] Bring back FAB
* [x] Bring back comment navigation
	* [x] Allow comment navigation to navigate between top level comments, and also child comments
	* [x] Ensure that the first navigation goes to the top of the first comment (see https://github.com/thunder-app/thunder/pull/1496#issuecomment-2263800481)
* [x] Bring back comment search - with the ability to scroll to specific child comments

#### Additional Issues (these have been found while daily driving the new page)
- [x] Collapsing comment threads does not show any animation (for the child comments being hidden)
<details>

https://github.com/thunder-app/thunder/assets/7417301/7f0b3b76-9df4-4c41-8fb7-a028f3cd69e8

</details>

- [x] Collapsing a comment thread with only deferred child comments does not hide the placeholder (this is especially confusing because if you try to load the deferred comments after hiding them, they'll disappear!)
<details>

https://github.com/thunder-app/thunder/assets/7417301/7f0b3b76-9df4-4c41-8fb7-a028f3cd69e8

</details>

- [x] Child comment threads are indented too much
<details>

![image](https://github.com/thunder-app/thunder/assets/7417301/3be61961-b3ff-4951-99f9-cdcd1e838e02)

</details>

- [x] Using the "navigate to post" feature after creating a post doesn't work

<details>

https://github.com/thunder-app/thunder/assets/7417301/8f2f5304-40b7-43c7-b6cd-c3cf99b55d50

</details>

- [x] Jumping to comment context (e.g., from notifications) shows the old page
- [x] Threads with many comments show a spinner at the bottom (maybe paging isn't working?)

<details>

https://github.com/thunder-app/thunder/assets/7417301/f2fd31b4-9b7b-474d-b854-89a8d0be4c03

</details>

- [x] There is a weird gap in the cross-posts widget

<details>

https://github.com/thunder-app/thunder/assets/7417301/f4ab620f-a847-4448-9382-8ab926c46a98

</details>

- [x] There is no pull-down to refresh (this isn't a big deal, and if the pulling down is either contradicting with another gesture or causing lots of rebuilds, maybe we don't need it)
[...]

PR #1416, opened 2024-06-03
Fix issue where navigating to post after post creation throws error

hjiangsu:
## Pull Request Description

This is a small PR which fixes an issue where navigating to the post page after creation would throw an error. The issue comes from the fact that the new post page only accepts `postViewMedia` as a parameter (for now) rather than both `postViewMedia` or `postId`. Since the snackbar attempts to pass in a `postId`, the new post page is unable to use that information to load the post.

Luckily, the `onPostSuccess` returns back the full  `postViewMedia`, so we can just use that!

## Issue Being Fixed

Issue Number: #1413 

## Checklist

### DIFF
--- a/lib/post/utils/navigate_create_post.dart
+++ b/lib/post/utils/navigate_create_post.dart
@@ -68,7 +68,7 @@ Future<void> navigateToCreatePostPage(
                     l10n.postCreatedSuccessfully,
                     trailingIcon: Icons.remove_red_eye_rounded,
                     trailingAction: () {
-                      navigateToPost(context, postId: postViewMedia.postView.post.id);
+                      navigateToPost(context, postViewMedia: postViewMedia);
                     },
                   );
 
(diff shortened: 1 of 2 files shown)


lib/post/utils/navigate_create_post.dart
```diff
@@ -68,7 +68,7 @@ Future<void> navigateToCreatePostPage(
                     l10n.postCreatedSuccessfully,
                     trailingIcon: Icons.remove_red_eye_rounded,
                     trailingAction: () {
-                      navigateToPost(context, postId: postViewMedia.postView.post.id);
+                      navigateToPost(context, postViewMedia: postViewMedia);
```
micahmo, line 71: Would it be worth removing the `postId` parameter from from `navigateToPost`? If the old post page needs it, maybe we can pass it farther along down the line.

micahmo (approved):
Nice, glad this was an easy fix!

lib/post/utils/navigate_create_post.dart
hjiangsu: Hmm, I tried this and noticed that there are a couple of widgets that depend on the `postId`. Primarily, the modlog and the report page. I think we can keep this in for now since it would add a bit more logic to pre-fetch the post information!

Also, the plan is to re-introduce the `postId` for scenarios that require it (or don't have the full post view context)

micahmo: Got it!

PR #1429, opened 2024-06-05
Fix collapse animation with deferred comments

hjiangsu:
## Pull Request Description

This PR fixes an issue where deferred comments were not being collapsed properly, and thus was always shown even when the parent comment was collapsed.

> Collapsing comment threads does not show any animation 

@micahmo I'm not sure if this also fixes this issue, so let me know! For me, it feels like if "Hide Parent Comment when Collapsed" is enabled, the animation is instantaneous whereas if it's disabled, it feels like a more proper "shrinking" animation.

## Issue Being Fixed

Issue Number: #1413

## Screenshots / Recordings

> I'm trying out the macOS build here, so the recording is using that!

https://github.com/thunder-app/thunder/assets/30667958/eda4bc4f-24f3-4a1a-9d45-d862f2b96266

## Checklist

### DIFF
--- a/lib/comment/widgets/comment_card.dart
+++ b/lib/comment/widgets/comment_card.dart
@@ -104,6 +104,21 @@ class _CommentCardState extends State<CommentCard> with SingleTickerProviderStat
   /// Whether we should display the comment's raw markdown source
   bool viewSource = false;
 
+  /// Animation controller for AdditionalCommentCard
+  late final AnimationController _controller = AnimationController(
+    duration: const Duration(milliseconds: 100),
+    vsync: this,
+  );
+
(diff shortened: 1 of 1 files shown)


micahmo (approved):
Awesome, this definitely fixes the problem with deferred comments!

With "Hide Parent Comment when Collapsed" disabled, there is still no animation for collapsing full threads, and even collapsing deferred comments seems to have an interesting curve.


https://github.com/thunder-app/thunder/assets/7417301/26464f15-745b-497d-8e28-5993e6307d53

hjiangsu:
Thanks for confirming! I'll try to fix that in a separate PR

Issue #1220, opened 2024-03-21
Post text preview not rendering markdown


Fxsch:
### Bug Description

The text preview of posts isn't rendering markdown, as you can see in this screenshot:
![Screenshot_20240321-114038_Thunder](https://github.com/thunder-app/thunder/assets/77849217/ed0815de-2a69-4399-9d46-25dd2d6cfcba)

### Expected Behaviour

Markdown should be rendered, like it is when you click on a post:
![Screenshot_20240321-114043_Thunder](https://github.com/thunder-app/thunder/assets/77849217/906c1164-be6f-4861-b242-31d0362e28d9)

### Steps to Reproduce

1. Look at post that uses markdown
2. No markdown rendering in preview

### App Version

0.3.0-1

### Device

Nothing Phone (1)

### OS

Android 13

PR #1431, opened 2024-06-05
Fix post text preview showing raw markdown

hjiangsu:
## Pull Request Description

This PR resolves an issue where the text preview for a post was showing raw markdown. To fix this, I used the `markdown` package to parse the markdown into it's AST tree, and took the `textContent` from the generated nodes. This seems to work for the most part, but may not fully render proper spacing for certain markdown.

Additionally, I made the text preview colour slightly lighter so that it doesn't clash with the main post title!

## Issue Being Fixed

Issue Number: #1220

## Screenshots / Recordings

| Before | After |
|-|-|
|image | image|
|image | image|
|image | image |

## Checklist

### DIFF
--- a/lib/community/widgets/post_card_view_comfortable.dart
+++ b/lib/community/widgets/post_card_view_comfortable.dart
@@ -3,8 +3,8 @@ import 'package:flutter/services.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:html_unescape/html_unescape_small.dart';
-
 import 'package:lemmy_api_client/v3.dart';
+import 'package:markdown/markdown.dart' hide Text;
 
 import 'package:thunder/account/bloc/account_bloc.dart';
 import 'package:thunder/community/utils/post_card_action_helpers.dart';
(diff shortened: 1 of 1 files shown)


micahmo (approved):
This looks good!

Keep in mind we already have a technique to convert markdown to plaintext by passing it through HTML. So maybe we could consolidate techniques.

https://github.com/thunder-app/thunder/blob/8c8ce53d2d1e60272e4020be1e590de68b100fc3/lib/notification/utils/local_notifications.dart#L100-L101

Although in this case, the HTML is needed anyway for the rich text notification, so it doesn't hurt to perform both steps.

hjiangsu:
Ahh right - I completely forgot about that. I can go ahead and see if using the HTML parser works better!

hjiangsu:
I adjusted the code using the same technique as what you referenced earlier. It should be good now!

micahmo (approved):
LGTM. The other technique parsed twice, and used `body` for the first pass and `documentElement` for the second. To be honest I don't remember why that was needed, but if this looks good in your testing, we'll go with it!

hjiangsu:
Ahh yeah weird... I didn't notice anything off when I tested it so it might be fine? I'll go ahead and merge this in and I'll monitor if anything looks off!

PR #1432, opened 2024-06-06
Add database import/export

micahmo:
## Pull Request Description

This PR adds a way to export and import the Thunder database. This can be helpful when performing a new installation to bring back all of the previously saved customizations that are outside of the preferences. This should also allow us to put even more info in the db without worrying that the user won't have any way to preserve it.

I have also added some descriptions to "Save Settings" and "Export Database" so the user understands what data is included. I've added snackbars for import/export so it's clear when the operations have succeeded.

Finally, a confirmation message is shown when exporting the database to warn the user that it may contain sensitive data (i.e., their JWT) and not to share it with anyone.

https://github.com/thunder-app/thunder/assets/7417301/ee7f2cfd-933b-4872-be6c-4c2662e121df

### DIFF
--- a/lib/core/database/database.dart
+++ b/lib/core/database/database.dart
@@ -2,6 +2,8 @@ import 'dart:io';
 
 import 'package:drift/drift.dart';
 import 'package:drift/native.dart';
+import 'package:flutter/material.dart' hide Table;
+import 'package:flutter_file_dialog/flutter_file_dialog.dart';
 import 'package:path/path.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:sqlite3/sqlite3.dart';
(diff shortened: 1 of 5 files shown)


lib/core/database/database.dart
```diff
@@ -38,3 +40,46 @@ LazyDatabase _openConnection() {
...
+  if (filePath != null) {
+    final Directory dbFolder = await getApplicationDocumentsDirectory();
+    final File file = File(join(dbFolder.path, 'thunder.sqlite'));
+
+    try {
+      // Read the selected db file
+      final List<int> bytes = await File(filePath).readAsBytes();
+
+      // Write the file out the location we read it from
+      await file.writeAsBytes(bytes, flush: true);
```
hjiangsu, line 73: While this works, we might want to think about doing this the way drift documentation suggests!

https://drift.simonbinder.eu/docs/examples/existing_databases/#exporting-a-database

hjiangsu (approved):
Just one comment up for discussion!

lib/core/database/database.dart
micahmo: Alrighty, I gave this a shot, but it didn't quite work the way I was hoping. The Drift documentation assumes you already have a path, but of course we need to prompt the user for a path using the save dialog. I found some weird behavior from `flutter_file_dialog` where the save file path didn't correspond to a real path on the device, so Drift could find the value to save into.

I even tried the `file_picker` package (which seems to be more popular), and it had a slightly different but similar issue where the path it returned from saving a file was a real path, but not the one I picked.

Maybe the path I'm picking is hitting an edge case in these packages, but at least with the originally implementation, `flutter_file_dialog` handles both picking the file path, and also transferring the data internally, so it doesn't matter that it returns an unusual path.

What do you think?

hjiangsu: Ahh okay, that's alright - we'll roll with this for now then! Thanks for trying it out though 😄

PR #1434, opened 2024-06-06
Add support for user labels

micahmo:
## Pull Request Description

This PR adds support for adding customized tags/labels for other users that you encounter. This is a popular feature of many Reddit and now Lemmy apps, and it lets you make a note of anything significant that you want to remember about a particular user. 

Note: This should probably not be added until #1432 so people have a way to save their labels!

> Hide whitespace.

https://github.com/thunder-app/thunder/assets/7417301/bcbfd209-93ed-4072-bd26-516e56520e1f

https://github.com/thunder-app/thunder/assets/7417301/d262a14e-e27d-4230-9848-b6bad6053f4e

### DIFF
--- a/lib/account/models/user_label.dart
+++ b/lib/account/models/user_label.dart
@@ -5,6 +5,8 @@ import 'package:thunder/core/database/database.dart';
 import 'package:thunder/main.dart';
 import 'package:thunder/utils/instance.dart';
 
+/// Represents a UserLabel, which is used to associate a textual description along with a Lemmy user.
+/// Contains helper methods to load/save corresponding objects in the database.
 class UserLabel {
   /// The ID of the object in the database (should never need to be set explicitly).
   final String id;
--- a/lib/account/models/user_label.dart
+++ b/lib/account/models/user_label.dart
@@ -1,0 +1,86 @@
+import 'package:drift/drift.dart';
+import 'package:flutter/foundation.dart';
+
+import 'package:thunder/core/database/database.dart';
+import 'package:thunder/main.dart';
+import 'package:thunder/utils/instance.dart';
+
+class UserLabel {
+  final String id;
+  final String username;
+  final String label;
+
+  const UserLabel({
+    required this.id,
+    required this.username,
+    required this.label,
+  });
+
+  UserLabel copyWith({String? id}) => UserLabel(
+        id: id ?? this.id,
+        username: username,
+        label: label,
+      );
+
+  static Future<UserLabel?> upsertUserLabel(UserLabel userLabel) async {
+    try {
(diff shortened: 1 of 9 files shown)


lib/shared/dialogs.dart
```diff
@@ -40,6 +42,11 @@ Future<T?> showThunderDialog<T>({
                 ),
         ),
         actions: [
+          if (tertiaryButtonText != null)
```
hjiangsu, line 45: Could we make this tertiary action be on the left-hand side of the dialog? This might help with accidental taps on this action!

lib/account/models/user_label.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:drift/drift.dart';
+import 'package:flutter/foundation.dart';
+
+import 'package:thunder/core/database/database.dart';
+import 'package:thunder/main.dart';
+import 'package:thunder/utils/instance.dart';
+
+class UserLabel {
+  final String id;
```
hjiangsu, line 9: Add comments to each of the parameters!

lib/account/models/user_label.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:drift/drift.dart';
+import 'package:flutter/foundation.dart';
+
+import 'package:thunder/core/database/database.dart';
+import 'package:thunder/main.dart';
+import 'package:thunder/utils/instance.dart';
+
+class UserLabel {
```
hjiangsu, line 8: Add a description of what this class is used for (just for clarity in the future)!

hjiangsu (approved):
This is super cool, thanks for implementing it! Just have some minor nitpicks but otherwise LGTM!

(I'll wait until #1432 as you mentioned 😄)

lib/shared/dialogs.dart
micahmo: Sure! Let me know if this looks good.

| image |
| - |

lib/account/models/user_label.dart
micahmo: Done!

lib/account/models/user_label.dart
micahmo: Done!

micahmo:
P.S. I added a downgrade strategy so users call roll back to a previous version without breaking the database.

lib/shared/dialogs.dart
hjiangsu: Looks good to me!

micahmo: Just a heads up, it looks like dialogs without a tertiary button now have the buttons left-aligned, so I will take a look at that when I can!

PR #1435, opened 2024-06-07
Always have closable snackbars

micahmo:
## Pull Request Description

This is a small PR which makes the snackbar always closable by default. I figured this would good to add because not all people know that it can be swiped away, and sometimes they appear for a long time (if there is a lot of text). And also, depending on where the snackbar appears, the swipe gesture can interfere can system gestures. The close button doesn't take up much space, and it is still compatible with other trailing icons/actions.

I also changed the snackbar actions from using `GestureDetector`s to using `InkWell`s for the tap animation.

> Review without whitespace.

### DIFF
--- a/lib/shared/snackbar.dart
+++ b/lib/shared/snackbar.dart
@@ -16,7 +16,7 @@ void showSnackbar(
   IconData? leadingIcon,
   Color? trailingIconColor,
   IconData? trailingIcon,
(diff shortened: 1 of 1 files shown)


ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1441, opened 2024-06-10
Fix dialog button alignment

micahmo:
## Pull Request Description

This is a super small PR to fix the alignment of dialog buttons when there is no tertiary button.


micahmo:
I pushed another small change to allow autocomplete suggestions.

hjiangsu (approved):
LGTM!

Issue #1443, opened 2024-06-12
Mark as read not persistent/working


gtrcoi:
### Bug Description

Mark as read post swipe gesture does not mark post as read persistantly.

`settings > general > "Mark Read After Viewing Media` does not mark post as read at all.

Previously mentioned gesture action marks post as read for a moment (1 secondish) before reverting to unread status.

**Clicking on post to view post and comments does mark as read as expected.**

### Expected Behaviour

Viewing media or using assigned gesture action to mark post as read should mark post as read.

### Steps to Reproduce

Set gesture action in `settings > gestures > Post Swipe Actions` to mark as read. Use gesture action.

Enable `settings > general > "Mark Read After Viewing Media`, then view media (without opening post to view comments)

### Additional Context

Started happening after a recent update to the app, iirc a week or so ago.

### App Version

0.4.0

### Device

Pixel3

### OS

Android 12

hjiangsu:
Thanks for submitting the issue. I found the cause and have a PR with the fix. This should land in the next nightly build!

PR #1444, opened 2024-06-12
Fix mark post as read when previewing media on 0.19.4

hjiangsu:
## Pull Request Description

This PR fixes an issue where instances on v0.19.4 cannot mark a post as read when tapping on the media thumbnail. This is because 0.19.4 removed `postId` from the MarkPostAsRead endpoint.

## Issue Being Fixed

Issue Number: #1443

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/30667958/21bc2e0a-1d71-403a-b5d8-23dd8a841746

## Checklist

### DIFF
--- a/lib/post/utils/post.dart
+++ b/lib/post/utils/post.dart
@@ -32,11 +32,21 @@ Future<bool> markPostAsRead(int postId, bool read) async {
 
   if (account?.jwt == null) throw Exception('User not logged in');
 
-  MarkPostAsReadResponse markPostAsReadResponse = await lemmy.run(MarkPostAsRead(
-    auth: account!.jwt!,
-    postId: postId,
-    read: read,
-  ));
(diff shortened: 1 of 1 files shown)


micahmo (approved):
LGTM!

I wonder if it would hurt to pass both `postId` and `postIds` for every call to support both versions without an `if` statement? Not a big deal either way.

hjiangsu:
> I wonder if it would hurt to pass both postId and postIds for every call to support both versions without an if statement? Not a big deal either way.

That was something I thought of as well, but I'm not entirely sure of how the backend processes requests (whether it rejects them or ignores them) so I just played it safe. I am thinking of maybe dropping support for 0.18.x in the near future (maybe in v0.6.0, but nothing set in stone) so it might not be a big issue regardless!

Issue #1450, opened 2024-06-14
Sidebar optimization for a huge amount of communities.


kratoz29:
### Bug Description

If you are not subscribed to a ton of communities then pulling the sidebar goes smoothly, but if you have plenty of them it lags quite a bit, using the gesture or the "hamburger" button.

### Expected Behaviour

That the sidebar triggers smoothly not depending on the amount of subscribed communities. 

### Steps to Reproduce

Subscribe to a lot of communities, if you are not already.
Use the gesture, or the hamburger button to trigger the sidebar.

### App Version

0.4.0

### Device

Poco F2 Pro (lmi)

### OS

Android 14

PR #1452, opened 2024-06-15
Fix sidebar lag when subscribed to a large amount of communities

hjiangsu:
## Pull Request Description

This PR fixes an issue where the drawer experiences lag when there are a large amount of subscribed communities. To do this, I migrated over to using slivers and a CustomScrollView which improves scrolling performance.

Additionally, I fixed an issue with the search page where if you subscribe to a community while in the "View All" view, it would refresh and reset the view back to the trending view.

## Issue Being Fixed

Issue Number: #1450

## Checklist


micahmo (approved):
LGTM!

Issue #1375, opened 2024-05-21
Central place to view all drafts


codenyte:
### Is Your Feature Request Related to a Problem?

I often start writing a comment, switch to another app to find a link I want to insert into my comment, then come back to Thunder and notice that the app reloaded. My comment is saved as a draft, but I often can't find the post anymore.

### Feature Description

Some central place that shows all saved comment/post drafts

### Additional Context

Perhaps this feature already exists, but I'm too dumb to find it?

PR #1455, opened 2024-06-17
Store drafts in database

micahmo:
## Pull Request Description

This PR is the first step in introducing a way to manage drafts. Initially, we'll just be migrating existing drafts from `SharedPreferences` to a new table the database. Once the migration has been shown to be stable, and users are able to save/load drafts with no problems, we can start adding the UI.

> As part of these changes, I discovered that edits for posts/comments were not currently being saved as drafts, so I fixed that. Then I discovered that clearing a draft would reset contents to empty, rather than to the original content we're editing, so I fixed that as well.

I tested all the scenarios I could think of including upgrades as well as functionality going forward, but I'd be happy to have additional testing!

## Issue Being Fixed

Issue Number: Necessary starting point for #1375

### DIFF
--- a/lib/account/models/draft.dart
+++ b/lib/account/models/draft.dart
@@ -1,0 +1,144 @@
+import 'package:drift/drift.dart';
+import 'package:flutter/foundation.dart';
+import 'package:thunder/core/database/database.dart';
+import 'package:thunder/core/database/type_converters.dart';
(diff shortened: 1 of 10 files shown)


hjiangsu (approved):
Thanks for doing this! This is definitely going to make managing drafts much better in the future.

I only did a skim through of it, but everything LGTM! 😄

Issue #1454, opened 2024-06-17
Profiles seem to show Subscriptions feed


orangeNgreen:
### Bug Description

Clicking on a user profile view their posts just seems to show the Subscribed feed, rather than that user’s posts. 

### Expected Behaviour

The profile should show a user’s posts as opposed to the Subscriptions feed. 

### Steps to Reproduce

1. Click a user. 
2. Try to view their posts. 

### Additional Context

Here are some pictures with various random users all showing the same posts on their profile, which are also the same as my Subscribed feed.

![IMG_4865](https://github.com/thunder-app/thunder/assets/140240648/bfa54cf5-c343-42ff-8e93-b155ff034b34)

![IMG_4864](https://github.com/thunder-app/thunder/assets/140240648/5fd118d6-5d7b-42b7-b6f0-6a88d571b8e7)

![IMG_4863](https://github.com/thunder-app/thunder/assets/140240648/fefe0692-f702-424f-91ca-a05b01d1057e)

### App Version

0.5.0-1

### Device

iPhone 15 Pro

### OS

17.5.1

hjiangsu:
Thanks for the report! This was caused with the recent switch to use the Lemmy account's feed settings rather than the in-app settings for determining the default feed type/sort!

PR #1436, opened 2024-06-08
Add additional account settings

hjiangsu:
## Pull Request Description

This PR adds a bunch of additional account settings that sync up with Lemmy, including:
- Display name
- Profile bio
- Email
- Matrix user ID
- Default feed type - this setting will override the existing app setting
- Default feed sort - this setting will override the existing app setting
- Show NSFW content

As a result of this, I also cleaned up the user settings page and moved discussion languages and blocking to separate pages.

Note: I opted to only sync settings that adjust underlying logic. I left out most of the settings which change the UI in some way (since I believe that's better handled in the app settings). These are the settings I omitted from including:
- Interface language
- theme
- Blur NSFW content
- Auto-expand media
- Show Avatars
- Bot account toggle
- Show notifications for new posts
- Send notifications to email
- Open links in a new tab

Some settings that are still missing but would be good to add in the future include:
- Avatar
- Banner
- Two-factor auth
- Exporting/importing settings

I did some basic testing for all of the new settings introduced, but there might still be some edge cases I missed!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

**Account Settings**

https://github.com/thunder-app/thunder/assets/30667958/e5db433f-e823-4ec9-a342-8156c76960e4

**Discussion Languages**

https://github.com/thunder-app/thunder/assets/30667958/b72b3ced-bfee-43ab-a05c-c21fa98fc50f

**Block Management**

https://github.com/thunder-app/thunder/assets/30667958/f75ffa7c-364b-4374-8c69-84292c1c61f1

**Feed Account Sort/Type**
Keep an eye out on the feed sort type

https://github.com/thunder-app/thunder/assets/30667958/c40af8cf-eae6-44ef-a34a-b395c56354a4

## Checklist

- [ ] Did you update CHANGELOG.md?
- [x] Did you use localized strings where applicable?
- [x] Did you add `semanticLabel`s where applicable for accessibility?

PR #1456, opened 2024-06-17
Fix user profile showing general feed

hjiangsu:
## Pull Request Description

This PR fixes a regression where navigating to the user's profile page would show the general feed rather than the user's posts. This was caused by #1436, where a default fallback was added in to fetch the current account's default feed type. That default fallback has been removed, and the calling function must now explicitly pass in the `postListingType` in order to fetch the general feeds.

## Issue Being Fixed

Issue Number: #1454

## Checklist


micahmo (approved):
LGTM!

PR #1463, opened 2024-06-21
Refactor Inbox Page

hjiangsu:
## Pull Request Description

This is a draft PR which refactors the inbox page to be more consistent with the rest of our UI. In particular:
- Switch to using TabBar and slivers instead of the chips and ListViews for replies, mentions, messages
- Adjust the mentions section to use the same CommentReference as the replies (messages will stay the same for now since it'll need to be overhauled)
- Improves the fetching logic for inbox replies, mentions, and messages. It should now only fetch the appropriate information depending on the active tab. This should reduce API calls to the server since we're not unnecessarily fetching other information

> Note: Because the underlying fetching logic was changed, we may need to do some quick testing on the push notifications page to make sure it still works as expected

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Here are some initial screenshots!

|**Replies**|**Mentions**|
|-|-|
|simulator_screenshot_7B4329DC-09FB-490A-8ED4-2D01E25125A4|simulator_screenshot_13A5F6AE-A2DE-451D-82C4-832D84C928F9|

## Checklist


hjiangsu:
This should be ready to go now! I haven't tested out notifications since that's not readily available on iOS yet, but it _should_ work as usual.

@micahmo Feel free to try it out and let me know if you find anything!

micahmo:
Sounds good, I think I will run this rather than doing an in-depth review and let you know if I find anything!

micahmo:
Hey just FYI, I am getting a gray screen exception when tapping on a notification. I can try to debug when I get a chance (not sure when that will be).

micahmo:
One more note: I'm not sure if you intended the "you've reached the bottom" message to have a gray background like it does for comments.

hjiangsu:
> I am getting a gray screen exception when tapping on a notification.

I can try this out on an Android emulator and see what's happening!

> I'm not sure if you intended the "you've reached the bottom" message to have a gray background like it does for comments.

Hmm, I'm not sure what you're referring to - do you have a screenshot of this? The background should be transparent

Also, I noticed an issue where the inbox doesn't refresh properly when switching accounts from authenticated -> anonymous so I'll fix that as well!

micahmo:
> > I am getting a gray screen exception when tapping on a notification.
> 
> I can try this out on an Android emulator and see what's happening!

Thanks!

> > I'm not sure if you intended the "you've reached the bottom" message to have a gray background like it does for comments.
> 
> Hmm, I'm not sure what you're referring to - do you have a screenshot of this? The background should be transparent

I'm just referring to the difference in background color here. Also as I took these screenshots, I realized the text styling is different too. Maybe this should be a shared widget?

|  |  |
|--------|--------|
| image | image | 

---

P.S. I found one more thing. There doesn't appear to be a progress indicator when the inbox is being refreshed.

hjiangsu:
> I am getting a gray screen exception when tapping on a notification.

This should be fixed now!

> I'm just referring to the difference in background color here. Also as I took these screenshots, I realized the text styling is different too. Maybe this should be a shared widget?

I adjusted this as well, so it should now match up with the feed reached end widget

hjiangsu:
> P.S. I found one more thing. There doesn't appear to be a progress indicator when the inbox is being refreshed.

Hmm, I'm seeing a refresh indicator when refreshing the inbox - can you clarify what you mean here?

[refresh.webm](https://github.com/thunder-app/thunder/assets/30667958/002fa2ae-18e5-4ff4-9cc1-b59a0d74b8a6)

hjiangsu:
I'll go ahead and merge this in so that it can be tested in the next nightly! Let me know if you encounter any more issues.

micahmo:
> Hmm, I'm seeing a refresh indicator when refreshing the inbox - can you clarify what you mean here?

Yes, sorry, I think the problem is less about the lack of a refresh indicator and more that something is wrong after navigating back from the notifications page. You can see what I mean here.

https://github.com/thunder-app/thunder/assets/7417301/26576b8f-4f77-406d-80d4-9ab3161a1f9f

hjiangsu:
Ahh I see, thanks for clarifying that! I'll take a look at it

micahmo:
Hey I noticed something else. Actually you can see it in the screen recording. Marking the last notification as read no longer closes the notification page. Not sure if that was intention and/or whether it was caused by your changes.

Issue #1464, opened 2024-06-22
Comment text area stops accepting taps after formatting


sukhmel:
### Bug Description

Sometimes, when writing a markdown text, input area blinks grey and stops accepting taps, making it impossible to open a keyboard after previewing and complicating navigation.

### Expected Behaviour

Text input area should always allow to tap for moving the caret position

### Steps to Reproduce

1. Start writing a comment
2. Let it's length exceed one line
3. Input `**` for creating a bold text

### Additional Context

It seems to never happen when formatted text is on the first line of the input 

IIRC, it sometimes requires to also input the paired `**`, but it may be not necessary 

It was happening before 0.4.0 but I was too lazy to report, sorry 😅

### App Version

0.5.0-1

### Device

Pixel 8 Pro

### OS

Android 14, build AP2A.240605.024

sukhmel:
Completely forgot to add the reproduction video:

https://github.com/thunder-app/thunder/assets/5804169/69ccbd10-9d5c-4d98-91c5-94597ac0dc47

hjiangsu:
Thanks for the report!

I looked into this, and it seems like Flutter's spell checking is causing exceptions when you input ** characters. I'm not sure if there's any way to fix this except for waiting the fix upstream. For now, I can disable spell checking on the inputs.

hjiangsu:
Seems like this issue was already noted down upstream and there is a fix, but we have not yet upgraded to the Flutter version that contains that fix. See https://github.com/thunder-app/thunder/pull/1465#issuecomment-2185282875 for reference.

micahmo:
By the way, I just wanted to chime in here and mention that you can use the markdown toolbar to make a word bold, rather than explicitly typing the `**`. That might help to mitigate this issue!

sukhmel:
Yes, I also noticed after reporting that using format buttons helps, it's just that a habit of inputting text manually is too strong 😅

Thank you, will look forward to an update

PR #1372, opened 2024-05-20
Chore: Upgrade dependencies for Flutter 3.22 compatibility

hjiangsu:
## Pull Request Description

This PR upgrades the necessary dependencies to work with Flutter 3.22. This is still in draft as there are currently some packages which have been overridden in order to allow Flutter 3.22 compatibility. Once the official packages have been updated, this will be taken out of draft status. In particular:
- `swipeable_page_route` - awaiting for PR merge
  - **Issue**: https://github.com/JonasWanke/swipeable_page_route/issues/52
  - **PR**: https://github.com/JonasWanke/swipeable_page_route/pull/54
  - **iOS PR**: https://github.com/JonasWanke/swipeable_page_route/pull/51
    - (I'm not sure if we need to wait for this one.)
- `flutter_inappwebview_android` (transitive dependency) - awaiting for PR merge
  - **Issue**: https://github.com/pichillilorenzo/flutter_inappwebview/issues/2139  
  - **PR**: https://github.com/pichillilorenzo/flutter_inappwebview/pull/2102
  - **See also**: https://github.com/pichillilorenzo/flutter_inappwebview/pull/2140
- `flex_color_scheme` - requires major version upgrade to work with Flutter 3.22 (WIP)
  - **See**: https://docs.flexcolorscheme.com/

There also needs to be some general testing to ensure no breaking changes resulting from this upgrade.

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1465, opened 2024-06-23
Fix text field spell check causing exception on ** text

hjiangsu:
## Pull Request Description

This is a small PR which fixes the error described in #1464. As mentioned in the original issue, this is caused by Flutter's spell checking engine. For now, I've disabled spell checking until a fix is confirmed from upstream.

Additional Info:
This is the stack trace (the exception doesn't explicitly print out the stack trace in the console):
image
image

The issue most likely comes from `final RegExp currentSpanTextRegexp = RegExp('\\b$currentSpanText\\b');` in `flutter/lib/src/widgets/spell_check.dart`.

## Issue Being Fixed

Issue Number: #1464

## Checklist

### DIFF
--- a/lib/comment/view/create_comment_page.dart
+++ b/lib/comment/view/create_comment_page.dart
@@ -391,6 +391,7 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
                                         minLines: 8,
                                         maxLines: null,
                                         textStyle: theme.textTheme.bodyLarge,
+                                        spellCheckConfiguration: const SpellCheckConfiguration.disabled(),
                                       ),
                                     ),
                                     crossFadeState: showPreview ? CrossFadeState.showFirst : CrossFadeState.showSecond,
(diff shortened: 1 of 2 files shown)


hjiangsu:
Update: I found the corresponding Flutter issue, which mentions that it should be fixed: https://github.com/flutter/flutter/issues/136032

Not sure what version it was fixed, but should double check if this is an issue once we upgrade to Flutter 3.22

micahmo (approved):
Nice find and fix! We can merge this for now as long as we remember to revert it once the upstream fix is included (I know spellcheck is very important to folks!).

On that topic, is there any progress on #1372?

hjiangsu:
> On that topic, is there any progress on https://github.com/thunder-app/thunder/pull/1372?

Not yet, still waiting for packages to be updated! I also saw that `flex_color_scheme` isn't compatible with Flutter 3.22 either, so we'll have to wait until the next version of that as well (updated the description on #1372 already).

Issue #1442, opened 2024-06-10
MP4 Videos originating from catbox.moe not work


DragonLee-X:
### Bug Description

Click on a catbox.moe linked video to watch it, the internal player pops up and buffers for 4 seconds, then gives an error.
The error popup asks to play the file in the browser, if you click the arrow, the file plays fine in the external browser.

This is a new issue since the new mediaplayer introduced in 0.4.0

### Expected Behaviour

Expect the video to play in the internal media player.

### Steps to Reproduce

Open a linked video (mp4 format) by clicking on the thumbnail or the link.

New screen opens with the internal media player

Mediaplayer buffers for ~4 seconds then gives an error: Video Can't be played. Retry.
Another error pops up saying "Failed to load video. Open link in browser?" 

Click the arrow on the error-popup and the video opens and plays fine in the external browser.

### Additional Context

Example lemmy post:

https://lemmy.world/post/6451438

The streamable source link works fine, the mirrors with mp4 videos do not.

### App Version

0.4.0

### Device

Samsung S23 Ultra

### OS

(samsung) Android 14 latest updates

hjiangsu:
Thanks for the bug report! Unfortunately, I'm unable to reproduce this issue. Can you confirm if this is still occurring?

Additionally, are you on a VPN? Sometimes, catbox doesn't play well with some VPNs and will fail to load content. I would also check to see if any of the following things apply to you under "Connectivity Issues" here: https://catbox.moe/faq.php

micahmo:
I can reproduce this issue on my Android emulator. 

It's the same error I've seen with other videos that `river_player` can't handle, which is `Video player had error androidx.media3.exoplayer.ExoPlaybackException: Source error`. Unfortunately it seems to be coming from ExoPlayer, which I guess `river_player` uses under the hood. I wonder if we need to consider options for a player that supports more codecs or something.

DragonLee-X:
> Thanks for the bug report! Unfortunately, I'm unable to reproduce this issue. Can you confirm if this is still occurring?
> 

Yes it is still occurring.

> Additionally, are you on a VPN? Sometimes, catbox doesn't play well with some VPNs and will fail to load content. I would also check to see if any of the following things apply to you under "Connectivity Issues" here: https://catbox.moe/faq.php
> 
I am not behind a vpn, had one installed but disabled, I removed it to test and it is still occurring. (Plus the video works fine in both firefox and chrome browsers)
[...]

PR #1466, opened 2024-06-23
Migrate Video Player

hjiangsu:
## Pull Request Description

This draft PR migrates the existing `river_player` with Flutter's `video_player`. The main reason for this is to use an actively maintained video player package. Since the `video_player` package does not come with built-in controls, I've created new widgets to handle most, if not all of the currently available features. This includes the following settings:
- Auto fullscreen (with a caveat)
- Muting videos on load
- Looping videos
- Video autoplay
- Default playback speed

To mimic auto-full screen when the device is in portrait-locked mode, I rotate the video player by 90 degrees. This means that the video player doesn't account for the landscape orientation of the phone (this is only while the device is portrait-locked)

TODO:
- [x] Auto-hide controls when video is playing

## Issue Being Fixed

Issue Number: #1442

## Checklist


micahmo:
Hey @hjiangsu I was just going to comment on this PR that I had run across another Flutter app which uses `video_player` directly with their own controls on top of it. I was going to link their code in case it had anything helpful.

https://github.com/j-fbriere/squawker/blob/master/lib/tweet/_video_controls.dart

Then I noticed that you had already marked this PR ready for review? Is that still the case? Would it be worth taking it for a test?

hjiangsu:
Ahh yeah, this PR is ready for review but I was going to wait until 0.6.0 to merge it in! I think we can take a look at how their implementation looks like, and take ideas to improve on the current implementation I have 😅

micahmo:
Just to double-check, your video player is not being used for YouTube videos, right? We would have to introduce a new step which retrieves the raw video URL to pass to the player?

hjiangsu:
> Just to double-check, your video player is not being used for YouTube videos, right? We would have to introduce a new step which retrieves the raw video URL to pass to the player?

Correct! This is only currently being used for the general video player but I would like to extend this to YouTube videos as well.

One idea we can do here is to use a package like https://pub.dev/packages/youtube_explode_dart to possibly retrieve metadata for a given YouTube url (without the need to provide an API key). However, my concern with using something like `youtube_explode_dart` is that it can be subject to break if YouTube itself changes their internal logic.

A (potentially) more robust but less trivial solution would be to convert the youtube link to a piped link (or similar) and retrieve video metadata there. If there's no dart library for this, we might have to build one ourselves to call the proper API endpoints.

micahmo:
Sounds good! You should be good to merge this after the next general release and we can start getting our hands on it.

hjiangsu:
Just merged this in - let me know if you encounter any quirks when using this!

PR #1475, opened 2024-07-08
Make exiting snackbar not closable

micahmo:
## Pull Request Description

This is a super small change, but I thought it would be a good idea to make the exiting snackbar not closable. This is because the duration that the snackbar is visible is very important; it corresponds directly to the period of time during which pressing back again will close the app. The user might think that closing the snackbar "cancels" the timeout. Then they might be confused that pressing back again closes the app, even once the snackbar is gone.

(Note that the snackbar can still be swiped away; this change only affects the close button.)


hjiangsu (approved):
LGTM!

PR #1476, opened 2024-07-08
Show sort on header

micahmo:
## Pull Request Description

This PR adds the sort type to the community and user headers. I noticed that there was no way to immediately see the sort type without either changing it or scrolling up so that it appears in the app bar.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

Note that on most phone screens with a better resolution, the sort type will not wrap down and will instead be in alignment with the other statistics.

| image |
| - |

## Checklist


hjiangsu (approved):
LGTM!

PR #1477, opened 2024-07-08
Fix inbox not refreshing after popping from notifications page

hjiangsu:
## Pull Request Description

This PR fixes an issue mentioned in https://github.com/thunder-app/thunder/pull/1463#issuecomment-2214113668. Whenever the notifications page is popped, the inbox bloc will be refreshed and will pull updates for all inbox types (replies, mentions, messages). This is because we don't know which page/tab the user will be at when the notifications page is popped.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/inbox/bloc/inbox_bloc.dart
+++ b/lib/inbox/bloc/inbox_bloc.dart
@@ -107,6 +107,34 @@ class InboxBloc extends Bloc<InboxEvent, InboxState> {
               ),
             );
             break;
+          case InboxType.all:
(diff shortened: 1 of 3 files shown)


micahmo (approved):
Perfect! I see this happened because of the the optimization of not refreshing all pages at once, which is a great improvement. I like the new `all` option as well. Thanks!

hjiangsu:
Yup, that's what happened 😅. Thanks again for catching this, and let me know if you notice any other quirks with the new page!

PR #1481, opened 2024-07-08
Add user label settings page

micahmo:
## Pull Request Description

This PR adds a settings page to manage user labels.
* All user labels will be listed (regardless of your current instance).
* A label can be edited by tapping on it, or deleted by tapping the x.
* New labels can be created. First you pick the user (only users on your current instance will be selectable), then you write the label.
  * If you happen to "create" a label that already exists, any changes you make in the dialog will be reflected in the list.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/f2a461d5-fc65-499b-a8f8-a8d0f5679245

## Checklist


lib/settings/pages/user_labels_settings_page.dart
```diff
@@ -0,0 +1,203 @@
...
+}
+
+class _UserLabelSettingsPageState extends State<UserLabelSettingsPage> with SingleTickerProviderStateMixin {
+  GlobalKey settingToHighlightKey = GlobalKey();
+  LocalSettings? settingToHighlight;
+
+  List<UserLabel> userLabels = [];
+
+  void _updateChangedUserLabel(({UserLabel? userLabel, bool deleted}) result) {
+    if (result.userLabel != null) {
```
hjiangsu, line 31: Invert the conditional statement to reduce nesting!

```suggestion
    if (result.userLabel == null) return;
```

lib/settings/pages/user_labels_settings_page.dart
```diff
@@ -0,0 +1,203 @@
...
+        // It doesn't exist in our list yet, add it!
+        setState(() => userLabels.add(result.userLabel!));
+      } else if (existingLabel != null) {
+        if (result.deleted) {
+          // It exists in our list and was deleted, so remove it.
+          setState(() => userLabels.removeWhere((userLabel) => userLabel.username == result.userLabel!.username));
+        } else {
+          // It exists in our list but was changed, so update it.
+          setState(() => userLabels[userLabels.indexOf(existingLabel)] = result.userLabel!);
+        }
```
hjiangsu, line 43: This is more of a preference thing (so no need to change anything!), but I generally like to return early on these types of boolean conditions to reduce the logic. Of course, this only works if this is the last thing that needs to be performed.

```suggestion
        if (result.deleted) {
          // It exists in our list and was deleted, so remove it.
          setState(() => userLabels.removeWhere((userLabel) => userLabel.username == result.userLabel!.username));
          return;
        }
        
        // It exists in our list but was changed, so update it.
        setState(() => userLabels[userLabels.indexOf(existingLabel)] = result.userLabel!);
```

hjiangsu (approved):
LGTM! Just a couple of very minor points

lib/settings/pages/user_labels_settings_page.dart
micahmo: Fixed this!

lib/settings/pages/user_labels_settings_page.dart
micahmo: That makes sense! In this case, since there's only one line of logic in the `else`, I think the parallelism of having both statements in conditions helps to see what's going on without too much clutter.

Issue #1447, opened 2024-06-13
☂️ Support for Lemmy 0.19.4


micahmo:
### Is Your Feature Request Related to a Problem?

After reading the following post, I have gleaned that the following items should be supported by Thunder. There may be more things we discover that we should support in the future, but I think these are the big ones. Of course, I'd be open to input from others!

https://join-lemmy.org/news/2024-06-07_-_Lemmy_Release_v0.19.4_-_Image_Proxying_and_Federation_improvements

- [x] Show whether a community is public or local only (https://github.com/LemmyNet/lemmy/pull/4350)
  - PR: ##1449
- [x] Post hiding (https://github.com/LemmyNet/lemmy/issues/1403 and)
  - Issue: #1176
  - PR: #1484
- [x] Show site content warning (https://github.com/LemmyNet/lemmy/pull/4393)
  - PR: #1485
- [x] Sync vote display setting (https://github.com/LemmyNet/lemmy/pull/4450)
- [x] Ability to set custom thumbnail for posts (https://github.com/LemmyNet/lemmy/pull/4425)
  - PR: #1495
- [x] Media management
  - PR: #1323
- [x] Alt text
  - Issue: #1461
  - PR: #1497
  - PR: #1640

### Feature Description

N/A

hjiangsu:
Just a quick update: I've updated https://lemmy.thunderapp.dev to v0.19.5 in case you wanted to do any testing on the new features!

PR #1484, opened 2024-07-09
Add support for hiding posts

micahmo:
## Pull Request Description

This PR adds support for the ability to hide posts introduced in Lemmy 0.19.4.
- There is a new swipe action and a new post sheet action.
- The action color can be customized.
- There is a new icon on posts that indicates whether they are hidden or not.

Unfortunately there does not seem to be a Lemmy account setting for whether to show/hide hidden posts in the feed, so for now I've added a Thunder setting. But we can migrate this to an account setting if it ever becomes available.

Notes:
* If you configure the hide gesture and then use an instance which does not support the feature, the gesture will not be available. The post sheet action will also be hidden.
* Unlike the web UI, hiding a post from Thunder will immediately dismiss it from the feed (if the user has chosen not to display hidden posts).
  * Auto-dismissal from the user's own user page is not yet supported, since it does not yet use the same feed widget as the general feed and other user pages. However, the hiding feature still works.

## Issue Being Fixed

Issue Number: #1447

## Screenshots / Recordings

Demo 1: Hide via swipe and button.

https://github.com/thunder-app/thunder/assets/7417301/5d3c1123-f486-4a46-b645-65aa14a58f0e

Demo 2: Hide via swipe and button, respecting show hidden. Then unhide.

https://github.com/thunder-app/thunder/assets/7417301/57ffcca0-65a6-4a35-a77e-4304205b5466

Demo 3: Swipe and button are not visible on older instances.

https://github.com/thunder-app/thunder/assets/7417301/69ee5f57-e100-46d1-a219-b553675178c0

## Checklist


lib/community/utils/post_card_action_helpers.dart
```diff
@@ -243,6 +244,15 @@ final List<ExtendedPostCardActions> postCardActionItems = [
     getOverrideIcon: (postView) => postView.read ? Icons.mark_email_unread_rounded : Icons.mark_email_read_outlined,
     shouldEnable: (isUserLoggedIn) => isUserLoggedIn,
   ),
+  ExtendedPostCardActions(
+    postCardAction: PostCardAction.hide,
+    label: l10n.hide,
+    getOverrideLabel: (context, postView) => postView.hidden == true ? l10n.unhide : l10n.hide,
+    icon: Icons.close_rounded,
```
hjiangsu, line 251: Thoughts about using `visibility` and `visibility_off` for the icons?

lib/community/widgets/post_card_view_comfortable.dart
```diff
@@ -113,6 +114,15 @@ class PostCardViewComfortable extends StatelessWidget {
               child: Text.rich(
                 TextSpan(
                   children: [
+                    if (postViewMedia.postView.hidden == true)
+                      WidgetSpan(
+                        child: Icon(
+                          Icons.close_rounded,
```
hjiangsu, line 120: If we're changing the icon for the action, this should also be updated to reflect that! (and any other instances)

hjiangsu (approved):
Just a minor comment! Otherwise LGTM

lib/community/utils/post_card_action_helpers.dart
micahmo: Ah yes those match the Lemmy UI. Perfect. Should be good to go now!

PR #1485, opened 2024-07-10
Show site content warnings

micahmo:
## Pull Request Description

This PR adds support for the new site content warning in 0.19.4.

See: LemmyNet/lemmy#4393

When adding an account or anonymous instance, we will check if the instance has a content warning. If so, we will show it to the user. If they decline, we will cancel adding the account.

If you look at the linked Lemmy PR, you'll see two other things mentioned.
* They say that you should not blur NSFW previews on instances that have a content warning. However, it doesn't look like the Lemmy web UI itself respects this recommendation, so for now I did not make any changes in Thunder. At the very least, I think we should add the "Blur NSFW content" setting to the account settings page to sync with Lemmy, and move our existing "Blur NSFW Previews" setting to the guest feed area. But that can be later.
* It also mentions that communities should have a new `only_followers_can_vote` setting, and that client should disable vote buttons if the setting is enabled and the user is not a subscriber. But I couldn't find a place to enable this setting on our 0.19.5 instance, so I skipped this as well.

## Issue Being Fixed

Issue Number: #1447

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/a175b1d7-297d-4a9a-9076-6a2dad3a358f

## Checklist

### DIFF
--- a/lib/account/pages/login_page.dart
+++ b/lib/account/pages/login_page.dart
@@ -7,12 +7,15 @@ import 'package:flutter/services.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:go_router/go_router.dart';
+import 'package:lemmy_api_client/v3.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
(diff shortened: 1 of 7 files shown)


ggichure (approved):

hjiangsu (approved):
LGTM!

Issue #1487, opened 2024-07-11
Setting for default feed view for instances


trymeouteh:
### Is Your Feature Request Related to a Problem?

When I change my account to an instance "anonymous view for lemmy.world" and go to the feed tab, it will have the feed under "Subscribed" which is an invalid option.

### Feature Description

Have a seperate setting for default feed for instances. Would be nice for the feed tab and search feed type filter.

The setting can be called "Default Feed Type For Instances" which can go under General->Default Feed And Sorting settings, right underneath "Default Feed Type" setting. The setting will have two options, "All" or "Local"

### Alternative Solutions

None

PR #1490, opened 2024-07-11
Can't make subscriptions default feed type in gest mode

micahmo:
## Pull Request Description

This PR removes the option to set "Subscribed" as the default feed type for guest mode, since that is not a valid option. I also added a migration in case anyone had that set before.

Note that, although we did allow "Subscribed" as a default option, it didn't seem to actually cause a problem. Having "Subscribed" selected from a user account also didn't cause a problem with switching to an anonymous instance.

## Issue Being Fixed

Issue Number: #1487

## Screenshots / Recordings

This is just a screen recording to show that I couldn't reproduce any issue with the way things are now (aside from potential confusion).

https://github.com/thunder-app/thunder/assets/7417301/26f4f5f3-7d9e-446a-af59-a4dd39af3196

## Checklist


hjiangsu (approved):
LGTM!

Issue #1425, opened 2024-06-04
Ability to set a default sort for different feed types


K4LCIFER:
### Is Your Feature Request Related to a Problem?

One can set a global default sort type for all feed types, but one currently cannot set a default sort type per feed. For example, set Active for All, and Scaled for Subscribed.

### Feature Description

It would be nice to be able to set the default sort type for each feed type.

micahmo:
I believe this request has some overlap with #1145 and #703!

Issue #1145, opened 2024-02-23
Save feed's "Sort Options" choice when switching feeds 


kellyaster:
### Is Your Feature Request Related to a Problem?

Right now in Settings > General, you can choose a setting for "Default Feed Sort Type", which assigns that preset to all feed views: _Subscriptions_, _Local Posts_,  _All Posts_, and all Communities. You can still change the sorting manually when you're actually viewing a feed, but the sorting style reverts to the preset if you leave and come back. This global one-setting-to-rule-them-all configuration is cumbersome because changing to another feed almost always means I have to tap on the "Sort Options" hamburger menu so it looks the way I want. I don't want every feed to be sorted the same way upon load. Sometimes you just want the memes community to be sorted by "New" whenever you load it, you know? 

The option to choose a default sort type is not a bad feature in and of itself; its persistence in spite of manual user changes to individual feeds is what makes it an inconvenience. 

### Feature Description

I propose persistent sorting choices across all feeds in Thunder. In an implementation such as this, the chosen "Default Feed Sort Type" will be honored in all feed views except for the ones in which the user makes a manual sorting change. 

**For example**: let's say this change to the codebase is implemented and that I have updated the app. I view the _Subscriptions_ feed, then I manually sort the list by "Active." Now this feed is exempt from the "Default Feed Sort Type" and will remember my manual sort change. From this point forward, whenever I visit _Subscriptions_, it is always sorted by "Active," even if I change the "Default Feed Sort Type."

micahmo:
Related to #703, specifically [this comment](https://github.com/thunder-app/thunder/issues/703#issuecomment-1741616889).

hjiangsu:
Thanks for the feedback! I think this is reasonable.

For devs:

I think we can introduce a new setting "Remember Feed Sort Type" (or a better name), and store this information in the app's `SharedPreferences`. We can optionally exclude this from the import/export settings.

I'm thinking of maybe a format `accountId_communityId: sortType` would work here since we'd want it to be on a per-account basis. If we also want to handle "Subscriptions/Local/All", then the format could change to `accountId_postListingType: sortType`

Issue #703, opened 2023-09-03
Unique default sort for Feed vs Community


Fmstrat:
**Is your feature request related to a problem? Please describe.**
If your default sort is something like `Top 12 Hours` things work great when viewing a feed like `All` or `Subscriptions`. But if you go to view a specific community that hasn't had a post in 12 hours, then it appears as though there are no posts at all. This is really common since many communities get about 1 or 2 posts a day.

**Describe the solution you'd like**
Instead of `Default Sort Type`, have a `Default Feed Sort Type` and a `Default Community Sort Type` so that the latter could be set to something like `New` or `Active`.

machinaeZER0:
I like this idea for similar reasons to you! One other approach would be the way Boost handles this sort of thing: in addition to the default sorting behavior, it has a toggle to "remember Sort by community" (and a sub-menu to further manage your saved Sorts). I like this idea because you can potentially get as granular as you like with your favorite communities, without altering your default feed sort whenever you're looking at something more specific. 

@hjiangsu @micahmo maybe one or both options would be worth considering in a future release?

PR #1492, opened 2024-07-12
Add custom sorting per community/feed

hjiangsu:
## Pull Request Description

This is a draft PR which implements the ability to save the sort type per-community/feed. A new database table was created to store the custom sorting, and a new option was added the Settings -> General -> Remember Feed Sort Type.

The precedence of sorting is as follows:
- If the user is not logged in, it will default to Thunder's sorting preference
- If the user is logged in, but Remember Feed Sort Type is disabled, then it will default to the user account's sorting preference
- If the user is logged in and Remember Feed Sort Type is enabled, then the default sorting will be the account's sorting preference unless it was previously changed. (e.g., when a user first visits Thunder's Lemmy community, it will be sorted based on their account's default sort type. If the user changes the sort type while in the Thunder's Lemmy community, any future visits to the community will use their last selected sort)

To do this, I had to move the logic of determining the sort type to the FeedBloc. FeedPage no longer has a parameter for the sort type, and the only way to change the sort type for a feed is to call the FeedChangeSortTypeEvent event.

This is still in draft because I want to do some additional testing for the database migrations to ensure it works as expected.

## Issue Being Fixed

Issue Number: #1425, #1145, #703

## Checklist

### DIFF
--- a/lib/account/models/custom_sort_type.dart
+++ b/lib/account/models/custom_sort_type.dart
@@ -1,0 +1,115 @@
+import 'package:flutter/foundation.dart';
+
+import 'package:drift/drift.dart';
+import 'package:lemmy_api_client/v3.dart';
+
+import 'package:thunder/core/database/database.dart';
+import 'package:thunder/core/database/type_converters.dart';
+import 'package:thunder/main.dart';
+
+class CustomSortType {
+  /// The type of sort (community/feed)
+  final SortType sortType;
+
+  /// The account id
+  final int accountId;
+
+  /// The community id
+  final int? communityId;
+
+  /// The feed type
+  final ListingType? feedType;
+
+  const CustomSortType({
+    required this.sortType,
+    required this.accountId,
+    this.communityId,
+    this.feedType,
+  });
+
+  CustomSortType copyWith({
+    SortType? sortType,
+    int? accountId,
+    int? communityId,
+    ListingType? feedType,
+  }) =>
+      CustomSortType(
+        sortType: sortType ?? this.sortType,
+        accountId: accountId ?? this.accountId,
+        communityId: communityId ?? this.communityId,
+        feedType: feedType ?? this.feedType,
+      );
+
+  /// Create or update a custom sort type in the db
+  static Future<CustomSortType?> upsertCustomSortType(CustomSortType customSortType) async {
+    try {
+      final existingCustomSortType = await (database.select(database.customSortType)
+            ..where((t) => t.accountId.equals(customSortType.accountId))
+            ..where((t) => customSortType.communityId == null ? t.communityId.isNull() : t.communityId.equals(customSortType.communityId!))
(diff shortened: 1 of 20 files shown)


micahmo:
Hey I know this is still a draft, but I wanted to leave a quick comment! You may want to consider storing the sorting info in the format `community@instance.tld` as the identifier, rather than using the community ID. That will allow us to remember the sort type for a given community regardless of what instance the user is on. I used a similar approach for user labels so they can work across instances. You can use these helper methods as inspiration, if you think it's a good approach. Up to you, of course!

https://github.com/thunder-app/thunder/blob/108605b8bd3473d1a078739454c1d513b17516a9/lib/account/models/user_label.dart#L106-L114

This is a super cool feature, by the way!!

hjiangsu:
> That will allow us to remember the sort type for a given community regardless of what instance the user is on.

Thanks for the suggestion! I originally made this feature to be account specific (e.g., each account can have it's own set of sort type for the same community/feed).

With your suggestion, were you suggesting that the sort type should be consistent across different accounts? (e.g., multiple accounts accessing the same community would get the same sort type) I'm not too sure which method is the best way to go about this, but if you have any thoughts or opinions, please let me know!

I'll pull in @K4LCIFER, @kellyaster, and @Fmstrat as well since they're the authors of the original feature requests.

micahmo:
> were you suggesting that the sort type should be consistent across different accounts? (e.g., multiple accounts accessing the same community would get the same sort type)

Yes, that's what I had in mind! But if you think it should be per community per account, that would work too!

Another thing we can do is see how other apps handle this feature.

Fmstrat:
In my opinion of 1, multiple accounts could be used 2 ways:
- Tracking different subscriptions
- Shit posting

In either situation, it's possible a user would want to view new posts in one subscription feed and top in another, so having a unique primary feed sort per qccount would be beneficial.

That being said, individual community feeds unique to each account probably won't matter much.

Assuming the PR allows for 'make my default community and (other) account views sort by "new"', and 'tailor my all and subscription feeds to sort by "top X"', it will cover most use cases regardless.

micahmo:
Thanks for the input @Fmstrat! I totally agree that being able to sort the main feed types (All, Local, Subscriptions) per account makes total sense, especially since different accounts are most likely on different instances.

In your example, it would be ok to have the same sort type for a community across accounts, right? I think that also makes sense to me. For example, the Thunder community doesn't have a ton of activity, so I always like to view it chronologically. I would like to make that change once and have it apply to all of my accounts.

> P.S. Boost remembers community sorts globally, but not feed sorts per account, so we would have an advantage there. (It also has a setting for whether or not to remember per-community sorts, and a screen to manage any saved ones.) Sync amazingly doesn't seem to remember sorting per community at all!

Again, this is also just my opinion as well and I'm hoping others will chime in!

Fmstrat:
I think that makes sense, yea.

PR #1495, opened 2024-07-12
Support custom thumbnails in posts

micahmo:
## Pull Request Description

This PR adds support for the Thumbnail URL field in the post creator. Note that, per the web UI, this option is not available when the main URL points to an image.

Notes
* The web UI shows the custom thumbnail field for posts with no URL (text-only), but it does not actually show the thumbnail. For now I've made Thunder match this behavior, but in the future we may wish to hide the custom thumbnail field when there is no URL (unless Lemmy decides to show the thumbnail for text posts).
* Thunder supports adding a custom thumbnail when initially posting, as well as adding one when editing.
* This requires a database change (for drafts) and I know there are two other PRs with database changes, so the schema and migration version numbers I have here will most likely have to change (depending on who goes first).

## Issue Being Fixed

Issue Number: #1447

## Screenshots / Recordings

https://github.com/user-attachments/assets/35dd9bcf-428c-4dd1-b1c8-61311258ce0e

## Checklist


hjiangsu (approved):
LGTM!

hjiangsu:
One other cool suggestion (that we can implement later) would be the ability to upload a custom thumbnail and use that rather than finding an existing link!

micahmo:
> One other cool suggestion (that we can implement later) would be the ability to upload a custom thumbnail and use that rather than finding an existing link!

Agreed! It should be pretty easy to add as well. The only reason I didn't was because the web UI doesn't (for whatever reason).

micahmo:
@hjiangsu Just FYI, since you may have been testing other changes with the same database migration number as this one, you may have problems starting Thunder after taking these changes. You might just need to execute some of the statements manually (regardless of the migration number) to get everything working. I know I did. 😊

PR #1496, opened 2024-07-14
Re-introduce comment navigator for new post page

hjiangsu:
## Pull Request Description

This PR adds back the comment navigator for the new experimental post page. The general logic remains similar to the current implementation, but with one behavioural change. Because the comments are now flattened, the comment navigator will navigate to each child comment (as opposed to the next parent comment). A separate PR will be made to allow navigation between top-level comments using this navigation.

With this, I've also added back the ability to scroll to the last comment (somewhat). There are some quirks here due to the transition to slivers. Slivers don't have a RenderBox associated with them, so fetching the height of the sliver is less trivial (from my testing). I do think I have it at a good point, but will need to do more testing on it to be sure.

More specifically, this PR fixes the following:
> - Bring back comment navigation
> - It should be possible to scroll the last comment to the top of the screen.

## Issue Being Fixed

Issue Number: #1413

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -1,3 +1,5 @@
+import 'dart:math';
+
 import 'package:flutter/material.dart';
 
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -15,6 +17,7 @@ import 'package:thunder/post/bloc/post_bloc.dart';
 import 'package:thunder/post/utils/comment_action_helpers.dart';
 import 'package:thunder/post/widgets/post_page_app_bar.dart';
 import 'package:thunder/post/widgets/post_view.dart';
+import 'package:thunder/shared/comment_navigator_fab.dart';
 import 'package:thunder/shared/cross_posts.dart';
 import 'package:thunder/shared/text/scalable_text.dart';
 import 'package:thunder/shared/text/selectable_text_modal.dart';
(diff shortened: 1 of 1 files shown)


micahmo:
Awesome, I will try daily driving this for a little while!

micahmo (approved):
The code looks good!

So far from my testing it's working fine! Like you said, it will require another change to only navigate between top-level comments.

As far as scrolling to the bottom, it seems to be working fine, although it allows the "you've reached the bottom" message to be scrolled to the top, rather than the last comment. I'm not sure if that was intentional, and it might not be an easy fix (since each comment is a separate widget, you'd have to add keys to each of the comments under the last top level comment and combine all their heights), but either way I think it's fine!

hjiangsu:
> As far as scrolling to the bottom, it seems to be working fine, although it allows the "you've reached the bottom" message to be scrolled to the top, rather than the last comment. I'm not sure if that was intentional, and it might not be an easy fix (since each comment is a separate widget, you'd have to add keys to each of the comments under the last top level comment and combine all their heights), but either way I think it's fine!

I left it there somewhat intentionally for the reasons that you mentioned (it would be difficult to calculate the full extent of the last comment, especially if it has children). We can leave it as is for now and adjust it when needed 😁

micahmo:
@hjiangsu I've noticed a small bug from this change which is that, when you press down for the very first time or any time you are scrolled above the top comment, it always goes to the second comment rather than the first.

I am going to add a note to #1413 as well.

https://github.com/user-attachments/assets/a87a9eb6-87e0-4eab-a15b-f9b09b20fdac

Issue #1309, opened 2024-04-16
Proposal: Make Post/Reply Button More Reachable


machinaeZER0:
### Is Your Feature Request Related to a Problem?

I love our post/comment/reply functionality (even moreso with some of the incoming updates to displaying quoted text), but one place I think we could improve is shifting the send/post icon (the little paper airplane/arrow) lower on the screen. Obviously our app is styled differently, but I think the way that most text messaging apps have their send button closer to the bottom makes for a more positive ergonomic experience.

### Feature Description

Since we have such a robust formatting bar for text entry, I don't know that there's a natural place to add a send icon as the app is currently laid out. That said, one idea would be to take the preview icon (the eyeball) and make that the send button, where a long press would give the option to preview instead? This feels rather elegant to me, especially as a person who doesn't often use the preview button, but sees both actions as similar enough that grouping them in a high-profile button "feels good." 

I'd be interested to hear other people's opinions on this - might be a non-issue, but figured it was worth at least bringing up :)

### Additional Context

If we did go the proposed route in the future, just noting that we'd want to make sure that the updated button made it obvious that long-pressing would provide additional options!

machinaeZER0:
@micahmo it sounded like you may have actually mocked up something along these lines at some point?

micahmo:
I think this issue has some overlap with #746 and #788. I will try to dig up my mockup and see if it's still useful! 😆

micahmo:
Ok I found my mockup! My idea was to address the concern of @notBradPitt in [this comment](https://github.com/thunder-app/thunder/pull/455#issuecomment-1648899943) where it was rightly mentioned that, if the button were moved lower, it would be very easy to hit by accident. (And I think that's especially true now that people have muscle memory associated with the preview button down there.) My idea was to show the send button at the bottom only once you've entered preview mode. The benefit is that you won't hit send by accident, but the downside is that you might not always want to preview first, which both @machinaeZER0 and @noodlejetski mentioned.

Anyway, here's the demo. Let me know if this is worth considering at all, or if we should start over. 😆

https://github.com/thunder-app/thunder/assets/7417301/0934147b-ea3d-426c-9aa9-c82d250ac376

noodlejetski:
another downside is that due to the keyboard collapsing on preview, the buttons move around which makes them more difficult to hit. sending a comment that way requires making to very deliberate presses, which makes you hunt for the button twice.

machinaeZER0:
[...]

Issue #788, opened 2023-09-30
Consider a setting for changing Send and Preview buttons' positions


noodlejetski:
**Is your feature request related to a problem? Please describe.**
when writing a comment, I keep hitting the Preview button thinking it's going to publish it, and I think the reason is twofold:
1. as an Android user, I'm used to the Send button being right above my keyboard, near the text field - exactly where the Preview button is currently located; the current layout is more akin to some iOS apps, as far as I know.
2. the preview button is much more prominent with the current design, which also makes it look like a Send button in other apps.

**Describe the solution you'd like**
I'd love an option to move the Send button to where the Preview button currently is, maybe with the Preview button placed next to it (and made less vibrant than Send). 

**Additional context**

compare the button design and placement in Thunder to apps like Moshidon (a Fediverse microblogging client) or Signal (an instant messenger):

![ResizedImage_2023-09-30_10-01-29_8359](https://github.com/thunder-app/thunder/assets/8598559/4a31df97-cfb6-47ff-aa51-86caa2991b66)

![ResizedImage_2023-09-30_10-07-56_2527](https://github.com/thunder-app/thunder/assets/8598559/832d554b-f631-4452-8a70-562d141d10b0)

![ResizedImage_2023-09-30_10-05-58_1337](https://github.com/thunder-app/thunder/assets/8598559/a081e77d-b8cd-412f-8318-39034a8821b8)

micahmo:
Hey @noodlejetski, thanks for the suggestion! I agree, I also expect the Send button to be right above the keyboard. In the PR that made this change, there was [one comment in favor of the placement](https://github.com/thunder-app/thunder/pull/455#issuecomment-1648899943) (as well as the original author, of course), but I'm sure we'd be open to discuss this again.

P.S. Would the suggestion #746 help?

noodlejetski:
> Would the suggestion #746 help?

I don't feel so hot about that solution, as it hides the _much_ more used action behind a lesser used one. I'd wager that people who open the reply window use the "send action" virtually every time (other than when they change their mind about replying or if they'd open the window by accident), while they preview their comment's format, what, every 10th time? every 20th?

Issue #746, opened 2023-09-17
Better accessibility for the post button


Tealk:
**Is your feature request related to a problem? Please describe.**
Submitting a comment is not possible with one hand

**Describe the solution you'd like**
By pressing the preview button for a long time, you can offer the possibility to send the preview.

Issue #1461, opened 2024-06-21
Please add Alt text when posting photos


Coehill:
### Feature Description

Lemmy now support Alt Text and you can Add it to photo posts from the Jerboa app. Can we please add Alt text to Thunder to make it accessible to those with screen readers / visual impairments?

hjiangsu:
> Add the ability for the screen reader to read the alt text when focused on a photo.

This might already be integrated with the markdown renderer (but will have to double check)!

Another thing to do could be to add the alt text to the image preview as well

micahmo:
Is this an 0.19.4 issue? If so, I can track it in #1447.

hjiangsu:
I do believe this was introduced in 0.19.4!

https://github.com/thunder-app/lemmy_api_client/blob/8d1513a3574665b464822595ac97bb166a5ffa89/lib/src/v3/models/post/post.dart#L35

PR #1498, opened 2024-07-15
Proposal for moving the create post button

micahmo:
## Pull Request Description

This PR aims to satisfy issues #746, #788, and #1309 by adding a "Create Post" button beneath the post body text and just above the keyboard. I know this has been a hotly debated topic, so I want to bring in the three creators of the three issues I linked, @Tealk, @noodlejetski, and @machinaeZER0 for any additional input!

The compromise is to keep both the preview and create buttons in the markdown actions areas. The size of the buttons has been reduced to not consume quite as much horizontal real estate. In addition, the old create button at the top has been left in place. Finally, the create button will not even appear until all of the validation has succeeded (otherwise it's just wasting space).

Please please provide feedback on this change!! I want to make sure it's something that makes the most amount of people happy, even if there is some compromise. Clearly there is a desire for *something* to change in this regard (hence the three separate issues), so I hope this is a step in the right direction!

https://github.com/user-attachments/assets/83259adf-cce0-4c2e-a878-144e675b7446


machinaeZER0:
Whoa, that's super cool! I like that implementation a lot!

The icon at the top feels a bit duplicative at that point - would we be choosing one or the other, or would the plan be to keep both send buttons?

Thanks so much for making this proof of concept! Interested to see what the others think.

Tealk:
I like it too <3

micahmo:
Thanks for the feedback everyone!

> The icon at the top feels a bit duplicative at that point - would we be choosing one or the other, or would the plan be to keep both send buttons?

I had initially kept it for two reasons.
* For anyone that had gotten used to pressing it at the top, it would still be there for their muscle memory.
* Until the required fields are filled out, the user might be confused that there is no send button. By keeping the old one up there but disabled (grayed out) until the post is ready, it gives some indication to the user that there more to do.

However, as I've been thinking about this some more, I thought a better approach might be to give a clear indication about which fields are required (which is really just community and title). If we have that, then it should be more clear to the user that they have more to do before the create button will be available. Thoughts on this?

I also made one more tweak to make the text fields look a little more "material"-y, at least to me. Please provide feedback on this latest demo!

https://github.com/user-attachments/assets/6fc32cf1-ae9c-4ec5-ae0f-4ec0a06c5125

machinaeZER0:
I like the new version! I do prefer the cleaner/simpler styling of the title/url fields in the previous video, but if that better matches the rest of the app experience then leaning towards consistency makes perfect sense to me!

hjiangsu:
Thanks for working on this @micahmo! I have a couple of possible suggestions and tweaks for the proposed changes. Let me know what you (and others) think about this!

For the fields marked as required, we can follow something similar to what the [material 3 guidelines show](https://m3.material.io/components/text-fields/guidelines#9ad90554-a793-4506-9075-6812fd7b381a):
- We can put the "required" text as a hint at the bottom of the text field
- We can use the error colour for any errors after submission/validation (similar to what we see in the guidelines)

Aside from that, my personal preference would be to keep the create button shown at all times, rather than only showing it when we fill in the required fields. My reasoning for this is that it _might_ reduce the potential for accidentally creating a post.

By keeping the buttons "static", we reduce the chances where a user accidentally creates a post when they meant to preview it instead. It reduces the user's cognitive load because they don't have to double check to see if the action they're tapping on is the preview or the create action.

Another minor suggestion (feel free to use/disregard this suggestion) would be to make the shape of the send icon different than the shape of the preview action (e.g., make the create action more elongated than the preview action). This does sacrifice horizontal real estate, but does make it easier to distinguish the actions aside from colour.

micahmo:
@hjiangsu Thanks for all of the suggestions, I will implement these! How do you feel about the styling of the text fields? Personally I feel like the outlines match material recommendations (like all the screenshots in the link you sent) and match the rest of our app, as well as many other apps, but I also understand that more content on screen makes things more cluttered as @machinaeZER0 said!

<details>
<summary>Expand for screenshots of other apps</summary>

| Sync |
|--------|
| image | 

| LiftOff |
|--------|
| image | 

| Summit |
|--------|
| image | 

| Eternity |
|--------|
| image | 

| Voyager |
|--------|
| image | 

| Jerboa |
|--------|
| image | 

</details>

micahmo:
@hjiangsu Alright, I believe this includes all of your suggested changes. Let me know what you think, and of course everyone else is more than welcome to chime in again as well!

https://github.com/user-attachments/assets/f66b1310-bc6e-432f-9e06-d7cadab8906d

micahmo:
Here's a new screenshot with the new custom thumbnail field!

| image |
| - |

hjiangsu:
Thanks for applying the suggestions! I would slightly tweak the width of the create button to be a bit more elongated (maybe 60?)

> How do you feel about the styling of the text fields?

Hmm, I'm half-half on this. The outlined border does make it more aligned with MD3 guidelines, but it does take up a lot more vertical space which arguably leaves less area for the body (which is the most important section in my opinion)

Could we play around with the spacing to see if we can make the text fields a bit more compact?

micahmo:
Alrighty, I increased the width of the create button to `60`, decreased the space between the form fields, and increased the density by lowering the `contentPadding` of the fields. Let me know what you think!

| Before | After |
|--------|--------|
| image | image |

machinaeZER0:
Would it make sense to also shrink the actual text in the field boxes slightly? I have a feeling that may be drawing from the user's font size so maybe not possible/desirable, but shrinking a little would make the smaller boxes look a bit less cramped, and also decrease the size disparity between the fields and the other text on the screen. Either way, I do think I like the more compact look!

hjiangsu:
I like the changes! I do agree here with @machinaeZER0 that it does feel a bit more cramped with the reduced padding. I would either try to shrink the text size as @machinaeZER0 mentioned, or increase the padding a bit more on the text fields.

micahmo:
Alrighty, I figured rather than decreasing the font size, which is standardized throughout the app and should be using a text scaler (not sure if the create post page is currently respecting that or not, but the point is that I didn't want to hard-code a font size) I but the back to halfway between the original and what I had changed it to. Thoughts now?

BTW, we will probably have to come up with a better long-term solution as more fields get added (like [alt text](https://github.com/thunder-app/thunder/issues/1461)). Maybe a way to expand/collapse optional fields.

| 1 | 2 | 3 |
|--------|--------|--------|
| image | image | image |

Tealk:
I like  2

noodlejetski:
just wanted to chime in that I like this proposed design a lot (especially after the subsequent adjustments), and I'm looking forward to seeing it in a future release!

hjiangsu:
I think #3 looks great!

> BTW, we will probably have to come up with a better long-term solution as more fields get added (like https://github.com/thunder-app/thunder/issues/1461). Maybe a way to expand/collapse optional fields.

Agreed - I think it makes sense to make non-essential fields expandable, or part of an "advanced" section to reduce clutter on the page. Obviously, this can be delayed until later!

hjiangsu (approved):
The changes LGTM! Let me know if this is ready to go 😁

micahmo:
Yes, the code matches screenshot 3 so we should be good!

Issue #1423, opened 2024-06-04
Add some more UI feedback when subscribing to a community


K4LCIFER:
### Is Your Feature Request Related to a Problem?

Currently, when you subscribe to a community by tapping on the plus in the circle at the top of a community page, it just shows that you tapped it by highlighting, but doesn't provide any further feedback until the subscription goes through. The main issue with this is that, depending on the server, this could potentially take a while, and the button doesn't change to the subscribed state until the subscription has gone through.

### Feature Description

I propose adding some feedback to show that the request for subscription was successful — the button could change to some intermediary state to show that the request was made, and then change to a subscribed state when the subscription has been successfully completed.

micahmo:
Hi @K4LCIFER! When you subscribe to a community that is not on your instance, the button should change from the plus icon to three dots, which indicates that your subscription is pending.

![Screenshot_20240604-201007.png](https://github.com/thunder-app/thunder/assets/7417301/cf1c7e5d-2fe8-4c91-95c5-daeb8e8260fb)

In addition, if you long-press the button, it will tell you that pressing the button again will unsubscribe you, and that your subscription is currently pending.

Can you let us know if the three-dot icon is showing for you and/or if you have any other ideas for more clearly indicating what's going on?

hjiangsu:
Maybe we can add in a snackbar message here when the action succeeds or fails?

K4LCIFER:
> When you subscribe to a community that is not on your instance, the button should change from the plus icon to three dots, which indicates that your subscription is pending.

I do see the 3 dots, but there is a delay that is dependent on the responsiveness of the server. Sometimes, this delay can be quite substantial — substantial enough that the user thinks that their request didn't go through, so they tap the button again.

---

> Maybe we can add in a snackbar message here when the action succeeds or fails?

Perhaps also a message, on first tap, that says something like "subscription request sent"?

PR #1499, opened 2024-07-16
Improved feedback when subscribing/unsubscribing to community

hjiangsu:
## Pull Request Description

This is a super small PR which adds snackbar messages for subscribing and unsubscribing from communities.

## Issue Being Fixed

Issue Number: #1423

## Screenshots / Recordings

https://github.com/user-attachments/assets/07945dea-55b4-4d18-a324-e711ec6b01e4

## Checklist

### DIFF
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -70,6 +70,14 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
         try {
           CommunityView communityView = await followCommunity(event.communityId, event.value);
 
+          if (GlobalContext.context.mounted) {
+            if (event.value) {
+              showSnackbar(AppLocalizations.of(GlobalContext.context)!.subscriptionRequestSent);
+            } else {
+              showSnackbar(AppLocalizations.of(GlobalContext.context)!.unsubscriptionRequestSent);
+            }
+          }
+
           emit(state.copyWith(status: CommunityStatus.success, communityView: communityView));
           emit(state.copyWith(status: CommunityStatus.fetching));
 
@@ -87,6 +95,7 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
             }
           });
         } catch (e) {
+          showSnackbar(AppLocalizations.of(GlobalContext.context)!.failedToPerformAction);
           return emit(state.copyWith(status: CommunityStatus.failure));
         }
         break;
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -68,17 +68,33 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
         break;
       case CommunityAction.follow:
         try {
-          CommunityView communityView = await followCommunity(event.communityId, event.value);
+          // Determines the desired subscribed type outcome based on the value
+          // If [event.value] is true, then the desired outcome is to subscribe. If [event.value] is false, then the desired outcome is to unsubscribe
+          SubscribedType? subscribedType = switch (event.value) {
+            true => SubscribedType.subscribed,
+            false => SubscribedType.notSubscribed,
+            _ => null,
+          };
 
-          if (GlobalContext.context.mounted) {
-            if (event.value) {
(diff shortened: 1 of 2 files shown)


lib/community/bloc/community_bloc.dart
```diff
@@ -70,6 +70,14 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
         try {
           CommunityView communityView = await followCommunity(event.communityId, event.value);
 
+          if (GlobalContext.context.mounted) {
+            if (event.value) {
+              showSnackbar(AppLocalizations.of(GlobalContext.context)!.subscriptionRequestSent);
+            } else {
+              showSnackbar(AppLocalizations.of(GlobalContext.context)!.unsubscriptionRequestSent);
```
micahmo, line 77: I'm not sure we need this one since I don't believe unsubscripions have the same delay as subscriptions where the can be in `pending` for a while. But I'll leave that up to you!

micahmo (approved):
LGTM

micahmo:
What happens if you test this on an instance that isn't yours? Do you get the "subscription request sent" message only?

lib/community/bloc/community_bloc.dart
hjiangsu: I'm not actually too sure how the pending states work in this situation, so I just applied it to both! I can remove the unsubscription one though since you may be correct here!

If you know of a way to trigger the pending state, let me know and I can test it out

micahmo:
Here's my understanding of how things work.

* On your own instance, requesting to subscribe always returns that you are immediately subscribed (so we shouldn't need the "sent request" message here).
* On _any_ instance, requesting to unsubscribe always returns that you are immediately unsubscribed (so we shouldn't need the "sent request" message here either).
* However, when subscribing to a community on another instance, it always returns pending. In the past, we had put in a trick ([here](https://github.com/thunder-app/thunder/blob/c3c7ceb775563a27d2442befb2d8bfeafce8416c/lib/community/bloc/community_bloc.dart#L76-L78)) to do another API request, with the idea that _most_ of the time, if we checked again after a second, the status would be updated from pending to subscribed. Although more recently I've noticed that this isn't always the case and I have to manually refresh after a significant amount of time before it returns that I am subscribed.

I think what I would recommend is...
1. Check the return value from `followCommunity`. If it matches what we expect for the given event, then just tell the user immediately they they're subscribed or unsubscribed.
2. If it doesn't match what we expect for the event, then show the "request sent" message. And only then, do the one-second delay (or maybe longer) and check again. If it the result matches after the second check, show the final message to the user. If it still doesn't match, I think we do nothing and let the user manually refresh or check later.

What do you think?

hjiangsu:
Thanks for the response! I think I did pretty much what you suggested but with a slight change (I didn't see your message before 😅)

The logic goes like this:
- We'll always send the "subscription requested" message when we attempt to subscribe to a community regardless of the instance. This happens before `followCommunity` is triggered, because we don't know how long that API call will take (depending on network conditions)
- From the result of `followCommunity`, we'll return back the subscribed/unsubscribed message immediately if it matches what we are expecting from `event.value`
- If it doesn't match (e.g., its in the pending state or it hasn't properly updated), then there's a delay of 1 second before we attempt to fetch again. If it works the second try, then display the subscribed/unsubscribed message to the user. Otherwise, we don't display anything

Here's some demo videos to show the differences:

Subscribing to local community (very slow connection):

https://github.com/user-attachments/assets/1ad7714c-073f-4dbf-9d19-a0ca82ee60a8

Subscribing to local community (fast connection):

https://github.com/user-attachments/assets/005a1611-acb4-445a-a943-51a45fc81a1e

Subscribing to federated community:

https://github.com/user-attachments/assets/c656e745-f753-45fd-a9c9-17af3f942f6f

Unsubscribing from federated community:

https://github.com/user-attachments/assets/fcc7593e-1840-420c-aa34-c31e459b8100

Unsubscribing from local community (slow connection):

https://github.com/user-attachments/assets/01782120-56de-476c-97ef-cb4fe9772566

micahmo (approved):
Haha we were on the same wavelength. This looks perfect, thank you!!

PR #1500, opened 2024-07-17
Migrate anonymous instances to database

micahmo:
## Pull Request Description

This PR makes the following database changes.
* Add index column to accounts table.
* Migrate anonymous instances to accounts table.

There should be no impact on the UI at this point, so please let me know if anything looks or acts funky after this migration!

Related to #1494.

### DIFF
--- a/lib/account/models/account.dart
+++ b/lib/account/models/account.dart
@@ -10,48 +10,121 @@ class Account {
   final String? username;
   final String? displayName;
   final String? jwt;
-  final String? instance;
+  final String instance;
+  final bool anonymous;
   final int? userId;
+  final int index;
 
   const Account({
     required this.id,
     this.username,
     this.displayName,
     this.jwt,
-    this.instance,
+    this.anonymous = false,
+    required this.instance,
     this.userId,
+    required this.index,
   });
 
-  Account copyWith({String? id}) => Account(
+  Account copyWith({String? id, int? index}) => Account(
         id: id ?? this.id,
         username: username,
         jwt: jwt,
         instance: instance,
         userId: userId,
+        index: index ?? this.index,
       );
 
   String get actorId => 'https://$instance/u/$username';
 
   static Future<Account?> insertAccount(Account account) async {
     // If we are given a brand new account to insert with an existing id, something is wrong.
-    assert(account.id.isEmpty);
+    assert(account.id.isEmpty && account.index == -1 && !account.anonymous);
 
     try {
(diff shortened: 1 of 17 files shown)


hjiangsu (approved):
Just did a quick skim of the changes - LGTM!

I'll do some testing on my end as well to check for any possible issues with this change. Thanks again for doing this! 😄

hjiangsu:
Hmm, I just tried to go from a clean install of 0.4.0 to the current changes, and the accounts are not appearing in the profile switcher anymore (both normal accounts and guest accounts)

I am seeing this error on the logs which might be the issue:

```
flutter: SqliteException(1): while executing, duplicate column name: custom_thumbnail, SQL logic error (code 1)
flutter:   Causing statement: ALTER TABLE drafts ADD COLUMN custom_thumbnail TEXT, parameters:
```

I think we might have to do some checks to see if the columns exist before adding/removing them in the migration code (or add try catch blocks to them)

hjiangsu:
I'll try to make some changes to the migration code to fix some of the issues

micahmo:
> duplicate column name

Are you sure it was really a clean install? I'm not sure how we could get `duplicate column name` unless maybe you already had that column in your database from some previous testing. Let me know what you find!

hjiangsu:
> Are you sure it was really a clean install? I'm not sure how we could get duplicate column name unless maybe you already had that column in your database from some previous testing.

This is what I did:
- Uninstalled Thunder from my device (this should remove all data associated with the app)
- Installed Thunder from the latest nightly build
- Added some accounts and anonymous instances
- Upgraded to latest version

I think this is what happened:
- If I run the migration from 0.4.0 up to the current version, the Drafts database will already contain `custom_thumbnail` when its created in `await migrator.createTable(drafts)`, because its present in the current table definition
- However, since this condition `from <= 3 && to > 3` is also satisfied, the migration tries to add the column when it already exists

If I go from a version where the Drafts table is created (e.g., version 2), but does not contain `custom_thumbnail`, then I _believe_ no error is thrown. Hopefully this makes sense!

I'm rewriting the migration logic right now so that it does migration step-by-step instead to hopefully avoid these types of situations: https://drift.simonbinder.eu/docs/migrations/step_by_step/#generating-step-by-step-code

micahmo:
Ahh yes, you are totally right about what happened! Good catch. And I like the `stepByStep` migration strategy. That should make things much easier. Thanks!

PR #1501, opened 2024-07-17
Fix database migration logic

hjiangsu:
## Pull Request Description

This PR attempts to fix some of the database migration logic going from 0.4.0 -> 0.5.0. The main changes are listed here:
- Exported the schema for each of the database version upgrades. This is used to generate migration steps to help ease migration logic. See https://drift.simonbinder.eu/docs/migrations/step_by_step/#generating-step-by-step-code for more details
  - To do this, I went back to the commits that introduced new schema versions, and generated the exported schema as outlined in https://drift.simonbinder.eu/docs/migrations/exports/#exporting-the-schema
- Updated the migration logic to use the step-by-step migrations from before. Note: I was unable to find official documentation on how to perform database downgrades, so that is missing from this PR. If there is an easy way to do this, please let me know!
- I set a default value for the `listIndex` in the Accounts table. This is to allow the migration to happen successfully (otherwise it throws an error that the column has a null value when creating the column)

I only tested this migration from 0.4.0 -> develop, and from the current nightly -> develop to check for any database issues.

## Issue Being Fixed

Issue Number: #1500 (related)

## Checklist


micahmo (approved):
Looks great, thanks so much for fixing this!

PR #1504, opened 2024-07-18
Fix issue with notifications staying page after marked as read

micahmo:
## Pull Request Description

This PR fixes an issue where read comments were not hidden from the notifications page. See my original comment here: https://github.com/thunder-app/thunder/pull/1463#issuecomment-2217485128

I think this might have been broken due to flipping the logic from `showAll` to `showUnreadOnly`, which would have reversed the logic of the notifications page (which is not creating an initial bloc event, so it mostly relies on the defaults).


ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1506, opened 2024-07-24
Bump intl and screenshot versions

gwbischof:
`intl 0.18.1` conflicts with lastest flutter version on stable channel.
this PR updates the intl version.
`flutter pub get` updated the pubspec.lock

```
Note: intl is pinned to version 0.19.0 by flutter_localizations from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.

Because every version of flutter_localizations from sdk depends on intl 0.19.0 and thunder depends on intl ^0.18.1, flutter_localizations from sdk is forbidden.
So, because thunder depends on flutter_localizations from sdk, version solving failed.
```

I also updated the flutter version used by the CI.  So the CI is expected to fail for this PR.

I also had to update the `screenshot` version to get this running on a `iPhone 15 Pro Max emulator`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,12 +15,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:  
+        with:
           submodules: 'recursive'
 
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.19.x'
+          flutter-version: '3.22.x'
           channel: "stable"
           cache: true
           cache-key: 'flutter-:os:-:channel:-:version:-:arch:-:hash:' # optional, change this to force refresh cache
@@ -29,7 +29,7 @@ jobs:
 
       - name: Dependencies
         run: flutter pub get
-      
+
       - name: Install arb_utils to check sorting of translations
         run: dart pub global activate arb_utils 0.6.0 # temporarily force version 0.6.0 to fix build errors
 
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -986,10 +986,10 @@ packages:
     dependency: "direct main"
     description:
       name: intl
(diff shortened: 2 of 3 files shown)


hjiangsu:
Hey! I think this might be related to this PR: https://github.com/thunder-app/thunder/pull/1372

Upgrading to Flutter 3.22 is currently on hold because of some issues with our dependencies (the main blocker being `flex_color_scheme`).

gwbischof:
I'm not getting any build problems for flutter 3.22 with the changes on this PR.  I have only built for iphone so far.  Are the problems only for android?

hjiangsu:
> I'm not getting any build problems for flutter 3.22 with the changes on this PR. I have only built for iphone so far. Are the problems only for android? Maybe we merge this one which fixes the 3.22 iphone build, and then I can help work out getting the android build working?

The main issue here is that `flex_color_scheme` is not yet fully compatible with 3.22, so some of the widgets and colours will be off on iOS/Android devices for those who are using accent colours. For example (pulled from this branch):

| 3.22 | 3.19.6 |
|-|-|
|simulator_screenshot_42685D61-F7BA-44AC-B69F-8F0D9FA99697| simulator_screenshot_23EC1CF0-3097-42CF-8E29-8C07471AD0D4|
|simulator_screenshot_68C32722-F717-42E0-9607-60BA94E190B7| simulator_screenshot_AA9A253D-B14B-4F1B-A654-DC3356F5E773 |


If we merge this PR before the fixes to `flex_color_scheme`, then it forces everyone to upgrade to 3.22 for development (since some of the updated dependencies require >=3.22)

gwbischof:
Got it, thanks for the explanation. I wonder if it `flex_color_scheme` could be fixed by adding the extra line described here.
https://docs.flutter.dev/release/breaking-changes/new-color-scheme-roles#migration-guide.

hjiangsu:
I believe there's much more under the hood that has to be changed. See this discussion for a bit more detail: https://github.com/rydmike/flex_color_scheme/discussions/232#discussioncomment-9484785

PR #1509, opened 2024-08-01
When navigating to non-Lemmy instance, allow opening in browser

micahmo:
## Pull Request Description

This is a super small change which allows the user to open a browser when they try to navigate to an instance when it is not really a Lemmy instance.

https://github.com/user-attachments/assets/891a1077-50a1-414f-8625-8c17893c56b9


hjiangsu (approved):
LGTM!

PR #1511, opened 2024-08-01
Allow navigating to user instance from post sheet

micahmo:
## Pull Request Description

Currently when you long-press a post, you can perform actions on the instance of the post/community, but not the user. (You can perform actions related to the instance of the user from a comment.) This PR improves the instance sub-menu in the post bottom sheet to provide actions related to both the post/community and the user. If they are on the same instance, only one set of actions is shown.

I also made a small tweak to only show the post metadata chips (i.e., the post language) on the main page of the sheet.

https://github.com/user-attachments/assets/b948239e-e03c-4fc1-b790-b6f8ecd84208


hjiangsu (approved):
LGTM!

PR #1515, opened 2024-08-06
Switch hint to label in create post page

micahmo:
## Pull Request Description

This is a super small PR which swaps hint text for label text on the create post page. I realized that you can't tell which field is which once you have entered your own text. And the terms we were using weren't really hints anyway, so I thought it made sense to switch all of these to labels.


ggichure (approved):

hjiangsu (approved):
LGTM!

PR #1516, opened 2024-08-07
Support alternate link sources

micahmo:
This PR was inspired by the following post: https://lemmy.world/post/18198313

In that post a screenshot of the Tesseract front-end was shown, demonstrating how it provides links to "Alternate Sources" for a given article link. The sources that Tesseract provides are these:

* Archive Today (https://archive.today)
* Ghost Archive (https://ghostarchive.org)
* 12ft.io (https://12ft.io)
* Ground News (https://ground.news)

Notes
* The first three sources provide archives of the given link (often to bypass paywalls or to see an article after it's been taken down). The last one finds similar articles from other media outlets so that folks can compare the same story from different perspectives.
* These are all auto-generated links. There are no API calls to any of these sites (e.g., to request an archive), so the user may navigate to one of these sites and find that there is no archive of the given page. We can't make any guarantees; we can just provide the link.
* We can't make a determination about whether a given link is really a news article or just a link to a web page. As a result, these alternate links may not always be completely appropriate (but they shouldn't cause any harm).
* Tesseract has a similar feature that provides links to alternate YouTube front-ends for YouTube videos, such as Piped and Invidious. It may even have other site-specific alternatives that I just didn't run across. I think these would all be cool to implement, but they can be done later on a site-by-site basis.
* Although I had to significantly change how the link bottom sheet is being built (from a `BottomSheetListPicker` to a manually-constructed list of `PickerItem`s) I did a back-and-forth comparison to ensure that the layout looks exactly the same as before.

I have decided to implement this feature with the following sources.
* Archive Today (https://archive.today) -- very popular archiving tool these days
* Internet Archive (https://archive.org/) -- the original archiving tool
* Ground News (https://ground.news)

I excluded the following sources.
* Ghost Archive, because we have two other well known archive sites. However, if users are strongly in favor of Ghost Archive, of course we can add it back.
* 12ft.io, because it is known to not work as well these days and not be totally reliable

The good news is that new sources are super easy to add (one line of code) so we can be totally flexible there.

The alternate sources can be accessed from the following places.
* Link long-press bottom sheet
* Post page (old and new) overflow menu
* In-app browser overflow menu

In all cases, the alternative links are displayed using the same bottom sheet widget, to reduce duplicate code

https://github.com/user-attachments/assets/40504e86-5690-4c4e-97eb-b1b8a9d0f611

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -101,6 +101,10 @@
   },
   "alreadyPostedTo": "Already posted to",
   "@alreadyPostedTo": {},
+  "alternativeSources": "Alternative Sources",
+  "@alternativeSources": {
+    "description": "Action for viewing a link from alternative sources"
+  },
   "always": "Always",
   "@always": {},
   "andXMore": "and {count} more",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -101,9 +101,9 @@
   },
   "alreadyPostedTo": "Already posted to",
   "@alreadyPostedTo": {},
-  "alternativeSources": "Alternative Sources",
-  "@alternativeSources": {
-    "description": "Action for viewing a link from alternative sources"
+  "alternateSources": "Alternate Sources",
+  "@alternateSources": {
+    "description": "Action for viewing a link from alternate sources"
   },
   "always": "Always",
   "@always": {},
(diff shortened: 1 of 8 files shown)


micahmo:
I pushed a small change which ensures that navigation to alternate sources once you're already in a webview occurs within that same webview instead of pushing a new one on top.

realeatham:
This is cool

hjiangsu (approved):
LGTM! Thanks for adding this feature 😄

Issue #1113, opened 2024-02-03
Keep post view state after reopening it


daviessm:
**Is your feature request related to a problem? Please describe.**
I have a habit of closing a post's comments and then deciding I'd like to view them again. Currently this re-loads the post when I select it again.

**Describe the solution you'd like**
Hide a post rather than closing it when returning to the post list. If the user selects to view the same post again, re-show the post. If they select a different one, discard it and load the new one.

micahmo:
Is this a dup of #742?

daviessm:
Yep! I also came from Relay.

Issue #742, opened 2023-09-15
Remember the state of comments when navigating out and back in


micahmo:
**Is your feature request related to a problem? Please describe.**
Currently, every time we navigate into a comment section, it is reloaded fresh.

**Describe the solution you'd like**
It would be nice if we could remember the state of the comments for a given post so that we can navigate out and back in without losing our spot. This would facilitate the following workflow (as an example).

* I'm reading through the comments
* I want to see the post's title/image as a reminder, so I navigate out to the main feed
* I navigate back into the comments and want to pick up where I left off

Note that to avoid things getting stale, we can do this for only the most recently visited post. If I go to the comments for post A, then B, then A again, it's ok to reload.

> For those familiar with Relay, this is how things worked there.

Glazed:
It's probably less work to provide some UI to let you peek at the post from anywhere in the comments.

micahmo:
> It's probably less work

Fortunately, the work to implement this feature has already been done! 😆 See #1520.

And yeah, like you said, there are any number of reasons to have this beyond just wanting to get a peek at the post again. Maybe you go back by accident, maybe you go back intentionally but then change your mind. It's just a nice QoL feature to have, and the Reddit client I came from (Relay) had it, so it's just one of those little things I miss on a daily basis. 😊

PR #1519, opened 2024-08-09
Preserve post page state

micahmo:
## Pull Request Description

This PR aims to preserve the state of the post page when navigating out and back in. This is something I've been wanting for a long time (missing this feature from Relay) and I finally got it working! It's done by preserving the `PostBloc` that is created when we navigate to a post page, as well as moving everything important to the `PostState` that wasn't already there (such as the list of collapsed comments and scroll position) instead of in the widget state.

Notes
* We will only save the most recently opened post, so you can navigate out and back in while preserving the state, but if you navigate to another post, everything will reset.
* I implemented this for both the new and old post pages.

## Issue Being Fixed

Issue Number: #742, #1113

## Screenshots / Recordings

(Too big for GitHub)

https://files.catbox.moe/p01rwn.mp4

## Checklist

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -95,6 +95,12 @@ class PostBloc extends Bloc<PostEvent, PostState> {
       _reportCommentEvent,
       transformer: throttleDroppable(throttleDuration),
     );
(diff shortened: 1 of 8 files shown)


micahmo:
Created from wrong branch.

PR #1520, opened 2024-08-09
Preserve post page state

micahmo:
## Pull Request Description

This PR aims to preserve the state of the post page when navigating out and back in. This is something I've been wanting for a long time (missing this feature from Relay) and I finally got it working! It's done by preserving the `PostBloc` that is created when we navigate to a post page, as well as moving everything important to the `PostState` that wasn't already there (such as the list of collapsed comments and scroll position) instead of in the widget state.

Notes
* We will only save the most recently opened post, so you can navigate out and back in while preserving the state, but if you navigate to another post, everything will reset.
* I implemented this for both the new and old post pages.

## Issue Being Fixed

Issue Number: #742, #1113

## Screenshots / Recordings

(Too big for GitHub)

https://files.catbox.moe/p01rwn.mp4

## Checklist

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -79,6 +79,12 @@ class PostBloc extends Bloc<PostEvent, PostState> {
       _reportCommentEvent,
       transformer: throttleDroppable(throttleDuration),
     );
+    on<UpdateScrollPosition>(
+      _onUpdateScrollPosition,
+    );
+    on<UpdateCollapsedComment>(
+      _onUpdateCollapsedComment,
+    );
   }
 
   /// Fetches the post, along with the initial set of comments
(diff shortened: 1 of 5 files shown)


micahmo:
@hjiangsu It looks like you have more time to work on Thunder right now than I do, plus you're more familiar with the refactorings that you've done recently. By any chance could I ask you to try to bring this PR up-to-date? I've really been wanting this feature, and I believe it works very well with this PR (or did when I originally submitted it). 😊 If you don't want to, that's fine, I'll get back to it eventually.

hjiangsu:
Definitely,  I can try to bring it up to date!

hjiangsu:
Alright, I've brought this PR up to date with `develop`! I believe I preserved all the original logic here (had to slightly tweak the positioning of some logic since the codebase has diverged quite a bit from when this was first implemented)

There is a slight visual glitch where navigating back to the previous post causes a slight movement upwards (see video). I'm not entirely sure why this is the case, but if you have any ideas let me know.

https://github.com/user-attachments/assets/6ade9973-6d14-41fc-976b-08759ef35dbf

micahmo:
> There is a slight visual glitch where navigating back to the previous post causes a slight movement upwards

Hey so I tried to reproduce this and I can't (see below video), but I have a theory.

https://github.com/user-attachments/assets/b21248af-cc79-41c0-9927-e54deabc527c

I'm guessing this is related to some image that's loading in. As we know, since there isn't a way to get the image dimensions for images in comments yet, we just show a standard sized placeholder which can resize (and thus shift the position of the rest of the comments) when the image loads in. When I look at your video, sure enough I do see in image above the comment that you scrolled to.

The good news is that the logic works regardless (we end up being scrolled to the right position). So I think this visual glitch will go away if/when we can properly size the image placeholders or come up with some other solution for the image problem.

With that, I'm good to merge this if you are!

hjiangsu:
> I'm guessing this is related to some image that's loading in. 

Ahh yeah, that makes sense. I'll go ahead and merge this in then!

hjiangsu (approved):

PR #1524, opened 2024-08-12
Potentially fix for markdown toolbar not showing above system keyboard

hjiangsu:
## Pull Request Description

This PR attempts to fix an issue where the markdown toolbar does not show up above the system keyboard. I believe the issue may be coming from KeyboardDismissOnTap from `flutter_keyboard_visibility`, but am not entirely certain as I cannot reproduce it on my end.

However, looking at the implementation for KeyboardDismissOnTap, there is some logic that checks whether or not to ignore taps which could be stopping Flutter from being informed when the keyboard is opened? This is a shot in the dark unfortunately.

image

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/comment/view/create_comment_page.dart
+++ b/lib/comment/view/create_comment_page.dart
@@ -269,7 +269,10 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
             }
           },
           builder: (context, state) {
-            return KeyboardDismissOnTap(
+            return GestureDetector(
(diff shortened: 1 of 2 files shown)


micahmo (approved):
Worth a try!

hjiangsu:
I'll merge this in and create another nightly build to see if this fix fixes the original issue!

Issue #1160, opened 2024-03-04
Hold to peek


Kommynct:
### Is Your Feature Request Related to a Problem?

I'd like to be able to hold down on an image to peek at it, without it opening the image or playing any audio

### Feature Description

Peeking would let you hold down on the thumbnail and show you the content/image without playing audio or counting it as a view in your history

### Alternative Solutions

There aren't any

micahmo:
+1 for this feature! It's something Relay had as well. I think at one point we even discussed putting another action on the preview long-press, and those of us that came from Relay advocated for leaving that action open for this feature (but I can't find that comment thread right now). 😊

PR #1526, opened 2024-08-12
Allow peeking at images

micahmo:
## Pull Request Description

This PR adds support for "peeking" at images by long-pressing on the media preview.

> This feature was inspired by Relay and should have a similar implementation.

Notes
* The thumbnail URL is preferred over the full-resolution image, with the idea that peeking should be very fast.
* Peeking an image does not mark the post as read.
* The original issue mentioned something about not playing audio, implying this feature could be used for video. However, this first pass only addresses images.
* Peeking is not yet supported for images in comments.
* As part of these changes, I restored the inkwell animation on image previews which had been lost at some point.

> Review without whitespace.

## Issue Being Fixed

Issue Number: #1160

## Screenshots / Recordings

https://github.com/user-attachments/assets/13bc4233-bbe1-48c7-9085-e259cbb0aedb

## Checklist

### DIFF
--- a/lib/shared/image_viewer.dart
+++ b/lib/shared/image_viewer.dart
@@ -29,12 +29,17 @@ class ImageViewer extends StatefulWidget {
   final int? postId;
   final void Function()? navigateToPost;
 
+  /// Whether this image viewer is being shown within the context of a peek.
+  /// This would cause us to hide unnecsesary things like the buttons at the bottom.
+  final bool isPeek;
+
   const ImageViewer({
     super.key,
     this.url,
     this.bytes,
     this.postId,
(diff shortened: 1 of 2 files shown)


Kommynct:
Would this also work for videos?

micahmo:
> Would this also work for videos?

Hey @Kommynct, great question! As mentioned in the PR description...

> The original issue mentioned something about not playing audio, implying this feature could be used for video. However, this first pass only addresses images.

So yes, this could work for videos, but it would be a separate implementation which I have not done here yet. Sorry!

hjiangsu (approved):
LGTM!

PR #1510, opened 2024-08-01
Show voting when hidden scores are enabled

hjiangsu:
## Pull Request Description

> Review with whitespace off

This PR adds a way to show the voting status of a given post when post scores are hidden. This works with the upvote, downvote, and score metadata widgets.

## Issue Being Fixed

Issue Number: Not a specific issue, but feedback from the community post: https://lemmy.world/post/18027749

## Screenshots / Recordings

Score Metadata:

https://github.com/user-attachments/assets/b8e63e30-eee1-4176-aa34-1785eca395c4

Upvote/Downvote Metadata:

https://github.com/user-attachments/assets/655efd68-e28d-4558-820e-4f4697873d15

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1530, opened 2024-08-13
Fix dim read posts setting not being applied for card view

hjiangsu:
## Pull Request Description

This PR fixes an issue where disabling "Dim Read Posts" was not being respected when using card view. Additionally, I've applied a bit more margin between each metadata widget to improve general readability.

Note: In #1510, I removed the `spacing: 8.0` from the Wrap widget, and moved the spacing to the margins of each metadata widget. However, I only applied a margin of `EdgeInsets.only(right: 4.0)` when it should've been `EdgeInsets.only(right: 8.0)` to keep the same spacing.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/user-attachments/assets/45101476-427c-41b5-af72-877d5e15c503

Padding Changes
| Before | After |
|-|-|
|simulator_screenshot_48F67986-E9D7-4B2D-8E1F-4B30BF629483 | simulator_screenshot_1E803391-CBE8-4730-9B70-D5E4483CADB9 |

## Checklist

### DIFF
--- a/lib/community/widgets/post_card_metadata.dart
+++ b/lib/community/widgets/post_card_metadata.dart
@@ -147,7 +147,7 @@ class ScorePostCardMetaData extends StatelessWidget {
     }
 
     return Container(
-      margin: const EdgeInsets.only(right: 4.0),
+      margin: const EdgeInsets.only(right: 8.0),
       child: Wrap(
         spacing: 2.0,
         crossAxisAlignment: WrapCrossAlignment.center,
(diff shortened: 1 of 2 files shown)


micahmo:
LGTM!

BTW, it's not related to these changes, but does the new combined up/down arrow and vote count change in response to being dimmed, even when colored to indicate a vote? I couldn't tell but it kinda looked like it didn't.

hjiangsu:
> BTW, it's not related to these changes, but does the new combined up/down arrow and vote count change in response to being dimmed, even when colored to indicate a vote? I couldn't tell but it kinda looked like it didn't.

I believe so? It's a bit hard to see due to compression, but both the card/compact views share the same logic so it should be applied!

| | |
|-|-|
|image|image|
|image|image|

Issue #1532, opened 2024-08-14
Remove usage of `USE_EXACT_ALARM` permission


hjiangsu:
### Bug Description

Google Play no longer allows the usage of `USE_EXACT_ALARM` permission unless the app is declared to be a calendar or alarm app.

Because of this, we are unable to publish new versions of Thunder to the Google Play store.

![image](https://github.com/user-attachments/assets/d97994ab-b9fd-4125-9aec-b04eca5d4f08)

### Expected Behaviour

N/A

### Steps to Reproduce

N/A

### Additional Context

This permission was added in this PR: https://github.com/thunder-app/thunder/pull/1343

More context: https://developer.android.com/about/versions/14/changes/schedule-exact-alarms#calendar-alarm-clock

### Device

All

### OS

Android

hjiangsu:
@micahmo

micahmo:
Want me to take care of this? It just means our debugging page won't be able to send local notifications within a minute, which is fine.

hjiangsu:
If possible, that would be great!

PR #1533, opened 2024-08-14
Remove USE_EXACT_ALARM

micahmo:
## Pull Request Description

This PR removes the `USE_EXACT_ALARM` permission to satisfy new Google Play requirements.

* I did not remove the permission from the debug manifest, so we can continue to use more precise alarms for testing. If this still causes issues with Google Play, I can remove it.
* I changed the minimum alarm time to `15` for the background local notification test, and I removed `forceAlarmManager`. Most likely I didn't really have to change the time since it won't be sooner than 15 minutes anyway, but I figured having that number in the code would at least make things more clear.
* I updated the message that is displayed when the user initiates the background local notification test to indicate that it will likely take around 15 minutes until the notification appears. I also tested it, and this was the case.
## Issue Being Fixed

Issue Number: #1532

## Checklist

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -8,8 +8,6 @@
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
     <uses-permission android:name="android.permission.VIBRATE" />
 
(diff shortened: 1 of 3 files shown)


hjiangsu (approved):
LGTM! Thanks for doing this 😄 - I'll let you know if the Google Play build processes properly

Issue #1529, opened 2024-08-13
Links in posts/comments are difficult to tap on


Terrance:
### Is Your Feature Request Related to a Problem?

I'm routinely finding it hard to correctly align a tap on links within the text of posts and comments.  I've turned the font size up to try and make the tap targets larger, though I'd rather have it at the regular size.

This is particularly painful with comments: if you miss, the comment tree gets collapsed instead which is quite jarring.

### Feature Description

Ideally, taps of links to be more forgiving.  I don't know how feasible that is though if it's just relying on a standard Android rich text view or similar.

### Alternative Solutions

Long-pressing a link brings up a panel with a preview and open/share links, whilst long-pressing a comment brings up a menu of actions -- perhaps a submenu of links in the comment could be added to the main actions menu, so that you can go Actions > Link Actions > Some link text > the preview panel as if long-pressing that specific link?

### Additional Context

Weirdly I have not noticed this problem in any other apps, which leads me to think they do something to expand the "hit" area of links.  From a brief play in Firefox for Android, it does feel like I don't need to be as accurate with my taps there -- I would not be surprised to find that browsers and webviews make links more usable in this regard as standard.

micahmo:
I'm not sure if this contributes much to the conversation, but one observation I've made is that it feels harder to tap links due to the lack of an inkwell on the link text itself. Other apps provide a clear visual feedback when you've actually made contact with the link. We tried to add inkwells a while back, but they broke some of the rendering. It might help to bring that back.

Whether this is a real issue with Thunder having a smaller touch target for links is a good question though!

I also like the idea of making all links in a given comment accessible from the long-press sheet. That shouldn't be too hard.

saccma:
The linked issue on the flutter codebase was closed as not planned last week. I'm not sure if that specific problem has been resolved elsewhere.

PR #869, opened 2023-11-01
Improve link handling in posts/comments

micahmo:
## Pull Request Description

This PR improves the functionality of links within posts and comments.

1. There is now an InkWell effect around links to make it more clear that you're tapping on an interactable element.
2. There is now a long-press menu with...
   1. A link preview.
   2. The full URL.
   3. The ability to open, copy, or share.

There are a few caveats with this implementation.
1. Elements that are formatted by markdown and then surrounded by a link lose their additional markdown formatting. This is probably fixable, but with a lot of work.
2. Even more strangely, as soon as we provide our own builder for the `a` element, all text _surrounding_ a link on the same line becomes tappable. To handle this, I removed our `onTapLink` handler, which does solve the problem, but it also means that the text is not tappable in any capacity. That causes a small issue when viewing a `CommentReference` because it means you can't tap on the text surrounding a link to navigate to the comment. You have to tap on another line or outside the body.

Feel free to experiment to see what I mean by these issues, and if they're dealbreakers, I can keep working on it.

> I'm going to leave this as a draft for now until I've had a chance to daily drive it.

## Issue Being Fixed

Issue Number: #309, #709

## Screenshots / Recordings

https://github.com/thunder-app/thunder/assets/7417301/61dc15c0-943d-40e3-8dfd-f7ce9d0899c5

## Checklist

- [ ] Did you update CHANGELOG.md?
- [ ] Did you use localized strings where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1540, opened 2024-08-16
Add a visual indicator to link taps in markdown

micahmo:
## Pull Request Description

Hey! After [over 9 months](https://github.com/thunder-app/thunder/pull/869), I am back with another attempt to add some visual indication when links are being tapped in markdown. This was inspired by #1529, and although it doesn't necessarily fix that issue, it does at least provide some visual or psychological improvement.

Because these changes might seem very odd, I want to describe the process I went through to arrive where I did. Hopefully it will all make sense in the end!

---

### Why the first attempt didn't work

My first attempt at doing this (#869) did not work because of the way that custom builders work in relation to `RichText` and `TextSpan`s. If your builder returns any widget that cannot be placed into the root widget, then the markdown builder will create a `Wrap`, with `RichText`s on either side, and your widget in the middle. As a result, widget wrapping will take precedence over text wrapping. So if you are dealing with a long link that should wrap lines, the custom widget will always be placed on the new line in its entirety, rather than filling the available space on the previous line and wrapping text, because it's not part of the same overall block of text that the rest of the paragraph is. In order to avoid this behavior and allow all text to be wrapped properly, the `LinkElementBuilder`'s `visitElementAfterWithContext` method MUST return widgets in this hierarchy: `Text.rich -> TextSpan -> <the text itself>`. Any other widget in this tree will cause wrapping issues.

The other key here was to ensure that I assigned the same `textScaleFactor` to my `TextSpan` that was assigned to the original `MarkdownStyleSheet`.

### How to show touch feedback without any other widget in the tree

Now I had figured out the formatting problems, so the text looked perfectly fine with my custom builder. But I was stuck with a conundrum. While I could control the widget for each link, I couldn't add an `InkWell` or any other similar widget to visually indicate touches. The best I could do was change the `TextStyle` to add a background. It seemed like a pretty good compromise. It wouldn't have a ripple animation, and I couldn't make the corners nice and rounded, but at least there would be some visual indication. However, while I could set the background statically, I couldn't change it dynamically in response to the tap because I was dealing with all stateless widgets!

### How can you change the background dynamically for a stateless widget

The solution here ended up being twofold. First, I had to make the `CommonMarkdownBody` stateful. But that wasn't enough. Even forcing it to rebuild didn't end up invoking `visitElementAfterWithContext` to give me a chance to recreate the text with a background. I had to go one step further and expose a `forceParseMarkdown` method in `ExtendedMarkdownBody`. With these two things (I double-checked that either one on its own isn't enough), I was able to rebuild the widget and re-parse the markdown in response to the touch!

### How to identify which link to highlight

Now that I had the ability to highlight links and the ability to force the widgets to be regenerated in response to a touch, how could I know which link was being tapped? I had a `TapGestureRecognizer` associated with the first instance of the text widget, but when everything rebuilt, that association was lost. The only information that is provided in `visitElementAfterWithContext` is the `context`, the `element`, and the text styles. The `element` alone is not unique enough to identify it across builds. I can access the text and the URL, but what if someone puts the same URL twice in the same body. I don't want them both to be highlighted.

The solution here was to take advantage of the fact that `Element`s have an arbitrary dictionary of `attributes`. With this, I was able to assign a `UniqueKey` to each `Element` the first time I saw it. Then in the markdown parser, I added recursive code to transfer the `attributes` from the previously built `Node`s to the new ones, every time it's parsed. Now in `visitElementAfterWithContext`, I can check the attributes, see that I recognize the `UniqueKey`, and see that it's associated with the tap event that happened on the previous instance of the widget.

---

Whew, that was a lot, I know! Now whether this is the best approach is highly debatable! But it's what I could come up with! If there is any better approach, or if something is added upstream, I will gladly defer.

I am also going to leave this PR as a draft while I daily drive this. I'm hoping to really put these changes through their paces before opening this up for review to find any potential edge cases that I missed.

# Demo

https://github.com/user-attachments/assets/8b2de068-91e0-445a-8434-0240bdf42a10


micahmo:
I may have found an edge case. 😢

https://programming.dev/post/18211918

Should look like this.

| image |
| - |

But looks like this.

| image |
| - |

hjiangsu:
Oh man, this is definitely a tricky one - I've had plenty of issues when I was working with custom builders for tags like `spoiler`, `subscript`, and `superscript`. I'm glad you were able to find a way to work around most of the issues!

I haven't looked at the implementation for this yet, but I'll let you daily drive this for a bit first before checking out the changes. I'll also let you know if I can think of any alternative solutions to this problem 😁.

Issue #1483, opened 2024-07-09
Search all instance posts


jujubeja:
### Is Your Feature Request Related to a Problem?

As user
I want to add multiple instances of peertube
So that you can view posts from all instances that do not follow each other

### Feature Description

Load List: import instance list by url .txt and .toml and .json
```
https://raw.githubusercontent.com/thunder-app/thunder/develop/instances.toml
https://github.com/thunder-app/thunder/blob/develop/instances.toml
https://github.com/thunder-app/thunder/blob/develop/instances.json
```

### Additional Context

https://lemmyverse.net/
https://data.lemmyverse.net/
https://lemmy.fediverse.observer/list

PR #1557, opened 2024-09-10
Add source from data.lemmyverse.net All instances #1483

ghost:
## Pull Request Description

Add source from data.lemmyverse.net All instances
## Issue Being Fixed

Issue Number: #1483

## Checklist


micahmo:
Hey @atlantis451, thanks so much for the contribution!

The lists that you have modified are automatically generated by a GitHub Action workflow, here:

https://github.com/thunder-app/thunder/blob/develop/.github/workflows/instances.yml

So if we were to merge your change as-is, it would immediately get overridden. 😊

As to why there are so many instances in your list that were not already in ours... well one difference is that you're using `data.lemmyverse.net` where our workflow uses `api.fediverse.observer`. But I think the real difference is that we're filtering by instances that have at least 50 monthly active users. This allows us to keep the list up-to-date with relatively active instances and avoid cluttering it with ones that people set up and then abandoned, or ones that almost no one uses. This list is used as the basis for the "Open by default" feature where you have to manually select each URL that you want Thunder to handle. It would be quite a pain to manually check 579 more instances! 😆

If you would like to either (a) consider changing our workflow to use `data.lemmyverse.net` instead of `api.fediverse.observer` or (b) propose a change to the minimum monthly active users so we can include more instances (as long as it's not a super low number like 0 or 1), I think we could consider that!

P.S. Your change seems unrelated to the issue you linked, #1483, which seems to be more about viewing/searching content across multiple unfederated instances.

PR #1572, opened 2024-10-04
Fix partial percent encoded links in markdown images

hjiangsu:
## Pull Request Description

This is a small PR which fixes an issue mentioned in the following post: https://lemmy.world/post/20288898. The reason why the image was failing to load properly was due to a partially encoded string.

Note: the markdown/image library that we use already handles encoding for URLs - this is a special case where the URL was only partially encoded.

The problematic string that was passed down to the markdown widget was `https://upload.wikimedia.org/wikipedia/commons/4/44/Trať_u_osady_Luh_%28u_Bezpráví%29_1.jpg`
- This is a partially encoded URL as `%28` and `%29` correspond to `(` and `)` respectively. However, all other non-ASCII characters were not percent-encoded

The fix here is to perform full decoding on the URL before passing it to our image library. The image library should automatically handle encoding!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Before | After |
|-|-|
| simulator_screenshot_DADD9979-8E1D-466C-AA41-9F5523D3F9C2|simulator_screenshot_9215BDEB-05C9-4A8B-9F77-E059B760F495|

## Checklist

### DIFF
--- a/lib/shared/common_markdown_body.dart
+++ b/lib/shared/common_markdown_body.dart
@@ -6,9 +6,9 @@ import 'package:markdown/markdown.dart' as md;
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_markdown/flutter_markdown.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
 import 'package:thunder/shared/text/scalable_text.dart';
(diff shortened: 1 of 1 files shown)


ggichure (approved):

micahmo (approved):
Nice fix!

PR #1573, opened 2024-10-04
hard code ndk version and pub upgrade

gwbischof:
I just updated my flutter to the latest stable version. With the latest stable version, `flutter.ndkVersion` is 23.x.
This creates a bunch of warnings when launching thunder on android because a bunch of dependancies require ndkVersion 25.1.8937393 or greater.  So I think the right way to handle this is to write the ndk version in the build.gradle file and not get it from flutter.

Also, I haven't been able to build the app for android after switching to flutter stable. `flutter pub upgrade` fixed it.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -32,7 +32,7 @@ android {
     namespace "com.hjiangsu.thunder"
     // Use variable for version requested by background_fetch and flutter_local_notifications
     compileSdkVersion rootProject.ext.compileSdkVersion
-    ndkVersion flutter.ndkVersion
+    ndkVersion = "25.1.8937393"
 
     compileOptions {
         // The following line is required by flutter_local_notifications
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -162,10 +162,10 @@ packages:
     dependency: "direct dev"
     description:
       name: build_runner
-      sha256: dd09dd4e2b078992f42aac7f1a622f01882a8492fef08486b27ddde929c19f04
+      sha256: "028819cfb90051c6b5440c7e574d1896f8037e3c96cf17aaeb054c9311cfbf4d"
       url: "https://pub.dev"
     source: hosted
-    version: "2.4.12"
+    version: "2.4.13"
   build_runner_core:
     dependency: transitive
     description:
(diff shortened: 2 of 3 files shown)


gwbischof:
It works without updating the ndk version. It just creates warnings.

But I think the it may affect performance of the built app?
https://upstackhq.com/blog/software-development/how-to-use-android-ndk

A bunch of the dependancies require version 25.x:

```
Your project is configured with Android NDK 23.1.7779620, but the following plugin(s) depend on a different Android NDK version:
- android_intent_plus requires Android NDK 25.1.8937393
- background_fetch requires Android NDK 25.1.8937393
- connectivity_plus requires Android NDK 25.1.8937393
- device_info_plus requires Android NDK 25.1.8937393
- dynamic_color requires Android NDK 25.1.8937393
- flutter_custom_tabs_android requires Android NDK 25.1.8937393
- flutter_displaymode requires Android NDK 25.1.8937393
- flutter_file_dialog requires Android NDK 25.1.8937393
- flutter_icmp_ping requires Android NDK 25.1.8937393
- flutter_inappwebview_android requires Android NDK 25.1.8937393
- flutter_keyboard_visibility requires Android NDK 25.1.8937393
- flutter_local_notifications requires Android NDK 25.1.8937393
- flutter_native_splash requires Android NDK 25.1.8937393
- flutter_plugin_android_lifecycle requires Android NDK 25.1.8937393
- flutter_sharing_intent requires Android NDK 25.1.8937393
- gal requires Android NDK 25.1.8937393
- image_picker_android requires Android NDK 25.1.8937393
- package_info_plus requires Android NDK 25.1.8937393
- path_provider_android requires Android NDK 25.1.8937393
- permission_handler_android requires Android NDK 25.1.8937393
- share_plus requires Android NDK 25.1.8937393
- shared_preferences_android requires Android NDK 25.1.8937393
- sqflite requires Android NDK 25.1.8937393
- sqlite3_flutter_libs requires Android NDK 25.1.8937393
- uni_links requires Android NDK 25.1.8937393
- unifiedpush_android requires Android NDK 25.1.8937393
- url_launcher_android requires Android NDK 25.1.8937393
- video_player_android requires Android NDK 25.1.8937393
- webview_flutter_android requires Android NDK 25.1.8937393
Fix this issue by using the highest Android NDK version (they are backward compatible).
```

These dependancies might not get compiled if we don't meet the NDK version requirement, instead they will get run in an interpreter, making the performance worse.  I don't really know, that's just my understanding after doing a few minutes of googling.

The warnings will remind us when we have to update the NDK version in the build.gradle file.

If we put the NDK version in build.gradle then thunder can pick which NDK version to use.  If we don't then it will be up to someones flutter installation to pick the version.

I just updated to flutter stable and it looks like the NDK version is 23.x, which isn't new enough for the dependancies.

gwbischof:
I'm good with whatever, just let me know what you want to do, and ill update the PR.

hjiangsu:
Thanks for the reply - in this case, I think it might be okay if we hard-code the NDK version then (assuming that there will be warnings that show up if we require a higher NDK version).

PR #1580, opened 2024-10-21
Add ability to go to prev/next parent for comment navigator

hjiangsu:
## Pull Request Description

This PR improves the comment navigator for the new post page. In particular, I've added long press actions which navigate to the prev/next top-level comment. I've also improved the logic for the navigator, and https://github.com/thunder-app/thunder/pull/1496#issuecomment-2263800481 should be fixed!

Note: I wasn't sure if the desired action for the long press was to navigate to the prev/next parent, or the prev/next top-level comment. I've opted to do the latter since that makes more sense to me, but this can always be changed!

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1413

## Screenshots / Recordings

https://github.com/user-attachments/assets/1c26a612-3244-47b2-8dc0-6793b653ca70

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -5,8 +5,8 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:lemmy_api_client/v3.dart';
 import 'package:super_sliver_list/super_sliver_list.dart';
-import 'package:thunder/account/models/account.dart';
 
+import 'package:thunder/account/models/account.dart';
 import 'package:thunder/comment/enums/comment_action.dart';
 import 'package:thunder/comment/models/comment_node.dart';
 import 'package:thunder/comment/widgets/comment_card.dart';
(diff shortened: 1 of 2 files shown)


micahmo (approved):
The code LGTM!

In terms of the behavior, I would advocate for the short-press to navigate between top-level comments. I believe that's the most common action, especially considering top-level comments are often off-screen, so it's not as easy to scroll to them. I also checked a couple other apps.

* Relay for Reddit: Short-press navigates top-level comments
* Sync for Lemmy: Short-press navigates top-level comments; long-press goes to the very top or bottom of the comments
* Boost for Lemmy: Same as Sync
* And of course, the old post page in Thunder also navigated top-level comments with a short-press, and we didn't have a long-press action on the arrows.

If you like long-press, maybe it could be a setting?

hjiangsu:
> In terms of the behavior, I would advocate for the short-press to navigate between top-level comments. I believe that's the most common action, especially considering top-level comments are often off-screen, so it's not as easy to scroll to them. I also checked a couple other apps.

I think that's reasonable! I went ahead and made the changes you suggested in dbb94e50f69970a0568afea7113fe68ed9af88b1

micahmo:
Nice, I think this is good to go!

Issue #1575, opened 2024-10-14
Post/reply text is entirely covered by keyboard


vitrola06:
### Bug Description

When writing a post or a reply, I can only see title/URL/language fields or the linked post, respectively. All of the text and also the editing options get covered by the keyboard.

### Expected Behaviour

I should be able to scroll further down to see the text and the editing options.

### Steps to Reproduce

Click the plus button to write a post, click on the text field, see bug.

Go to a post, click the reply button, click on the text field, see bug.

### Additional Context

![Screenshot_20241014-150953](https://github.com/user-attachments/assets/e9f61dfc-eac1-4f7a-ab1e-182f1c145fbc)
![Screenshot_20241014-151036](https://github.com/user-attachments/assets/45e99e6a-5bc1-4489-886c-ee1865109543)
This is all I am able to see.

### App Version

0.5.1 F-Droid

### Device

Motorola One Power

### OS

Android 10

vitrola06:
If you need anything else I could help with, such as getting logs or using my device for tests, let me know.

hjiangsu:
Hey, just an update here - I've pushed out what I _think_ its a fix for this issue. As I can't replicate this issue myself, I would appreciate it if you could update me on the potential fix!

I've released v0.6.0-1 which has the potential fix mentioned in https://github.com/thunder-app/thunder/pull/1581

vitrola06:
Unfortunately, it doesn't seem to have been fixed. The behavior stays the same :(

-------- Mensagem original --------
Em 08/11/2024 19:32, Hamlet Jiang Su  escreveu:

hjiangsu:
Thanks for testing it out! Unfortunately, I'm a bit out of ideas on what could be causing this issue to occur 🙁

Is your device running stock Android? It might only be affecting a specific flavour of Android.

vitrola06:
It's the stock flavor by Motorola, although it should be pointed out that it hasn't been updated since October 2021. Full version name is QPKS30.54-22-27.

-------- Mensagem original --------
Em 09/11/2024 14:21, Hamlet Jiang Su  escreveu:

RandomDennisLee:
Happening here as well, Huawei stock android 12.0.0.175.

hjiangsu:
Just a quick update - I released a new nightly version v0.6.0-2 which _might_ resolve this issue. For some context, there was possibly a misconfiguration in the build file that prevented some devices from re-scaling the screen size when the keyboard is open.

[...]

PR #1581, opened 2024-10-22
Fix create page markdown toolbar visibility

hjiangsu:
## Pull Request Description

This PR potentially fixes an issue where the markdown toolbar is not visible when the keyboard is opened. I haven't been able to reproduce this issue myself, so this fix is a general guess and would have to be tested on an affected device.

The potential fix is to apply a margin equal to the keyboard's height for the toolbar, such that when the keyboard is open, the markdown toolbar is shifted up by the same amount. I've also set `resizeToAvoidBottomInset` to be false so that the scaffold does not automatically scale the page (which might potentially cause overflow issues that may hide the toolbar as well?)

@micahmo if you could just do a quick test on this on your device and let me know if  the toolbar still shows up properly on your end, that would be great!

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1575

## Checklist

### DIFF
--- a/lib/comment/view/create_comment_page.dart
+++ b/lib/comment/view/create_comment_page.dart
@@ -274,6 +274,7 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
                 FocusManager.instance.primaryFocus?.unfocus();
               },
               child: Scaffold(
+                resizeToAvoidBottomInset: false,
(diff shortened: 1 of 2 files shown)


micahmo (approved):
Tested locally and it seems to work well!

PR #1583, opened 2024-10-25
fix the macos build

gwbischof:
Macos build was failing with this error:
```
Error: CocoaPods's specs repository is too out-of-date to satisfy dependencies. To update the CocoaPods specs, run: pod repo update
```

Running this in the macos folder fixed it:
```
pod update sqlite3
pod install --repo-update
```


hjiangsu (approved):
LGTM!

PR #1585, opened 2024-10-26
WIP: Fix the web build

gwbischof:
It seems like we don't need sqflite and sqlite3 deps? 
Drift supports both native and web, and has its own sqlite3 dep.
This PR removes sqflite and sqlite3 deps, and does a conditional import to create the correct database on web and mobile.


hjiangsu:
> Do we need the sqflite and sqlite3 deps? It looks like they are only being used for migrating from the the old stile database. Can we use Drift to connect to the old style database and migrate it to the new one?

You're correct, we're only using `sqflite` for database migration tests. However, I would still like to keep the dependencies here for a bit longer (just in case there are still users using the old database schemas). In the future, we'll most likely remove the dependencies and the migration tests, but for now, I would keep them as is!

On a different note, is it okay if I apply some modifications to this PR? I think I might be able to simplify some of the logic further!

gwbischof:
Yep go for it, do you need permissions?
I was thinking a new PR might be easier than adding the SQLite deps back to this one.

hjiangsu:
Thanks, I can make a commit directly to this branch to keep things in order!

hjiangsu:
I'll list out the changes I made here (https://github.com/thunder-app/thunder/pull/1585/commits/f92770aceda3d011b5cef9ee5125acfa48afb442):
- Updated `drift` and related dependencies
- Brought back `sqflite` and `sqflite_common_ffi` as `dev_dependencies` (we'll remove these in the future)
- Improved some of the logic for the database initialization (removed `constructDb` in favour of `driftDatabase`)
  - This is how they seem to handle it in their example app: https://github.com/simolus3/drift/tree/develop/examples/app

@gwbischof Could you run a few tests on your end and let me know if this builds properly for all the platforms you mentioned above? Thanks!

gwbischof:
Looks good, I tested on all of my platforms.  There are some handled exceptions on web, but its building and it looks like its working.

lib/settings/pages/debug_settings_page.dart
```diff
@@ -244,19 +244,19 @@ class _DebugSettingsPageState extends State<DebugSettingsPage> {
                   contentText: l10n.deleteLocalDatabaseDescription,
                   onSecondaryButtonPressed: (dialogContext) => Navigator.of(dialogContext).pop(),
                   secondaryButtonText: l10n.cancel,
-                  onPrimaryButtonPressed: (dialogContext, _) async {
-                    String path = join(await getDatabasesPath(), 'thunder.db');
+                  // onPrimaryButtonPressed: (dialogContext, _) async {
```
gwbischof, line 247: I think we need to figure out a way to handle this.
I don't see a drift method to delete the whole database.

hjiangsu (approved):
LGTM!

PR #1565, opened 2024-09-24
Chore: Upgrade dependencies for Flutter 3.24 compatibility

hjiangsu:
## Pull Request Description

This PR updates Thunder to be compatible with Flutter 3.24. I've upgraded all the packages to their latest versions (using `flutter pub upgrade --major-versions`)

Note: The package `flex_color_scheme` has not been updated yet as the next version is still a WIP. However, I opted to continue with the upgrade regardless to reduce potential tech debt from upgrading between multiple major Flutter versions (e.g., 3.22, 3.24, etc.)

Due to this, there may be some visual bugs associated with the upgrade. I'll see if there are any simple workarounds for these issues in future PRs.

@micahmo I've tested the build on Android and it looks to be building properly, but let me know if you encounter any issues! I'll be closing this PR (https://github.com/thunder-app/thunder/pull/1372) as this one supersedes it

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1587, opened 2024-10-29
Fix theming issues with dependency upgrades

hjiangsu:
## Pull Request Description

> Review with whitespace off

This PR fixes some of, if not all, the theming issues mentioned in https://github.com/thunder-app/thunder/pull/1565. I've also upgraded the `flex_color_scheme` package to `8.0.0-dev.2` which contains compatibility with Flutter 3.24.

Notably, the following issues should be fixed:
- bottom nav bar not using shaded colour
- inbox reply cards not using tinted background
- drawer header having coloured background
- search bar background not correct
- account switcher backgrounds not matching bottom sheet

I'm not entirely sure if these issues are fixed (from https://github.com/thunder-app/thunder/pull/1565#issuecomment-2411877790):

> - It looks like the text field hint color is darker (it looks like text that has been typed in, rather than a hint)
> - I think the sizes of the text fields in the login modal are not as tall any more.

Please let me know if you encounter any additional theming issues!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/account/widgets/profile_modal_body.dart
+++ b/lib/account/widgets/profile_modal_body.dart
@@ -242,7 +242,7 @@ class _ProfileSelectState extends State<ProfileSelect> {
                     child: Padding(
                       padding: const EdgeInsets.fromLTRB(10, 0, 10, 0),
                       child: Material(
-                        color: currentAccountId == accounts![index].account.id ? selectedColor : null,
+                        color: currentAccountId == accounts![index].account.id ? selectedColor : Colors.transparent,
                         borderRadius: BorderRadius.circular(50),
                         child: InkWell(
                           onTap: (currentAccountId == accounts![index].account.id)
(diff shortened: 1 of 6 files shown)


micahmo (approved):
All the code changes LGTM! After you merge I'll take a look at the other two issues I found and see if they look any better now. Thanks a ton!!

micahmo:
@hjiangsu Two quick followups on this so far.

* the system navigation bar on Android no longer has the same color app nav bar
* the background color of the instance tagline should be a different color
* the title bar should have a different color once the user scrolls down (this is only broken when the top bar is _not_ pinned)

hjiangsu:
Thanks for mentioning those issues! Would it be possible for you to send some screenshots over of what you mean by them?

micahmo:
> the system navigation bar on Android no longer has the same color app nav bar

Just FYI, I did some looking into this and found some interesting results. It seems like the `SystemUiOverlayStyle` that we're setting is not being respected at all. As a test, I set the nav bar color to purple. You can see in the following screen recording that if I hot reload the app, it respects the color, but as soon as I perform any action, it goes back to (nearly) white. I haven't been able to find any bugs or known breaking changes related to this, though, so I'm not sure what else to do at this point.

https://github.com/user-attachments/assets/59112210-ad56-4986-9a42-d06be47a16a2

I'm wondering if it has something to do with he `flex_color_scheme` upgrade. This issue seems *maybe* related??
https://github.com/rydmike/flex_color_scheme/issues/258

hjiangsu:
Interesting, thanks for looking into that! I'm wondering if there's a theming setting within `flex_color_scheme` that colours the system navigation bar. If so, that could be why it's being overridden when an action is performed.

I can also take a quick look to see if I spot anything regarding this.

PR #1588, opened 2024-10-31
Fix linting and deprecation warnings

hjiangsu:
## Pull Request Description

This PR cleans up some additional linting errors and deprecation notices on the project. As a summary:
- I've decided to ignore `use_build_context_synchronously` messages in `analysis_options`. I haven't noticed any issues with using context within async functions, and the linting messages end up flooding the rest of the more important infos/warnings. If you feel strongly against this move, let me know and I can bring it back!
- Fixed deprecations with regards to colours (background, onBackground, etc.)
- Removed unused imports, and unused variables

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
LGTM!

Should any of the color changes fix any issues, or are they only to address deprecations?

> I've decided to ignore `use_build_context_synchronously` messages in `analysis_options`. I haven't noticed any issues with using context within async functions, and the linting messages end up flooding the rest of the more important infos/warnings. If you feel strongly against this move, let me know and I can bring it back!

I am totally good with this! The `if (context.mounted)` thing has always felt like more of a chore with not much benefit. If the context really isn't mounted in some rare cases, then it'll just silently fail.

hjiangsu:
> Should any of the color changes fix any issues, or are they only to address deprecations?

They should only affect deprecations as far as I'm aware. The issues that you mentioned are still present (some of them will be addressed in a different PR, and others will hopefully be fixed when the package transitions out of a dev build)

PR #1590, opened 2024-11-07
Upgrade flex_color_scheme to v8.0.0 stable version

hjiangsu:
## Pull Request Description

This is a small PR which updates our `flex_color_scheme` dependency to the stable version. Note that this doesn't seem to fix the visual issues mentioned in https://github.com/thunder-app/thunder/pull/1587#issuecomment-2447918608, so we'll have to do some manual fixes for them.

We can apply fixes to the visual issues mentioned above in a later nightly build (as they are fairly minor and I would like to push out a nightly build in the next couple of days)

@micahmo it would be helpful if you could do a quick check to make sure everything is working as expected and that there aren't any additional regressions from this upgrade ☺️!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Aside from the known issues this looks good!

PR #1594, opened 2024-11-17
Use image dimensions from API (v0.19.6)

hjiangsu:
## Pull Request Description

This PR adds the ability for Thunder to use the image dimensions provided from the Lemmy API (for instances using v0.19.6 and above). This should help improve overall app performance for those using instances v0.19.6 and above since it allows us to bypass `retrieveImageDimensions` function to pre-load images to fetch their dimensions.

As a result, this will also potentially help reduce layout jank on the feed page since more posts will be able to properly fetch the dimensions.

I have also added in the error message for the snackbar that pops up after creating a new post. This should help narrow down the issue found in the Matrix room.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Nice, this should be a big improvement!

Issue #1592, opened 2024-11-08
Filters that limit first page cause hard stop


Fmstrat:
### Bug Description

Today I added a few political filters, and uncovered a bug.

If the entire first page is caught in the filter, Thunder acts like you've hit the bottom of the feed.

### Expected Behaviour

If the first page is empty due to filters, grab the next.

### Steps to Reproduce

Set a filter for words until it catches the first page.
Behold empty feed.

### App Version

0.5.1

### Device

Pixel 8

### OS

Android

PR #1596, opened 2024-11-18
Fix incorrectly triggered end of feed when too many filters are applied

hjiangsu:
## Pull Request Description

This PR fixes an issue where the feed stops fetching for new posts when there are too many filtered/deleted posts which cause the total number of fetched posts to be zero. Instead of determine the end of the feed from the filtered posts, it now uses the original response's length to determine the end of the feed.

## Issue Being Fixed

Issue Number: #1592

## Checklist

### DIFF
--- a/lib/feed/utils/post.dart
+++ b/lib/feed/utils/post.dart
@@ -51,6 +51,9 @@ Future<Map<String, dynamic>> fetchFeedItems({
         showHidden: showHidden,
       ));
 
+      // Keep the length of the original response to see if there are any additional posts to fetch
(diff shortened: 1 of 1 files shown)


micahmo (approved):

micahmo:
@hjiangsu I'm not sure if it's related to this change, but I've started noticing that my feed always displays "No more items to load" at the end (instead of a spinner) before it proceeds to load more items. Here's a recording.


https://github.com/user-attachments/assets/b79d0835-3284-4ef3-b9f1-992e3945efd5

Issue #1521, opened 2024-08-11
Account section is endlessly refreshing


halawnss:
### Bug Description

Account section is not working , it only continues to refresh

### Expected Behaviour

To show account's info

### Steps to Reproduce

?

### App Version

0.4.0

### Device

Redmi 7

### OS

Android 10

micahmo:
Hey @halawnss, thanks for the bug report! Any chance you could provide some additional info?

* What instance are you using?
* If you remove your account and log back in (in Thunder) does that fix the issue?
* Can you navigate the account settings on the web UI of your instance?
* Could you provide a screen recording of the issue?

Thanks for any and all info you can provide!

halawnss:
1- lemdro.id

2- I've tried before and it didn't fix it

3- Yes

4- 
https://github.com/user-attachments/assets/f09b2d70-d742-4c3a-8e23-504e9320667d

hjiangsu:
Hey @halawnss, thanks for this - I believe the workaround for now is to manually press the refresh icon on the account page. This should hopefully load up the account information properly!

For devs, I have seen this similar behaviour but I'm not sure how to properly reproduce it. I believe this regression might've been introduced in #1354?

halawnss:
Thank you for your efforts
I can reach the account info now

PR #1597, opened 2024-11-19
Update Account page logic

hjiangsu:
## Pull Request Description

This draft PR updates our Account page logic to use the feed page under-the-hood. ~~This feature is currently only enabled when experimental features are on.~~ Migrating the account page to use the feed page under the hood provides us with a few major benefits:
- Overall improved performance, since we will be using slivers instead of the standard ListView builder
- Automatic hiding of user header bar (similar to user feed pages) as it uses the same logic as the feed page. This allows us to view more of our posts/comments on the screen
- Reduces code complexity and allows us to deprecate older legacy code that was kept for the old Account page
- Reduces the number of bugs due to missing blocs/context since it is all shared now. This allows us to perform all actions as we would on the user feed pages.
- Enables the ability to change the sorting of our posts/comments

There are a few changes that had to be made in order for this to work:
- When viewing the Account page, the app bar has been adjusted to show the current account page actions. The app bar also includes additional actions present in the general user feed pages (sharing user profile and sorting)
- The app bar now has a toggle to switch between saved posts/comments, and your own posts/comments. While it would be possible to bring back the old behaviour eventually, this is currently the simplest way to handle the logic for now without too many dependencies being created across widgets.

## Issue Being Fixed

As a side effect, this PR should also solve #1521!

Issue Number: #1521

## Screenshots / Recordings

Some screenshots of the new Account page.

| Posts/Comments | Saved Posts/Comments |
|-|-|
|Screenshot_1732076085 |Screenshot_1732076100|

## Checklist


micahmo:
Looking forward to this! 😊

hjiangsu:
Alright, I believe I implemented most of the functionality now!

@micahmo feel free to try this out and let me know if you notice any issues or any missing features that were present in the previous implementation!

lib/account/widgets/account_page_app_bar.dart
```diff
@@ -0,0 +1,212 @@
...
+  final void Function(bool showSaved)? onToggleSaved;
+
+  @override
+  Widget build(BuildContext context) {
+    final l10n = AppLocalizations.of(context)!;
+    final feedBloc = context.read<FeedBloc>();
+
+    return Row(
+      children: [
+        IconButton(
```
micahmo, line 143: Any chance you could add a tooltip here?

micahmo (approved):
Very nice!!

lib/account/widgets/account_page_app_bar.dart
hjiangsu: Done in https://github.com/thunder-app/thunder/pull/1597/commits/a736b9d0eff5dcfff5a689c9d993091c88fb2076!

PR #1600, opened 2024-11-20
Fix background color

micahmo:
## Pull Request Description

Alrighty I finally took the time to look into some of the color issues. As much as I wanted to clean up my old PRs before moving on to new stuff, I think these issues need to be fixed before we can release a new stable version. I also finally built the last version of the app prior to the color changes, so I can easily reference back and forth. (If anyone else is interested in doing the same, you can use commit 0e148d16cba2bbdff8fa3ab34a75aca804581213. You will also have to downgrade your local Flutter version to `3.19.x`. And for some reason I also had to upgrade `share_plus` to `^10.0.0`.)

I am starting by tackling `getBackgroundColor`, which is a function we use quite a bit throughout the app to generate a nice "background color" in various circumstances. It was previously based on `theme.dividerColor`, but either that has changed, or the way `darken` and `lighten` work has changed. To maintain consistency, I have decided to just hard-code the colors to what they used to be. It's not really based on the theme anyway, so I don't think it's a big deal.

One of the easiest places to see the color in use is behind instance taglines, so that's what I'll use for my screenshots, but it is used elsewhere throughout the app as well. Note that the difference in dark/black mode is very subtle, but I used a color picker to make sure it's exactly the same now.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/pull/1587#issuecomment-2445469432

## Screenshots / Recordings

|                  | Light                                                  | Dark                                                  | Pure Black                                             |
|------------------|--------------------------------------------------------|------------------------------------------------------|-------------------------------------------------------|
| **Old Version**  | old_light | old_dark | old_black |
| **Before**       | before_light | before_dark | before_black |
| **After**        | after_light | after_dark | after_black |

## Checklist

### DIFF
--- a/lib/utils/colors.dart
+++ b/lib/utils/colors.dart
@@ -1,13 +1,12 @@
-import 'package:flex_color_scheme/flex_color_scheme.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:thunder/core/theme/bloc/theme_bloc.dart';
+import 'package:thunder/utils/constants.dart';
 
 /// Gets a tinted background color that looks good in light and dark mode
 Color getBackgroundColor(BuildContext context) {
   final bool darkTheme = context.read<ThemeBloc>().state.useDarkTheme;
-  final ThemeData theme = Theme.of(context);
-  return darkTheme ? theme.dividerColor.darken(5) : theme.dividerColor.lighten(20);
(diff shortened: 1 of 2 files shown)


hjiangsu (approved):
LGTM! Just curious, but how does it look on Material You themes?

micahmo:
> how does it look on Material You themes?

Seems ok!

| image | image |
| - | - |

PR #1603, opened 2024-11-25
Fix create post snackbar error message

hjiangsu:
## Pull Request Description

This PR should fix the error message when displaying the snackbar after creating a new post. I just added a conditional to check if the FeedBloc is present within the tree, and trigger the FeedItemUpdatedEvent only if its available.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/post/utils/navigate_create_post.dart
+++ b/lib/post/utils/navigate_create_post.dart
@@ -72,7 +72,7 @@ Future<void> navigateToCreatePostPage(
                     },
                   );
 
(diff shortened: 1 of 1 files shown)


hjiangsu:
I just checked the logic again, and there are cases where we might not have the FeedBloc in the widget tree, the main one being the FeedFAB. The FAB is located above the FeedPage (in `thunder_page.dart`), which means that it likely doesn't have the FeedBloc context (and thus, why it was triggering the error)

We may need to solve this issue in a different PR, but it doesn't seem like it affects anything major at the moment so we should be okay with just the fix I have now!

Issue #1598, opened 2024-11-19
No way to use pure black theme while following system theme


rrconkle:
### Bug Description

When following system theme there is no way to use the pure black dark mode.

### Expected Behaviour

There is a toggle to use pure black.

### Steps to Reproduce

.

### App Version

0.5.1

### Device

Pixel 8

### OS

Android 15

PR #1605, opened 2024-11-27
Add option to toggle pure black mode when using system theme

hjiangsu:
## Pull Request Description

This PR adds the ability to toggle on pure black mode (OLED) when using the system theme. When this is enabled, any time the system theme is set to dark mode, Thunder will use the pure black variant rather than the default dark mode variant for theming.

This setting is only visible when the user chooses "System" as the theme.

Let me know if you have a better setting name/description for this setting!

## Issue Being Fixed

Issue Number: #1598

## Screenshots / Recordings

https://github.com/user-attachments/assets/535af804-2ddb-450f-94a9-ecab6dec9910

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -465,6 +465,7 @@ extension LocalizationExt on AppLocalizations {
       'nestedCommentIndicatorColor': nestedCommentIndicatorColor,
       'reduceAnimations': reduceAnimations,
       'theme': theme,
+      'systemThemePureBlack': systemDarkMode,
       'themeAccentColor': themeAccentColor,
       'useMaterialYouTheme': useMaterialYouTheme,
       'actionColors': actionColors,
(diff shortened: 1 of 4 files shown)


micahmo (approved):
LGTM

glyph-se:
> Let me know if you have a better setting name/description for this setting!

I would combine this new setting with the above selector/dropdown.

Either making it have 5 values:

* Light
* Dark
* Pure Black
* System with Dark
* System with Pure Black

Or changing it to be:
* System
* Light
* Dark

And then if Dark is selected, show the toggle:

Pure Black - Enable pure black theme instead of regular dark mode.

micahmo:
@glyph-se, thanks for the suggestions. I think @hjiangsu and I both both appreciate any advice on UX. 😊

> * Light
> 
> * Dark
> 
> * Pure Black
> 
> * System with Dark
> 
> * System with Pure Black

Personally I would find this a bit confusing. For example, the option "System with Dark" really means "follow the system setting (whether it's light or dark), but when it's dark, just use regular dark mode, not pure black mode". In my opinion, that full meaning can't really get communicated in a single option.



> * System
> 
> * Light
> 
> * Dark
> 
> 
> And then if Dark is selected, show the toggle, which will affect both "Dark" and "System":
> 
> Pure Black - Use pure black theme instead of dark mode.

I like this idea a lot more! It's similar to what @hjiangsu already implemented (with a separate toggle for pure black) but without the extraneous explicit option for just pure black. In this way, the pure black mode is always an additional option on top of regular dark mode. I could even see the toggle being included in the bottom sheet so that everything color/theme related is in one dialog. So this would be my vote!

PR #1606, opened 2024-11-27
Add setting to automatically expand taglines

hjiangsu:
## Pull Request Description

This PR adds a new setting under General -> Feed which allows taglines to be automatically expanded. I've also defaulted `taglineIsLong` to be true as I noticed long taglines cause a weird animation as it transitions from expanded  -> collapsed when initially loading the feed.

## Issue Being Fixed

Issue Number: Feature request from the Thunder Lemmy community: https://lemmy.world/post/22153408

## Screenshots / Recordings

https://github.com/user-attachments/assets/88f2c3b7-80c6-4ad0-99c9-30e13f651f69

## Checklist


micahmo (approved):
LGTM!

Something else you could do is change the threshold at which we determine a tagline to be long so that, even without this new setting enabled, people are more likely to see full taglines without having to expand.

hjiangsu:
I've doubled the height threshold for determining long taglines - ideally, this should translate to seeing at most 2 lines of text before the tagline is considered too long (but won't always be the case due to per-device font sizing).

Let me know if this works!

micahmo:
That sounds good!

Issue #1607, opened 2024-11-27
Thunder re-loads in background too often.


shortwavesurfer2009:
### Bug Description

If I go back and forth between Fennec and Thunder, I will often find that Thunder reloads in the background, and I lose posts I was creating.

### Expected Behaviour

Thunder should be able to remain open.

### Steps to Reproduce

1. Open an article you wish to share in fennec
2. Open thunder
3. Open menu and choose a community
4. Tap "+" to create a new post
5. Use app switcher to switch to fennec, copy article title
6. Switch back to thunder via app switcher (problem might occur at this point)
7. Paste title in title box
8. Switch to fennec with app switcher, open menu, share, copy the url
9. Switch to thunder via app switcher (problem may occur here as well)

### Additional Context

This bug does not occur all the time, but it occurs often enough to be quite irritating.

### App Version

<= 0.6.0-1

### Device

OnePlus Nord N200

### OS

LineageOS 21 (Android 14)

hjiangsu:
This problem generally occurs when Android runs out of memory and tries to free up memory by suspending background apps.

**For devs:**
Flutter has the [ability to restore state when the OS kills the app](https://docs.flutter.dev/platform-integration/android/restore-state-android) (when freeing up memory), however, we don't currently use that feature. Implementing this will take a bit of effort but we could possibly break this up into stages:
- Implement restorable states for routing (e.g., opening the app would restore back the navigation that we were in previously). Initial work is started with #1613.
- Implement restorable states for inputs (e.g., text inputs, dropdowns, toggles, etc)
- Implement restorable states for more complex states (bloc related data such as feed, posts, comments, etc). This part will likely be the most time-consuming/complex implementation from this list
- Implement restorable states for scroll position (e.g., keep the specific scroll position within the feed)

shortwavesurfer2009:
See, that's what's weird though, is that my phone has a decent amount of RAM, and other apps do not do this, and I don't have a bunch of other apps open in the background at all times either.

RodoMa92:
I can also still reproduce this easily enough on a 8 GB smartphone on latest release (0.6.1).

PR #1613, opened 2024-11-27
Remove go_router dependency for navigation

hjiangsu:
## Pull Request Description

This PR removes our dependency on `go_router`, and migrates our existing usage of `go_router` with the built in Navigator. There's a few reasons for this refactor:
- We currently use a mix of `go_router` and Navigator. Since we are only using `go_router` for the Settings page, it seems a bit redundant to have the package when we could implement the same logic using Navigator. Thus, it makes more sense to use Navigator unless we require a specific feature from `go_router` (e.g., ShellRoutes)
- Using only one of Navigator / `go_router` allows us to simplify state restoration later on (e.g., when the app is killed by the OS to free up memory). Navigator has a simple method of creating restorable states whereas `go_router` is a bit more complex (and may not fully work well in some cases)

This is still in draft as there's a bit more testing / refining that is needed:
- Testing deep links on Android / iOS (tapping a link from outside Thunder to open in Thunder)
- DONE: Testing of routing within the Settings page to ensure all routing is properly handled (also, checking to see if Thunder-setting links open up properly #1348)

## Issue Being Fixed

This tangentially relates to, and is a first step to hopefully solving #1607. There is still a lot of additional steps that is required, but being able to restore the navigation state is the first step (e.g., to be able to navigate back to the CreatePostPage)

Issue Number: N/A

## Checklist


hjiangsu:
So after some testing, it turns out that deep links does not seem to work in the current build on iOS (both `develop` and this branch).

@micahmo Could you check if deep linking works on the current `develop` for Android? If it's not working on Android either, I'll open up a new PR for that issue rather than to place the fix here.

Aside from that, I've tested everything else and it should be working so this is ready to go!

micahmo (approved):
The code LGTM! Did you still need me to test anything in particular?

micahmo:
> @micahmo Could you check if deep linking works on the current `develop` for Android?

Yes deep links are still working fine in `develop`!

PR #1621, opened 2024-12-03
Fix text hint color

micahmo:
## Pull Request Description

This PR fixes an issue where the text field hint text color looks much darker than before (and darker even than the text that gets typed!).

I didn't bother doing an exact color match, I just wanted the text to look slightly faded.

Again I tested with light, dark, and pure black.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/pull/1565#issuecomment-2411877790

## Screenshots / Recordings

You'll notice that the text field size is also different. I plan to address that separately. It was also mentioned in the linked comment.

https://github.com/user-attachments/assets/0a806d55-4659-40d8-8c6c-747d9e8e66e7

## Checklist

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -234,6 +234,22 @@ class _ThunderAppState extends State<ThunderApp> {
               theme = theme.copyWith(pageTransitionsTheme: pageTransitionsTheme);
               darkTheme = darkTheme.copyWith(pageTransitionsTheme: pageTransitionsTheme);
 
+              // Set some additional styling on the themes
+              theme = theme.copyWith(
+                inputDecorationTheme: InputDecorationTheme(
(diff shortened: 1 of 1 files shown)


lib/main.dart
```diff
@@ -234,6 +234,22 @@ class _ThunderAppState extends State<ThunderApp> {
               theme = theme.copyWith(pageTransitionsTheme: pageTransitionsTheme);
```
hjiangsu, line 234: I think perhaps moving the additional styling into here would make more sense (since we're already modifying the themes here)!

micahmo: Done!

hjiangsu (approved):
LGTM!

PR #1622, opened 2024-12-05
Fix TextField height

micahmo:
## Pull Request Description

This PR fixes the appearance of `TextField`s which seemingly shrunk in height after the upgrades. It turns out that the ones with this problem were marked as `isDense`. To be honest, it might have been a bug before that they were marked as dense but didn't look any different, and maybe now after the upgrade the dense flag is being respected more. However, unless we like the change, removing the flag allows us to restore the previous appearance.

Note: I left `isDense` in the CreatePostPage because we have intentionally shrunk those fields in order to make better use of the screen real estate (see #1498). For whatever reason, this usage of `isDense` looked the same before as it does now.

Also I believe this is the last of the color/style issues since the big upgrades, so it will be nice to be able to move on from this and get back to working on features. :-) It also makes me feel a lot better about doing a new release soon.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/pull/1565#issuecomment-2411877790

## Screenshots / Recordings

https://github.com/user-attachments/assets/2864bac2-a013-4a9d-a2dd-d2c692bc0a5e

https://github.com/user-attachments/assets/0a6d3be5-006f-4a39-9f92-380249f400cb

## Checklist

### DIFF
--- a/lib/account/pages/login_page.dart
+++ b/lib/account/pages/login_page.dart
@@ -319,7 +319,6 @@ class _LoginPageState extends State<LoginPage> with SingleTickerProviderStateMix
                       focusNode: focusNode,
                       inputFormatters: [LowerCaseTextFormatter()],
                       decoration: InputDecoration(
-                        isDense: true,
                         border: const OutlineInputBorder(),
                         labelText: AppLocalizations.of(context)!.instance(1),
                         errorText: instanceValidated ? null : instanceError,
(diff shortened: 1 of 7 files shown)


hjiangsu (approved):
LGTM! Thanks again for looking into the colour/styling issues! ☺️

PR #1624, opened 2024-12-06
Do not re-load inbox when already selected

micahmo:
## Pull Request Description

This PR fixes a pretty small issue where tapping on the inbox page when it was already selected would cause it to reload again (and for some reason it would clear out the contents of the inbox). Rather than trying to figure out why that happens, I decided to make it so that tapping on the inbox when it's already selected does nothing. I think that makes more sense anyway!

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/user-attachments/assets/46a3e935-581a-478b-bdf4-90517f6596a9

#### After

https://github.com/user-attachments/assets/40643c8b-89ac-45b4-bfd7-86a3ea50465f

## Checklist


hjiangsu (approved):
LGTM!

PR #1626, opened 2024-12-06
Migrate from uni_links to app_links for deep links

hjiangsu:
## Pull Request Description

This PR migrates our existing usage of `uni_links` over to `app_links`. The main reason for this being that `uni_links` has been discontinued, and has been replaced by `app_links` instead. See https://pub.dev/packages/uni_links for more details.

@micahmo I know you just tested deep links, but it would be helpful if you could re-test them with this migration! I did a quick test on an Android emulator and it seems to be working, so I'm hoping it works seamlessly on a physical device (especially in profile/release modes as I needed to add some proguard rules according to their setup).

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -26,7 +26,7 @@
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
             android:hardwareAccelerated="true"
-            android:resizeableActivity="true">
+            android:resizeableActivity="true"
             android:windowSoftInputMode="adjustResize">
             <!--Allowing support for multi windows (split-screen mode) & picture-in-picture mode -->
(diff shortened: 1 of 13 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -26,12 +26,12 @@
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
             android:hardwareAccelerated="true"
-            android:resizeableActivity="true">
+            android:resizeableActivity="true"
```
hjiangsu, line 29: I think I may have also inadvertently found the issue that was causing #1575! I'm not entirely sure if this is the case, but there was an extra `>` which might've caused `android:windowSoftInputMode="adjustResize"` to not be respected.

micahmo: Nice find!

micahmo (approved):
The code LGTM! I tested both `thunder://` links and Lemmy links and everything's good!

PR #1628, opened 2024-12-09
Allow navigating to modlog for removed comment

micahmo:
## Pull Request Description

This PR allows us to see the modlog for comments deleted by mods. This can be helpful if a comment was deleted which removes the whole context of the comment thread. Since the information is already available in the modlog, I don't see a problem with making it easy for the user to navigate there. This option is only shown for comments removed by a mod.

Note that since the modlog API call doesn't have a parameter to specify a particular comment, we have to manually check/filter the results. However, initially filtering by `ModlogActionType.modRemoveComment` helps to narrow the scope.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/user-attachments/assets/34cbcb3b-3e33-4f75-b0fe-85eb592d45d8

## Checklist

### DIFF
--- a/lib/core/singletons/lemmy_client.dart
+++ b/lib/core/singletons/lemmy_client.dart
@@ -95,6 +95,7 @@ enum LemmyFeature {
   multiRead(0, 19, 0, preRelease: ["rc", "1"]),
   hidePosts(0, 19, 4),
   customThumbnail(0, 19, 4),
+  commentModLog(0, 19, 4),
   imageDimension(0, 19, 6),
   ;
 
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2813,6 +2813,10 @@
   "@viewCommentSource": {
     "description": "Menu item for viewing a comment's source"
   },
+  "viewModlog": "View Modlog",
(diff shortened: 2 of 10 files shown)


hjiangsu:
> Note that since the modlog API call doesn't have a parameter to specify a particular comment, we have to manually check/filter the results. However, initially filtering by ModlogActionType.modRemoveComment helps to narrow the scope.

Hmm, can you clarify what you mean here? I believe there is a parameter for `commentId` that we can pass to the modlog API (although do note that this is limited to Lemmy versions v0.19.4 and above). See https://github.com/LemmyNet/lemmy-js-client/compare/0.19.3-alpha.1...0.19.4#diff-d5cb9035704a2a8bf23588d5a9a725d376f9cde198a6c77763d4583efdd16e09

micahmo:
Wow, good catch. I was even passing `commentId` to `GetModlog`. Duh!

Ok I removed the custom filtering, and also put a conditional on the menu item to only work for 0.19.4+. Should be good to go now!

hjiangsu (approved):
LGTM! Ignore the failed CI check, that seems to be a different issue that I'll take a closer look at.

PR #1630, opened 2024-12-09
Fix the status bar color on the post page

micahmo:
I noticed that the post page suffered from the same issue as #1615. Fortunately, the same fix worked perfectly!

> Review without whitespace.


hjiangsu (approved):
LGTM!

PR #1634, opened 2024-12-12
Fix issue with app hanging after update to v0.6.0-2

hjiangsu:
## Pull Request Description

This PR attempts to fix an issue with the latest nightly build hanging at the splash screen. I believe this is the cause of the issue but am not 100% sure, however, it could be due to a null check error when Material You theme is enabled.

I'll be creating a new nightly build off of this branch since there's been commits made already to the `develop` branch!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/globals.dart
+++ b/lib/globals.dart
@@ -1,1 +1,1 @@
-const String currentVersion = '0.6.0-2+68';
+const String currentVersion = '0.6.0-3+69';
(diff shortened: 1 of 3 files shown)


hjiangsu:
> will the removals break the color scheme?

This might only affect users using Material You and a dark theme applied - I figured it'd be better to remove it now and make a proper fix for it once I push out a new nightly build!

PR #1635, opened 2024-12-12
Fix a couple more color/theme things

micahmo:
This PR fixes two more color issues
- The modlog status bar (same fix as #1630 and #1615)
	- This appears to be the last case of `SafeArea.top` being determined by the `hideTopBarOnScroll` setting, so hopefully we're done with these fixes.
	- I also allowed the `SafeArea` to go under the bottom nav bar, which makes the app feel more edge-to-edge, and also matches the post/comment page.
- The create post/comment page nav bar
	- Pretty similar fix here. I allowed the `SafeArea` to go under the nav bar and added a `Container` with the same background color as the `MarkdownToolbar` to fill the space. I verified that opening the keyboard only causes the toolbar to jump up and not the extra container.

### DIFF
--- a/lib/comment/view/create_comment_page.dart
+++ b/lib/comment/view/create_comment_page.dart
@@ -281,6 +281,7 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
                   centerTitle: false,
                 ),
                 body: SafeArea(
+                  bottom: false,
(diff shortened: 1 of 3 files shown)


hjiangsu (approved):
LGTM!

PR #1638, opened 2024-12-12
Add ability to import/export Lemmy account settings

micahmo:
## Pull Request Description

In an effort to bring Thunder up to parity with recent Lemmy features, this PR adds support for importing and exporting Lemmy account settings. The experience should roughly match the web UI (settings page), including the auto-generated filename.

image

Notes
- The import/export endpoints seem to be very heavily rate-limited, so it's a little hard to test a bunch!
- There is error handling for file operations as well as Lemmy exceptions.
	- However, Lemmy doesn't seem to report if an invalid file is uploaded. But by restricting the file extension to `.json` and parsing the file in Thunder, we can at least ensure that's valid JSON. Beyond that, the user can uplaod pretty much any JSON file and Lemmy will accept it, so we'll just have to keep that in mind.

## Issue Being Fixed

Issue Number: #762

## Screenshots / Recordings

https://github.com/user-attachments/assets/ca327d34-0fdd-42fe-b822-eb5e64978f93

---

NOTE! Although everything appears to work, the import doesn't actually take effect. Lemmy returns `success: true`, but none of the settings change. I can take the `.json` file generated by exporting the settings through Thunder, and upload that to the web UI, and it works fine. So I think the export is fine, but I'm not sure what's going on with the import. Any help would be appreciated! I'll leave this as a draft until we can get it working.

**Update**: @hjiangsu discovered that we need to force the data to be passed as a `String` instead of encoded back into JSON. This needs to be done in the API layer, so hopefully there will be a corresponding change coming soon!

---

Thanks to @hjiangsu for solving the API problem! See this commit: https://github.com/thunder-app/lemmy_api_client/commit/6d0e593503abd69e4fa6b2389712ba2031d4b47f

Here's a new demo video showing that the import works for subscribed communities and account settings.

https://github.com/user-attachments/assets/103d473b-5b8a-41fa-b952-1150d62f65af

I'm going to mark this PR as ready, although it will need one more update to the API package once that's merged in.


micahmo:
Just updated the package reference based on thunder-app/lemmy_api_client#29, so this should be ready to go!

lib/l10n/app_en.arb
```diff
@@ -877,6 +901,10 @@
   "@exportDatabaseSubtitle": {
     "description": "Subtitle of setting for exporting the db"
   },
+  "exportLemmyAccountSettingsDescription": "Export your Lemmy account settings as JSON",
```
hjiangsu, line 904: I would just simplify the wording here to "Export your Lemmy account settings" (users may not understand what JSON is)

lib/l10n/app_en.arb
```diff
@@ -1071,14 +1099,26 @@
...
+  },
   "importExportSettings": "Import/Export Settings",
   "@importExportSettings": {
     "description": "Category for settings related to import and export."
   },
+  "importExportThunderSettings": "Import/Export Thunder Settings",
+  "@importExportThunderSettings": {
+    "description": "This is specifically used in search results to distinguish from Lemmy import/export"
+  },
+  "importLemmyAccountSettingsDescription": "Import your Lemmy account settings from JSON",
```
hjiangsu, line 1118: Same as before, I would omit the JSON part of the message!

lib/user/pages/user_settings_page.dart
```diff
@@ -455,6 +459,109 @@ class _UserSettingsPageState extends State<UserSettingsPage> {
...
+                              Padding(
+                                padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 8.0),
+                                child: Text(l10n.importExportSettings, style: theme.textTheme.titleMedium),
+                              ),
+                              SettingsListTile(
+                                icon: Icons.file_download_rounded,
+                                description: l10n.exportLemmyAccountSettingsDescription,
+                                widget: const SizedBox(height: 42.0, child: Icon(Icons.chevron_right_rounded)),
+                                onTap: () async {
+                                  dynamic exportSettings;
```
hjiangsu, line 472: Can we specify the type here instead of having it as a `dynamic` type?

hjiangsu (commented):
Just a few minor comments, otherwise LGTM!

As an aside, the CI workflow seemed to have failed here - it seems to be related to `freezed_annotation`?

micahmo:
> As an aside, the CI workflow seemed to have failed here - it seems to be related to `freezed_annotation`?

Yeah, I think I've seen this before, and I think I fixed it adding the `freezed` package. I'm not sure if it got removed at some point. But anyway, I added it back and that seemed to fix the build.

lib/l10n/app_en.arb
micahmo: I am certainly willing to change the wording, but I did want to mention (as in the original PR description) that I was just copying this from the Lemmy web UI.

395274054-069661bd-7d53-449f-98b2-aff7e71cfedc

I agree that JSON is a more technical term, but exporting your account settings is also somewhat technical. For people who don't care or don't know what it means, it doesn't hurt to say. But for those who do care, I think it's useful info.

But again I'm happy to get either way. Let me know what you think!

lib/user/pages/user_settings_page.dart
micahmo: I actually don't think I can change the type here because `dynamic` is what's returned by the `ExportSettings` API call (it implements `LemmyApiQuery<dynamic>`).

https://github.com/thunder-app/lemmy_api_client/blob/f352a17afad3d7a4fa98e1711a4886d1fa3929ae/lib/src/v3/api/user/user.dart#L105

micahmo:
@hjiangsu I fixed your build and responded to your comments; let me know what you think!

lib/l10n/app_en.arb
hjiangsu: > I agree that JSON is a more technical term, but exporting your account settings is also somewhat technical. For people who don't care or don't know what it means, it doesn't hurt to say. But for those who do care, I think it's useful info.

Hmm, I'm just contrasting this with our own import/export settings (which also saves app settings as a json file, but is not explicitly mentioned)

image

I would argue that it would be more important to know what contents are being saved, rather than the format of the file! For example, we could do something like this:

Title: "Export Lemmy account settings"
Subtitle: "Includes subscribed communities, blocklists, and account preferences"

If we want to add in the file format, then maybe placing it in the subtitle would make more sense:

Title: "Export Lemmy account settings"
Subtitle: "Includes subscribed communities, blocklists, and account preferences (as JSON)"

Thoughts?

lib/user/pages/user_settings_page.dart
hjiangsu: Alright, sounds good!

lib/l10n/app_en.arb
micahmo: I'm totally fine with this! Although, as I was adding a subtitle to the setting, it looked funny duplicated on both export and import, but it also looked funny on only one, so I added it as a subtitle to the section heading. What do you think?

| image |
| - |

I pushed this, so if you like it, you can merge.

hjiangsu: Looks great! Thanks for the changes

hjiangsu (approved):

PR #1639, opened 2024-12-13
Fix an issue with comment scrolling offset

micahmo:
## Pull Request Description

This is a quick fix related to comment scrolling in the new post page. It looks like my change in #1630 inadvertently allowed the comments to be scrolled to the very top of the display, due to the extended `SafeArea`, which is now partly covered by a `Container`. This fix adds an offset to `animateToItem` to account for the status bar.

However, this turned out to be tricky, because as a child widget of `SafeArea(top: false)`, accessing the status bar height with `MediaQuery.of(context).padding.top` always returns `0`. Therefore I had to capture the height outside the `SafeArea` and pass it in. I also had to remove the offset when the top bar is pinned. Now it works well!

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/user-attachments/assets/fc7cb99e-e27d-443a-8601-1376ef846ec2

#### After

https://github.com/user-attachments/assets/bb34e116-7549-48f4-b229-3154cd9cbd39

## Checklist

### DIFF
--- a/lib/post/pages/legacy_post_page.dart
+++ b/lib/post/pages/legacy_post_page.dart
@@ -99,6 +99,8 @@ class _PostPageState extends State<PostPage> {
     final theme = Theme.of(context);
     final ThunderState thunderState = context.watch<ThunderBloc>().state;
     final AppLocalizations l10n = AppLocalizations.of(context)!;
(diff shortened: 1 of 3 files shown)


hjiangsu (approved):
LGTM!

PR #1640, opened 2024-12-13
Add the ability to specify alt text for image posts

micahmo:
## Pull Request Description

This PR adds the ability to specify alternative text for images when creating an image post.

Notes:
- I was anticipating the need to add a way for optional fields to be collapsible on the create post page. However, as I was testing this feature, it occurred to me that if the URL field is populated with an image, which is the only case in which we show the alt text field, then we don't also show the custom thumbnail field. So we will only ever show one or the other, and as a result, the alt text field does not occupy any additional space on the create post page. We still may want to make some fields collapsible in the future, but I don't think it needs to be right now.
- I verified that drafts as well as edits work properly with the new field. Do note that the post page needs to be refreshed to see the new alt text. I think it would be a separate PR if we wanted to "fix" that.
- Along with [#1323](https://github.com/thunder-app/thunder/pull/1323) and [#1638](https://github.com/thunder-app/thunder/pull/1638), this should be the last change needed to bring Thunder in parity with new Lemmy server features added in 0.19.4 (or at least all the ones we planned on adding)!
- I'm not sure why there are so many changes to `database.g.dart`. Below are my steps for updating the database. Let me know if I did anything wrong! Also, I don't think we've officially documented how to update the db since [#1501](https://github.com/thunder-app/thunder/pull/1501), so if what I did is anything close to "right", maybe we should add it to the README. :blush:

---

1. Update schema version in `lib\core\database\database.dart`.
2. Add new tables/columns in `lib\core\database\tables.dart`.
3. Export schema
	- `dart run drift_dev schema dump lib/core/database/database.dart drift_schemas/`
4. Update migrations
    -  `dart run drift_dev schema steps drift_schemas/ lib/core/database/schema_versions.dart`
5. Go back to `lib\core\database\database.dart` and implement the migration function
6. Update the auto-generated db table models
	- `dart run build_runner build`
7. Reformat the auto-generated files
    - `dart format --set-exit-if-changed -l 200 lib`

## Issue Being Fixed

Issue Number: #1447

## Screenshots / Recordings

https://github.com/user-attachments/assets/a84dc5b1-cf11-4c1a-b92c-404ac4876ba9

## Checklist


hjiangsu (approved):
LGTM, and thanks for implementing the starting flow for db downgrade strategies!

> Below are my steps for updating the database. Let me know if I did anything wrong!

I believe all that looks good. I think the reason for all the changes in `database.g.dart` is likely due to version upgrades for `drift`. I wouldn't worry too much about all those changes as long as everything seems to work.

As you mentioned, we can definitely add the steps needed to generate the step-by-step migrations to the README!

PR #1641, opened 2024-12-16
Hide profile/hamburger during feed refresh

micahmo:
## Pull Request Description

I noticed that when you refresh the feed, everything disappears (including all app bar action widgets) except the hamburger menu button (or the profile picture, depending on the settings). I think it looks a little funny, especially when it's the profile. Sticks out like a sore thumb. So I thought I'd just make a super minor change so that whatever icon is in the top-left also disappears during a reload.

Please review without whitespace.

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/user-attachments/assets/24e5839a-c09d-4652-9dc2-1a16eb7b4cf7

#### After

https://github.com/user-attachments/assets/44e79213-30be-4e7c-93b4-d2efd40c3f6a

## Checklist


hjiangsu (approved):
LGTM!

PR #1642, opened 2024-12-17
Oauth updates

gwbischof:
Oauth updates.
This PR depends on https://github.com/thunder-app/lemmy_api_client/pull/28
This can be tested against the lemmy instance at hopandzip.com which I deployed with the lemmy:main


pubspec.yaml
```diff
@@ -13,9 +13,10 @@ dependencies:
   push:
     path: packages/push/push
   lemmy_api_client:
-    git:
-      url: "https://github.com/thunder-app/lemmy_api_client.git"
-      ref: 874005c9013de53f526562f882fe8231eb0ce5ae
+    path: ../lemmy_api_client
+    #git:
+    #  url: "https://github.com/thunder-app/lemmy_api_client.git"
+    #  ref: 16d14a1c13ac9522e85188ad9cf23d8912ec8fee
```
gwbischof, line 19: Should merge this first https://github.com/thunder-app/lemmy_api_client/pull/28
And then update this.

lib/core/auth/bloc/auth_bloc.dart
```diff
@@ -184,8 +188,229 @@ class AuthBloc extends Bloc<AuthEvent, AuthState> {
...
+    on<OAuthLoginAttempt>((event, emit) async {
+      LemmyClient lemmyClient = LemmyClient.instance;
+      String originalBaseUrl = lemmyClient.lemmyApiV3.host;
+      String instance = event.instance ?? state.oauthInstance!;
+      ProviderView provider = event.provider ?? state.oauthProvider!;
+
+      try {
+        emit(state.copyWith(status: AuthStatus.loading, account: null, isLoggedIn: false));
+
+        if (instance.startsWith('https://')) instance = instance.replaceAll('https://', '');
```
micahmo, line 201: You can do `replaceFirst` here to make sure you only replace it at the beginning (not that a scheme should ever appear elsewhere in a URL!).

lib/core/auth/bloc/auth_bloc.dart
```diff
@@ -184,8 +188,229 @@ class AuthBloc extends Bloc<AuthEvent, AuthState> {
...
+        lemmyClient.changeBaseUrl(instance);
+
+        // Build oauth provider url.
+        var authorizationEndpoint = Uri.parse(provider.authorizationEndpoint);
+        String redirectUri = "https://thunderapp.dev/oauth/callback"; // This must end in /oauth/callback.
```
micahmo, line 206: Do you want to maybe pull this URL out into the globals file or at least define it as a constant in this file? That also lets you reuse it in part 2.

lib/shared/input_dialogs.dart
```diff
@@ -527,3 +527,78 @@ void showInputDialog<T>({
     }),
   );
 }
+
+/// Shows a dialog which takes input and offers suggestions
+Future<String?> showBlockingInputDialog<T>({
```
micahmo, line 532: This seems pretty similar to `showInputDialog`. Is there any way to combine these?

pubspec.yaml
micahmo: For now can you update this to point at your fork/branch? That will make building and testing easier. Then we can point it back to `thunder-app/lemmy_api_client` before merging.

micahmo (commented):
Hey, I just wanted to say, nice going on this!! I know it was a beast to get working, but in the end the code doesn't seem too crazy!

I left a few comments, mostly stylistic things. I haven't actually tested the flow yet but I would love to!

lib/shared/input_dialogs.dart
gwbischof: `showInputDialog` doesn't block, so it didn't work to display the dialog where you enter your username on your first oauth login. So I modified it to create `showBlockingInputDialog`. So yea, they are pretty similar.  I think `showInputDialog` is used in one other place in the code, maybe we could make that one blocking also?

lib/core/auth/bloc/auth_bloc.dart
gwbischof: updated, thanks :)

lib/core/auth/bloc/auth_bloc.dart
gwbischof: I made it a constant at the top of this file.

lib/account/pages/login_page.dart
```diff
@@ -176,6 +180,54 @@ class _LoginPageState extends State<LoginPage> with SingleTickerProviderStateMix
...
+                  context.read<AuthBloc>().add(const CancelLoginAttempt());
+                }
+              }
+            } else if (state.status == AuthStatus.oauthSignUp) {
+              bool completedSignUp = false;
+              String? username;
+
+              await showBlockingInputDialog<String>(
+                  context: context,
+                  title: "Sign Up",
```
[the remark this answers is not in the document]
gwbischof: Which file do I update to do that?

lib/account/pages/login_page.dart
```diff
@@ -176,6 +180,54 @@ class _LoginPageState extends State<LoginPage> with SingleTickerProviderStateMix
                   context.read<AuthBloc>().add(const CancelLoginAttempt());
                 }
               }
+            } else if (state.status == AuthStatus.oauthContentWarning) {
+              bool acceptedContentWarning = false;
+
+              await showThunderDialog<void>(
```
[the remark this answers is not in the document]
gwbischof: yep, it's the same content warning code.  The content warning is already is a reusable method.  I can update the code to have less code duplication.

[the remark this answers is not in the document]
gwbischof: It updated it, but I had to nest another if statement.

lib/account/pages/login_page.dart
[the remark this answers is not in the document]
gwbischof: Updated

micahmo (approved):
Nice, the code looks good now! I will try to test if/when I get the chance. If everything is working for you and @hjiangsu, don't worry about waiting for me!

lib/shared/input_dialogs.dart
gwbischof: By blocking I mean `showBlockingInputDialog` will wait for the user to click `accept` before running the code that follows it. So you can await `showBlockingInputDialog` and that doesn't work with `showInputDialog`

https://github.com/thunder-app/thunder/pull/1642/files#diff-4144a1c5223a9dd35fda6e4094acae17e6f1d3f1886b50f47977d043c54f5317R192

gwbischof: I will investigate and see if I can parameterize `showInputDialog`

micahmo: Ahh ok I gotcha! In that case, I think it makes sense to transition `showInputDialog` into a `await`able method. Existing code that doesn't care to `await` can just not do that and let it run in the background. Let me know what you think.

gwbischof: Yea, I think that makes sense. I can do that

micahmo:
I just pushed a small change which adds a heading above the OAuth providers (to make it clear what they are and that they're an alternative to non-SSO login), and I also added a little space between each provider.

| image |
| - |

hjiangsu:
> Nice, the code looks good now! I will try to test if/when I get the chance. If everything is working for you and @hjiangsu, don't worry about waiting for me!

Sounds good, the changes look good to me after a quick review, and I've already tested a previous implementation of this! However, I'm going to hold off on merging this in until v0.20.0 is closer to release, or at least until Lemmy's API changes are more stable (at the moment, it looks like there's still a lot of major changes to the API that we'll need to support eventually). 

This way, we can re-test the implementation closer to the next major release!

PR #1644, opened 2024-12-18
Add app-link support for thunderapp.dev

hjiangsu:
## Pull Request Description

This PR adds support for Android's app links - specifically, for https://thunderapp.dev. Note that this will only work for the release builds (GitHub and Google Play) since their release key SHA-256 fingerprints have been added to `assetlinks.json`.

@gwbischof If you would like to test this locally, you'll have to provide me the SHA-256 fingerprints from your debug signing keys. See https://developer.android.com/training/app-links/verify-android-applinks#web-assoc on how to obtain the fingerprints.

Once you provide this to me, I can temporarily add the fingerprints to `assetlinks.json` to enable app links for your debug builds!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Nice!

hjiangsu:
I've updated this to also include support for iOS universal links!

This is equivalent to the app links on Android. The one difference here is that you'll have to navigate to https://thunderapp.dev/oauth/ in Safari to see the option to navigate back to Thunder.

@gwbischof If you want to test deep linking using iOS, you'll likely have to provide me the team ID so that I can add that into the `apple-app-site-associations` file on https://thunderapp.dev. Otherwise, only release builds will be associated with the universal links. To find this information, see https://developer.apple.com/help/account/manage-your-team/locate-your-team-id/

hjiangsu:
One more update, this PR also includes support for universal links on macOS!

Note that these implementations just allow us to open up Thunder from https://thunderapp.dev - there was no implementation change to handle the deep links (so this will likely result in "This type of link is not supported at the moment" message.)

hjiangsu:
@micahmo I've merged this into `develop`, but let me know if you notice any issues with the commits after you previously approved it!

lib/thunder/pages/thunder_page.dart
```diff
@@ -109,9 +109,11 @@ class _ThunderState extends State<Thunder> {
     }
 
     WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-      handleSharedFilesAndText();
-
       if (!kIsWeb && (Platform.isAndroid || Platform.isIOS)) {
+        handleSharedFilesAndText();
+      }
+
+      if (!kIsWeb && (Platform.isAndroid || Platform.isIOS || Platform.isMacOS)) {
```
micahmo, line 116: I'm not sure if we want to `handleNotifications` for `MacOS`.

PR #1525, opened 2024-08-12
Highlight last tapped post

micahmo:
## Pull Request Description

This PR sort of goes along with #1520. It's a separate feature, but it's something that I realized is even more important once we have the ability to navigate out and back into posts. It introduces a small highlight color to the last selected post. This is super handy when you navigate out to check the title or media, for example, and then you want to go back into the comments. This lets you see what post you were previously looking at very easily. It's also handy when you accidentally go into a post and want to mark it as unread.

Notes
* All of the other contents respect the read status, so it is still easy to tell whether a post is read or unread, even when it has this highlight.
* The highlighting goes away when refreshing the feed.

> This is another feature inspired by Relay, and I tried to match the implementaiton.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/user-attachments/assets/e4f373e0-eee2-4161-b0d8-9751ecd51b47
## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1648, opened 2024-12-23
Highlight the last tapped post after navigating via image

micahmo:
## Pull Request Description

This PR is a small followup to #1525. I noticed that if you open an image, and then navigate to the comments via the button, the post does not get highlighted as the last viewed post in the feed. I tested Relay (which is what I based this feature on), and it does highlight the post in this scenario, so I decided to add the same thing here.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/user-attachments/assets/fa84aa9d-0d1c-42a9-822a-ae5aaf530553

## Checklist


hjiangsu (approved):
LGTM!

PR #1649, opened 2024-12-28
Upgrade dependencies for Flutter 3.27

hjiangsu:
## Pull Request Description

This PR upgrades our packages and dependencies to be compatible with Flutter 3.27! To read more about the changes in 3.27, see https://medium.com/flutter/whats-new-in-flutter-3-27-28341129570c

One notable change here is that Flutter 3.27 now uses Impeller as the default rendering engine on Android, which should hopefully improve the overall smoothness and experience on Android devices!

I'll also have a follow-up PR to update our CI to use 3.27 once this is merged in

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1652, opened 2025-01-01
Fix icon sizes

micahmo:
## Pull Request Description

This is a super small PR that fixes the issue mentioned in this comment: https://github.com/thunder-app/thunder/pull/1649#issuecomment-2565984092

As to _why_ the issue suddenly happened after the latest Flutter upgrade, I'm not sure. But by specifying a size (rather than assuming/hoping it'll get the right size automatically), the problem seems to be fixed. This is probably a good thing to be doing anyway!


hjiangsu (approved):
LGTM, thanks for finding and fixing the issue!

PR #1657, opened 2025-01-10
Clean up image permissions for Android

hjiangsu:
## Pull Request Description

This PR removes the following permissions from `AndroidManifest`:
- `READ_EXTERNAL_STORAGE`
- `READ_MEDIA_IMAGES`
- `READ_MEDIA_VIDEO`

The main reason for this is to comply with some Google Play Store requirements. Here is an article containing more information about this: https://support.google.com/googleplay/android-developer/answer/14115180

As a result, I've also updated the way permissions are requested when saving images in the image viewer. We no longer use `permission_handler`, but rather handle permissions through the `gal` package. Additionally, I've removed the `device_info_plus` dependency since it looks like it's no longer needed.

All in all, this reduces some of our permission logic, and cleans up some of our dependencies!

@micahmo I did some testing on Android emulators (specifically, API versions 35 and 29), so it _should_ be working as expected. However, please let me know if you encounter any issues saving or uploading images!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -2,10 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" 
           xmlns:tools="http://schemas.android.com/tools">
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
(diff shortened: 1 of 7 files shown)


ggichure (approved):

micahmo (approved):
LGTM, I'll let you know if I run into any issues with these changes!

PR #1671, opened 2025-01-22
Show progress indicator during post refresh

micahmo:
## Pull Request Description

This PR fixes a small quirk of the experimental/new post page that has been "bothering" me for a while. The quirk is that while the comments are loading, the spinner goes away before the comments are fully loaded and there is a brief pause. I've tried to capture it in the screen recording below, but the pause is much "worse" when you're on a slower network (which is what implied to me that there was still something loading, as opposed to just being a visual/layout glitch).

Turns out all I had to do was handle an additional state, and it looks much better now!

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/user-attachments/assets/9e5af36b-930e-43b0-b578-f8930d4e665e

### After

https://github.com/user-attachments/assets/325917d4-e6cf-415d-8f59-ab375e64d6ac

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -201,7 +201,7 @@ class _PostPageState extends State<PostPage> {
                           viewSource: viewSource,
                         ),
                       ),
-                      if (state.status == PostStatus.loading)
+                      if (state.status == PostStatus.loading || state.status == PostStatus.refreshing)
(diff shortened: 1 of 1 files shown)


ggichure (approved):

hjiangsu (approved):
LGTM!

Issue #178, opened 2023-07-02
Handshake error in client (Os error CERTIFICATE_VERIFY_FAILED: certificate has expired (handshake.cc:393))


fantoche7:
**Description**  
unable to connect , I beleive It has something to do with flutter on older android devices

**Expected Behavior**  
be able to connect 

**Device & App Version:**  
 - OS: android
 - Version: 5.1

hjiangsu:
Could you provide me with some more details about this issue? That'll help me figure out the root cause better!

For example:
- OS version
- Instance you're trying to connect to
- Are you logged in, or does it simply show when you start the app

fantoche7:
OS : android 5.1
The error show at startup ,and when ere trying to do something, like searching, I bet the app can't make any connetion to the internet.

ddCeka:
I have the same problem.
OS: Android 10

The problem(handshake not resolved) shows when I tried to use the app routing it with adguard. While I excluded it from adguard, it works as usual. Look like adguard certificate causing this?

hjiangsu:
Hmm, interesting - I'm not too sure what it could be. Do you have Sentry enabled in the settings (this is only because you mentioned adguard which might be blocking Sentry)? Try to disable Sentry if its enabled and see if that changes anything

The certificate errors usually seem to indicate that there was an issue with checking for the instance's credentials (through HTTPS)

Could you let me know the instances that you've tried to connect to? I can try to see if I can replicate it on my end

ddCeka:
> Oh hmm, I have to go through an application process in order to sign up with your instance. I have tried to set the default instance temporarily to `lemmy.dbzer0.com` and have not yet encountered an issue (at least on the iPhone simulator)
> 
> ![simulator_screenshot_4293345F-1069-4B10-B4D4-110D10C1B148](https://user-images.githubusercontent.com/30667958/250582448-beaeb94b-f334-473f-849d-5bf302332af7.png)

Yeah, at the moment the only way I do just disabling HTTPS filtering on adguard for this and work just fine

hjiangsu:
I'm looking into this: https://adguard.com/kb/general/https-filtering/what-is-https-filtering/#https-filtering-issues-on-android-7

It might be possible for Thunder to trust user-installed certificates. I would have to take a deeper look into it but this might be the root cause of the issue

hjiangsu:
[...]

PR #1675, opened 2025-01-24
Fix certificate issue with LetsEncrypt for older Android versions

hjiangsu:
## Pull Request Description

This PR adds a fix to #178 with older Android versions accessing instances using LetsEncrypt for SSL. To fix this, the new certificate was manually added and loaded at startup.

You can cross-verify the authenticity of the certificate by visiting: https://letsencrypt.org/certificates/ under ISRG Root X1. This is where I obtained the certificate: https://letsencrypt.org/certs/isrgrootx1.pem

image

## Issue Being Fixed

Issue Number: #178

## Checklist


micahmo (approved):
LGTM! I assume this root cert would come installed by default on newer devices/OS? Then it should be pretty safe to do this. I'm also assuming this would not bypass any other standard certificate validation (like expiration).

hjiangsu:
> I assume this root cert would come installed by default on newer devices/OS?

I believe so - the following screenshot was taken from the latest Android Emulator (API 35). We can cross-reference the system-installed certificate with the certificate that we add in this PR by comparing their SHA fingerprints:

**Android System Certificate**
Screenshot_1737745164

**LetsEncrypt Certificate Information** (https://crt.sh/?id=9314791)
image

The crt.sh link can be accessed by visiting: https://letsencrypt.org/certificates/ under ISRG Root X1 and selecting the links under `Certificate details (self-signed)`
image

PR #1674, opened 2025-01-24
Revert "Show progress indicator during post refresh"

hjiangsu:
Reverts thunder-app/thunder#1671

PR #1680, opened 2025-02-05
Show progress indicator during post refresh

micahmo:
This is a followup to #1671 and #1674. See further explanation and video demo there. I've been daily driving this and it seems to fix the issue just fine!


hjiangsu (approved):
LGTM!

Issue #1684, opened 2025-02-07
Keyboard pops up when going back to main screen from post


Asplundh:
### Bug Description

When going back from a post (or a settings category) to the main screen, the keyboard pops up if the search function has been used since app start-up.

### Expected Behaviour

Go back to the main screen without having the keyboard pop up.

### Steps to Reproduce

1. Open search tab and tap the search input field.
2. Open a post from feed, account or inbox tabs (or open a category from within settings tab).
3. Hit back button or go back using gesture navigation to return to the main screen.

### App Version

0.6.0

### Device

Pixel 7

### OS

GrapheneOS 2025020500 (Android 15)

micahmo:
I believe this is a dup of #1110.

Issue #1110, opened 2024-02-02
Search input field never gets unfocused, keeps bringing up keyboard


fmbarina:
**Description**  
Search text input field doesn't get automatically unfocused and keeps bringing up keyboard in other screens. So far, pressing enter on the search field is the only way I've found to fix this behavior without needing to close and reopen the app. It's easy to forget to do that when search results appear as you type, though.

**How to Reproduce**
1. Go to the search tab
2. Focus the text input field, but don't press enter
3. Go to any other screen and start interacting with it
  - Open and then close community, upon returning, keyboard pops up
  - Press post thumbnail to open image, upon closing it, keyboard pops up
  - Select settings category (e.g. 'Gestures'), upon returning, keyboard pops up
  - Change accounts by long pressing the button on the navbar, keyboard pops up
  - Open sidebar, then close it, keyboards pops up

**Expected Behavior**  
The search field shouldn't bring up the keyboard when you're not on the search screen. It should probably be unfocused when the user goes to another screen or interacts with any search results.

**Device & App Version:**  
 - Device: Asus Zenfone 3 Zoom 
 - OS: Android 8
 - App Version 0.2.8

**Additional Context**

Random idea from someone trying to learn Flutter: maybe unfocusing the search field at `onDestinationSelected` in `bottom_nav_bar.dart` could help? No need to answer, and I defer to your experience. Might try it later.

May have found another bug while testing this, unsure if related.

Note that even while browsing search results, the keyboard is brought up every time you return to them. Specifically when searching posts, however, tapping a thumbnail while the keyboard is still open erases the search field and clears all results.

hjiangsu:
Thanks for the detailed report!

> maybe unfocusing the search field at onDestinationSelected in bottom_nav_bar.dart could help? 

If you'd like to give it a try, by all means go ahead! I think thats a reasonable approach to solving this issue.

hjiangsu:
As mentioned in https://github.com/thunder-app/thunder/pull/1686#issuecomment-2643494602, I'll keep this issue up until all other Search related issues are fixed!

PR #1686, opened 2025-02-07
Fix unintended keyboard popup when navigating back to main screen

hjiangsu:
## Pull Request Description

This PR fixes an issue described in #1684 and #1110 where the keyboard would show up on the main pages after using search. I mainly tested this on an Android emulator as iOS doesn't seem to exhibit this issue (difficult to reproduce)

Note: this fix still keeps the old behaviour of double-tapping the search icon to open up the keyboard!

## Issue Being Fixed

Issue Number: #1684, #1110

## Screenshots / Recordings

Original Issue

https://github.com/user-attachments/assets/50cf04c2-54f3-4efb-987b-4c1fd4a18497

Fixed Issue

https://github.com/user-attachments/assets/66039da6-0565-47a6-a757-fcf166769f59

## Checklist

### DIFF
--- a/lib/thunder/widgets/bottom_nav_bar.dart
+++ b/lib/thunder/widgets/bottom_nav_bar.dart
@@ -141,6 +141,10 @@ class _CustomBottomNavigationBarState extends State<CustomBottomNavigationBar> {
             context.read<ThunderBloc>().add(const OnFabToggle(false));
           }
 
+          if (widget.selectedPageIndex == 1 && index != 1) {
+            FocusManager.instance.primaryFocus?.unfocus();
+          }
+
           if (widget.selectedPageIndex == 0 && index == 0) {
             context.read<FeedBloc>().add(ScrollToTopEvent());
           }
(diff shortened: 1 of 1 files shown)


micahmo:
LGTM!

BTW, to fully fix #1110, I believe we also need to minimize the keyboard when we navigate away from the search page in any way for example, as mentioned in that issue, clicking a search result. This behavior is a little annoying.

- Search for something (do not minimize keyboard)
- Click a result (keyboard automatically goes away)
- Go back
  - Keyboard automatically comes back :(

I believe there are some more scenarios mentioned in the bug. But that can be separate. I think this fully fixes #1684.

micahmo (approved):

hjiangsu:
> I believe we also need to minimize the keyboard when we navigate away from the search page in any way

Ahh I see - this can definitely be fixed in a separate PR then as it's a bit more involved! I'll keep https://github.com/thunder-app/thunder/issues/1110 open until all the issues related to the Search page are resolved.

PR #1690, opened 2025-02-07
Improve UX for toggling pure black theme

hjiangsu:
## Pull Request Description

This PR applies the feedback from https://github.com/thunder-app/thunder/pull/1605#issuecomment-2543364403 to improve the UX for enabling and disabling pure black theme. Here's the new flow:
- Theme only contains light, dark and system options
- When the theme is set to dark or system, an option to enable pure black theme will be shown
- When the pure black theme is enabled, dark themes will use the pure black themes (applies to dark mode, and system dark mode)

I've also added some migration steps for those who are currently using the pure black theme directly! I've tested most if not all possible variations, and also tested the migration steps.

**Note:** I've kept `setting_theme_system_pure_black` and `systemThemePureBlack` the same to prevent issues with users who were previously using that setting. I have however, renamed the variables to match the new context.

## Issue Being Fixed

Issue Number: #1598

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -199,7 +199,7 @@ enum LocalSettings {
   /// -------------------------- Theme Related Settings --------------------------
   // Theme Settings
   appTheme(name: 'setting_theme_app_theme', key: 'theme', category: LocalSettingsCategories.theming, subCategory: LocalSettingsSubCategories.theme),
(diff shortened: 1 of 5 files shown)


micahmo:
The code LGTM!

The only comment I have is about searching. When I search for "dark" I get a result called "System Dark Mode", but clicking that search result brings me to the new "Pure Black" toggle. I think we probably just need to change the resource string for that setting to "Pure Black".

https://github.com/user-attachments/assets/605b1008-d213-46a3-950c-30df7186b873

micahmo (approved):

hjiangsu:
> I think we probably just need to change the resource string for that setting to "Pure Black".

Done!

PR #1693, opened 2025-02-11
Fix color of block icon on block user button

micahmo:
## Pull Request Description

This is a small PR which fixes another color bug that I just discovered. Previously the icon associated with the Block User button was red, but now it's the default theme color. Turns out that this is a known breaking change where the button style no longer applies to icons. (https://github.com/flutter/flutter/issues/154644)

Also note that it looks like the size had also changed; however, I don't hate the new size, so I left that alone.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

| Old | New | Fixed |
|--------|--------|--------|
| image | image | image | 

## Checklist


hjiangsu (approved):
LGTM!

PR #1701, opened 2025-02-17
Fix profile modal staying open after logging into new account

micahmo:
## Pull Request Description

This is a quick PR which steals a fix that I put into @gwbischof's PR #1642. The problem that I fixed was not introduced in his PR, and it's been bothering my in daily usage lol! So if his PR isn't going to be merged soon, I'd like to get this particular fix in sooner. It just ensures that the profile modal closes itself after you switch accounts.

## Issue Being Fixed

Issue Number: Cherry-picked d8720e41512654ec23ff1934dc03be361cd8c735

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1706, opened 2025-02-18
Fix profile modal not closing after switching accounts

micahmo:
## Pull Request Description

This is a quick followup to #1701. For some reason, even though the modal was closing after adding a new account, it wasn't closing after switching accounts. I found that adding `rootNavigator: true` helped. To be honest I'm not sure exactly why (maybe it was a breaking change from a Flutter upgrade, since I only noticed it recently). But this seems to do the trick.


hjiangsu (approved):
LGTM! I also noticed this issue and it was bugging me as well 😅

PR #1707, opened 2025-02-19
Consolidate all navigation together

hjiangsu:
## Pull Request Description

This PR attempts to consolidate all our existing methods of navigating to specific pages (posts/comments/modlog, etc) into one centralized file (`navigation.dart`). The reason for this is to help reduce the fragmentation of navigation logic, and to hopefully make navigation more consistent.

In addition to this, there are a few more changes including:
- Updating `flex_color_scheme` to 8.1.1. This update applies a hotfix that fixes https://github.com/thunder-app/thunder/pull/1693 on a more general basis
- Updating the settings navigation logic to also include support for sub-pages (e.g., Account -> Discussion Languages, Account -> Media Management, Appearance -> Posts, etc.).
  - This was mainly done to consolidate the settings-related navigation logic, but includes a side benefit where you can directly search for a sub-page (e.g., searching for Manage Media allows you to go straight to the Manage Media page rather than the Account page)

I tested most of the navigation logic, however, since there are so many ways of navigating to different pages, there might be some edge cases that I may have missed! I think the easiest way to know is to merge this in early and daily it to see if we encounter any issues.

There may be some follow up PRs for this to further consolidate and clean up the individual navigation functions!

## Issue Being Fixed

Issue Number: N/A

## Checklist


lib/core/theme/bloc/theme_bloc.dart
```diff
@@ -43,7 +43,7 @@ class ThemeBloc extends Bloc<ThemeEvent, ThemeState> {
       bool usePureBlackTheme = prefs.getBool(LocalSettings.usePureBlackTheme.name) ?? false;
       if (usePureBlackTheme && (themeType == ThemeType.dark || (themeType == ThemeType.system && brightness == Brightness.dark))) themeType = ThemeType.pureBlack;
 
-      bool useDarkTheme = themeType == ThemeType.dark || themeType == ThemeType.pureBlack;
+      bool useDarkTheme = themeType == ThemeType.dark || themeType == ThemeType.pureBlack || (themeType == ThemeType.system && brightness == Brightness.dark);
```
micahmo, line 46: I feel like we make a similar check in a lot of places. Maybe we can pull this out into a helper method too.

lib/feed/widgets/feed_page_app_bar.dart
```diff
@@ -240,35 +237,14 @@ class FeedAppBarCommunityActions extends StatelessWidget {
...
-                          // Create a new SearchBloc so it doesn't conflict with the main one
-                          BlocProvider.value(value: searchBloc),
-                          BlocProvider.value(value: thunderBloc),
-                        ],
-                        child: SearchPage(communityToSearch: feedBloc.state.fullCommunityView!.communityView, isInitiallyFocused: true),
-                      ),
-                    ),
-                  );
-                },
+                onTap: () async => navigateToSearchPage(context),
```
micahmo: Does this need to be `async`?

micahmo (approved):
LGTM! We definitely had a lot of navigation stuff all over the place, and a lot of copy-pasting. I like the consolidation. I just did a quick review a left a couple small comments, but otherwise I think we can get this in and start testing it.

lib/feed/widgets/feed_page_app_bar.dart
hjiangsu: Done - also removed a couple other instances where `async` was still present!

lib/core/theme/bloc/theme_bloc.dart
hjiangsu: I just checked the codebase, and pretty much all usage of `useDarkTheme` references the ThemeBloc so we might not need any helper functions for this.

There was one case where we were calculating the `useDarkTheme`, but that turned out to be in a widget that we no longer use, so I removed it! See https://github.com/thunder-app/thunder/pull/1707/commits/0c25ca674da1aff973e26c15a1361586dd598696

Issue #1417, opened 2024-06-04
Smart image sizes in card view


K4LCIFER:
### Is Your Feature Request Related to a Problem?

Currently, there are only two options for image sizing when using card view: full-height images, or all images are cropped to the same height. Both cases can result in troublesome behavior. Full-height images can result in images appearing partly off screen so one has to scroll to view them (this can be quite unpredictable if the images are very tall). Cropping all images results in the majority of images being cropped, so one has to tap on them to view them in any realistic manner.

### Feature Description

I propose a sort of smart resizing to the images dependent on screen height. If an image is smaller than the screen height + the padding of the post (e.g. title, user, community, etc.) then show the full image, but if the image is larger than the screen height plus padding, then crop it to fit the screen.

Cris-Edmundson:
I know on some reddit apps if the image was too tall and you had the app set to show the whole image, they would show the whole image but make it small so the image would still be a reasonable total height, making it clear the user can view open the image viewer and scroll if they wanna real the whole like vertical comic strip or whatever it was that had a crazy tall aspect ratio

This implementation might be worth taking inspiration from :)

PR #1709, opened 2025-02-21
Limit image height when it exceeds the device height

hjiangsu:
## Pull Request Description

This PR adds a constraint to image heights when they exceed the device height. If the image height is larger than the device height, then the image will be constrained to the device's height. This constraint is applied to both the feed page (when full height images are enabled) and the post page.

In these cases, the image's width might be stretched to compensate for the constrained height (which might make the media preview blurry as seen below)

## Issue Being Fixed

Issue Number: Somewhat related to #1417, but addresses this issue https://lemmy.world/post/25819194

## Screenshots / Recordings

https://github.com/user-attachments/assets/82a0d19c-122f-4a8f-b4ab-1926a38470a8

https://github.com/user-attachments/assets/643f3164-ab54-46bd-9eec-44117653968e

## Checklist

### DIFF
--- a/lib/shared/media_view.dart
+++ b/lib/shared/media_view.dart
@@ -181,6 +181,30 @@ class _MediaViewState extends State<MediaView> with TickerProviderStateMixin {
     );
   }
 
+  double getMinHeight() {
+    if (!widget.showFullHeightImages) return ViewMode.comfortable.height;
+
+    if (widget.postViewMedia.media.first.height != null) {
+      if (MediaQuery.of(context).size.height < widget.postViewMedia.media.first.height!) return MediaQuery.of(context).size.height;
+      return widget.postViewMedia.media.first.height!;
+    }
+
+    return ViewMode.comfortable.height;
+  }
+
+  double getMaxHeight() {
+    if (!widget.showFullHeightImages) return ViewMode.comfortable.height;
+
+    if (widget.postViewMedia.media.first.height != null) {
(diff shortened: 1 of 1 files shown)


micahmo (approved):
The code LGTM.

I'm not sure I understand why the preview is blurry in the demo video. When you open the image and zoom in to the same level (so that the width of the image matches the width of the device) it's not blurry. Maybe I don't understand what's happening.

hjiangsu:
> I'm not sure I understand why the preview is blurry in the demo video. When you open the image and zoom in to the same level (so that the width of the image matches the width of the device) it's not blurry. Maybe I don't understand what's happening.

Ahh - I just double checked this and it seems like this is an issue when viewing it on lemmy.ca. For other instances (e.g., lemmy.world or lemmy.ml), the preview looks correct. It could be due to how lemmy.ca generates the thumbnail images (as the preview uses the thumbnail image)?

It's not a Lemmy version issue as lemmy.ml and lemmy.ca are both running v0.19.9


| lemmy.ml | lemmy.world | lemmy.ca |
|-|-|-|
|simulator_screenshot_F127E5CF-0CA8-4D2F-BC3F-F9DD09AC228C| simulator_screenshot_665A41BE-6742-4C6E-84CE-E4FF45E5F228|simulator_screenshot_53EFA8E0-01BD-4F4E-B21A-9E2B1A0D10F7|

micahmo:
> the preview uses the thumbnail image

Ah ok, this explains it! You're probably right that different instances generate their thumbnails differently. Thanks!

PR #1715, opened 2025-02-27
Chore: Refactor image preview for MediaView

hjiangsu:
## Pull Request Description

This is a continuation of https://github.com/thunder-app/thunder/pull/1714, this time for the image preview within `MediaView`. I've extracted the `previewImage` function into its own widget and refactored to component to be a bit more generic.

Note: There is an existing `ImagePreview` widget in `/shared/image_preview.dart` which I would like to merge together at some point as well. This widget is used to display images within the markdown body (e.g., comments), link preview cards, and the advanced share sheet!

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1720, opened 2025-03-03
Chore: Refactor link previews and MediaView widget

hjiangsu:
## Pull Request Description

This PR is a continuation of the overall `MediaView` refactor - this time focusing on the `MediaView` widget and link previews. Here are some high level notes on the changes:
- Removed `LinkPreviewCard` and replaced its usages with `MediaView`. `LinkPreviewCard` shared a lot of related logic with `MediaView` and made the code harder to follow. This led to some visual inconsistencies across link previews vs. image/video previews.
- Started the deprecation of `scrapeMissingPreviews` option - I'm not sure how useful this option has been recently. In the past, there were issues with thumbnail generation which were mostly resolved in the backend. Having this option complicates the logic a bit, and has some potential performance issues as it has to fetch and parse through HTML content to find potential thumbnails (which should ideally be done at the backend regardless). That being said, do let me know if you feel strongly against this change!
- Combined shared logic in `MediaView` together (e.g., shared container and box constraints) - this removes duplicated logic across the image/video widgets. 
- Removed dependency on `PostViewMedia` within `MediaView`. `MediaView` now only depends on a `Media` rather than the full `PostViewMedia`. This allows the widget to be more flexible and be used in places that don't have access to `PostViewMedia` (e.g., creating a new post with images)

This should more or less conclude the major refactors for MediaView, although there may be some smaller additional PRs to possibly split up/simplify the logic even more (if needed). Again, I did try to do as much testing as possible but there may be some bugs that were not caught during my testing, so please mention any issues that pop up!

This PR should also fix the issues mentioned here: https://github.com/thunder-app/thunder/pull/1715#issuecomment-2692383906!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
> Started the deprecation of `scrapeMissingPreviews` option - I'm not sure how useful this option has been recently. In the past, there were issues with thumbnail generation which were mostly resolved in the backend. Having this option complicates the logic a bit, and has some potential performance issues as it has to fetch and parse through HTML content to find potential thumbnails (which should ideally be done at the backend regardless). That being said, do let me know if you feel strongly against this change!

I'm good with this! There are definitely still plenty of cases where Lemmy fails to find a thumbnail for the post, but these issues seem to be consistently reported to them and fixed by them. Plus users now have the option to add a custom thumbnail. So I don't see any reason for us to do additional work in this area.

micahmo (approved):
LGTM!

PR #1721, opened 2025-03-03
Remove scrape missing previews option

hjiangsu:
## Pull Request Description

This PR removes the remaining logic for scrape missing previews option mentioned in #1720. I've also added a migration step to remove existing instances of `setting_general_scrape_missing_previews` in `SharedPreferences`.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/enums/local_settings.dart
+++ b/lib/core/enums/local_settings.dart
@@ -153,8 +153,6 @@ enum LocalSettings {
 
   // General Settings
 
-  scrapeMissingPreviews(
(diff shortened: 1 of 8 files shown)


micahmo (approved):
LGTM!

Is there also a package we were using for this feature that could possibly be removed now?

hjiangsu:
> Is there also a package we were using for this feature that could possibly be removed now?

There's the `link_preview_generator` package which was originally used for scraping but was later replaced by our own `getLinkInfo` function. From what I can tell, `link_preview_generator` is still used in two places:
- The bottom sheet when long-pressing on a link
- The create post page when generating a suggested title for a URL

While this is out of scope for this PR, we could fully remove `link_preview_generator` by updating these use-cases to use `getLinkInfo` instead!

PR #1722, opened 2025-03-03
Fix tapping on media not marking post as read

hjiangsu:
## Pull Request Description

This PR fixes a regression with the recent MediaView refactor where tapping on the MediaView would not mark the post as read if the option was enabled.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
LGTM

PR #1713, opened 2025-02-26
Add back FAB actions for experimental post page

hjiangsu:
## Pull Request Description

This PR continues work on https://github.com/thunder-app/thunder/issues/1413 and resolves some of the core remaining issues (e.g., re-introducing FAB and comment search). More specifically, the following issues have been resolved:
- Bring back FAB (both with combine comment navigator and separate FAB)
- Bring back comment search and scrolling to search result
- When you create a comment, it does get displayed immediately, but it's not highlighted.

With this PR, I've also removed the legacy post page and any widgets/functions related to it. The new post page will now be displayed by default! I'll likely merge this in a bit sooner to be able to daily this change and find any additional bugs or issues with the new post page.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1413

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1728, opened 2025-03-05
Fix null check error when navigating to post from Report page

hjiangsu:
## Pull Request Description

This PR fixes an issue where attempting to navigate to a post from the Report page would cause a null check error because `PostViewMedia` was not provided. To fix this, the `navigateToPost` function has been modified to fetch the corresponding post if `postViewMedia` is null.

This is likely a regression from #1713 where the legacy post page was removed.

> Additionally, I've removed the conditionals mentioned in #1713 here too!

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

The issue:

[issue.webm](https://github.com/user-attachments/assets/e3a2fad8-595d-4793-9db3-a37ab5065fce)

Fixed:

[fixed.webm](https://github.com/user-attachments/assets/54dc3ad7-fd57-42b2-90e4-41633fb25e66)

## Checklist

### DIFF
--- a/lib/shared/comment_navigator_fab.dart
+++ b/lib/shared/comment_navigator_fab.dart
@@ -154,13 +154,6 @@ class _CommentNavigatorFabState extends State<CommentNavigatorFab> {
   }
 
   void navigateToParent() {
-    if (widget.comments == null) {
-      // This is a placeholder to allow the previous post page to function correctly.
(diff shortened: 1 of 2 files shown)


micahmo (approved):
LGTM!

PR #1729, opened 2025-03-05
Fix inconsistency with comment navigator

hjiangsu:
## Pull Request Description

This PR fixes a consistency issue with the comment navigator. More specifically, it fixes an issue where the comment navigator does not work on posts that are not scrollable.

The reason for this is likely due to the fact that the bottom spacer height is determined within the scroll controller's listener event, which only gets triggered when the user attempts to scroll the page. However, if the page is not scrollable (e.g., posts with very few comments), then that logic never gets triggered and the bottom spacer height is never calculated.

To resolve this, I've moved the logic for calculating the bottom spacer height into its own function, and added a `Timer` that gets triggered on build only if the bottom spacer height is not yet determined. In my testing, the bottom spacer height is determined within the first few build phases so this _should not_ have any major performance impacts!

> I've also added some cleanup with the `dispose` override for the controllers and timers.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

[issue.webm](https://github.com/user-attachments/assets/70607c0e-bd27-4f1b-b36c-0a268ea8bad6)

[fixed.webm](https://github.com/user-attachments/assets/934892a9-197d-499a-bec9-1b3b97b2baa6)

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -92,6 +92,9 @@ class _PostPageState extends State<PostPage> {
   /// The ID of the comment that should be highlighted
   int? highlightedCommentId;
 
+  /// The timer for calculating the bottom spacer height
+  Timer? _calculateBottomSpacerTimer;
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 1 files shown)


micahmo (approved):
LGTM! I think this used to work; I'm not sure why the post-frame callback doesn't work any more, but a Timer is just as good I think.

Does this also mean that when you open a post, you'll be able to manually scroll down such that the last comment can always go up to the very top of the screen?

hjiangsu:
> Does this also mean that when you open a post, you'll be able to manually scroll down such that the last comment can always go up to the very top of the screen?

Yup! Since the bottom spacer height gets calculated and updated, it causes the post page to be scrollable

Issue #1737, opened 2025-03-10
Reduce slow frames and jank


hjiangsu:
### Is Your Feature Request Related to a Problem?

This is an umbrella issue to track PRs related to improving overall performance of Thunder, which includes reducing or eliminating sources of slow frames or jank that occurs while using the app.

### Feature Description

The goal here is to find sources of jank, and optimize the slow frames to make the overall experience smoother. To scope it down, focus will be on the feed pages and related widgets to start. After that, we can put focus on the post page.

There are a few things we can do:
1. Perform performance profiling to find cases of slow frames and jank
2. Perform CPU profiling to find long running and expensive operations
3. Find sources where we're triggering widget builds unnecessarily

_This will take some trial and error as I'm not too familiar with doing performance profiling but hopefully others can chime in on how to better optimize app performance!_ 

Some resources:
- https://docs.flutter.dev/perf

### Additional Context

The experience in Thunder seems to vary depending on the phone and operating system. For iOS, the experience is generally good with little or no stuttering or jank. However, the experience on Android differs depending on the specific device (possibly CPU).

I recently got access to a physical Android device (Pixel 9 Pro), so I can finally do some tests on a physical device.

hjiangsu:
I'll place some observations and ideas here in case anyone would like to work on them! I'll also be updating this comment when I have more ideas.

- [x] Remove `VisibilityDetector` from `FeedPostCardList` when 'Mark Read on Scroll' is not enabled. I believe `VisibilityDetector` performs some expensive calculations when trying to determine if a widget is within view.
- [x] Use `SliverList` instead of `SliverMasonryGrid` when tablet mode is not enabled as it might be more performant.
- [ ] Investigate the use of isolates for more expensive functions to move them off the UI thread.
- [ ] Improve parsing of MediaView for text posts
- [ ] Flatten widget layout tree for post metadata and author widgets

PR #1743, opened 2025-03-13
Use SliverList when tablet mode is disabled

hjiangsu:
## Pull Request Description

This PR applies a small change where if tablet mode is disabled, the feed post list will fallback to use `SliverList` rather than `SliverMasonryGrid`. 

**Context**
After doing some CPU profiling, I observed that using `SliverMasonryGrid` took up a lot more CPU time as compared to `SliverList`. I'm not entirely sure why this is the case, but my assumption is that `SliverMasonryGrid` has to do more calculations to determine how to position the corresponding elements.

Since non-tablet mode always displays the elements in a list rather than a grid, switching from `SliverMasonryGrid` to `SliverList` shouldn't have any perceived differences.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1737

## Checklist


micahmo (approved):
Nice improvement!

PR #1744, opened 2025-03-13
Fix issue with media not being marked as read for compact posts

hjiangsu:
## Pull Request Description

This PR fixes an issue where tapping on the media thumbnail for compact posts would not mark the post as read. It turns out that `postId` was not properly passed into MediaView, which is needed in order to determine which post to mark as read!

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Thank you!!

micahmo:
Just FYI, not sure if it's related to this PR, but it looks like in compact mode the image thumbnail (for an image or a link) is not getting darkened when the post is marked as read.

**EDIT**: Fixed in https://github.com/thunder-app/thunder/pull/1768

Issue #1718, opened 2025-03-02
Community icons for compact view posts list


Matth7878:
### Is Your Feature Request Related to a Problem?

When looking at post lists you can see for meta data on which community an item has been posted. But you can't choose to display it's icon.
Option exists but only for card view.

### Feature Description

Allow toggle to display community icons for both compact and card view.
Same as for choosing to show post author for instance.

Matth7878:
Additional note. When showing community icon in compact view for posts list then list may be a little confusing as icons won't be aligned since metadata is "user to community".

To fix it and have a nice rendering, with community icons it should be changed to : "community from user"
This way all community icons would be nicely aligned on left just below post title.
Having those icons when reading title is really useful because sometimes you would read a post you won't have only based on post title. Having community icon just below allow you to recognize instantaneously community, it allow also to more quickly scroll list and spot post from a community you are especially looking for.

PR #1746, opened 2025-03-15
General post card improvements, and ability to toggle community icons in compact mode

hjiangsu:
## Pull Request Description

This PR applies various improvements to the post card and related widgets.

**Functional Updates**
There's been some slight functional changes alongside some general refactoring (see below), including:
- The ability to show community icons in both compact/card views. With this change, the username/community information will now be split into two rows rather than one row. Instead of having `user to community`, the community will show up first with the user below it.
  - This change helps avoid the situation where the community icon is displayed beside the username, which might be confusing. Additionally, this change helps remove the `to` string that would otherwise be difficult to localize to other languages (especially those that don't have an equivalent string)
  - The community icons setting has been moved to the shared settings area under post appearance (was previously under the card settings) to reflect this change.
- The post title's default text size has increased by 0.5. This is to accommodate for the above change to better help distinguish the title from the username/community information.

**Refactoring**
There has been quite a bit of refactoring to the post card, author/community widget, and feed card divider to improve overall readability and hopefully performance. Specifically, these are the changes made:
- Refactored the logic in `PostCard` to only build the swipe-related widgets when swipe gestures are enabled. I also attempted to optimize and organize the logic a bit more (e.g., debouncing the `onUpdate` callback for the Dismissible widget and `_onPointerMove`). 
- Created a new widget `PostCardActionBackground` that contains the logic for handling the swipe action from the post card's Dismissible.
- Refactored `PostCommunityAndAuthor`, splitting up the username/community widgets into their own widgets.
- Refactored `FeedCardDivider` to only listen to required changes from the ThunderBloc.

Like all the other refactoring efforts, I've only been able to test this briefly so there may be slight issues that pop up due to the refactoring. I hope to catch any issues after I daily the changes for a while (or when the next nightly is released, which I hope to do soon)

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1718, https://github.com/thunder-app/thunder/issues/1737

## Screenshots / Recordings

Here are some screenshots on how this new look looks like on compact mode with different settings enabled:

| | |
|-|-|
|Screenshot_1742011960| Screenshot_1742011987 |
|Screenshot_1742012268 |Screenshot_1742012043 |
|Screenshot_1742012065|Screenshot_1742012083| |

## Checklist


Matth7878:
Thanks a lot for working on it.

I noticed you splitted community and user on two lines and use a big icon for community taking both lines.
What would happen if you only show community and not user? Would community icon be shrinked to take only one line? Or compact mode will become less compact? 🫤

Is it possible when not showing post author to have community icon be the same line as text and make it take only a slightly bigger vertical space. This way showing icon wouldn't make you lose a lot of vertical space.
I admit as the one having requested community icon in compact posts list that I would choose not to use it if it's like in the screenshot with an empty line for author. (Or if icon and community name are vertically centered but take 2 lines)

**Edit : To sum it up it would be nice if community icon especially when not showing author would be displayed the same way users icon is displayed for comments. Or for post in post detail screen where there already was author and community icons on one line. (The screen with comments)**

Should I create a separate feature request ?

Anyway It's already nice when showing authors so thanks again. For those showing author it will looks great! 👍

hjiangsu:
Thanks for providing some feedback on this! I think it does make sense to make the icons smaller when post author is not shown. I've adjusted the logic slightly to accommodate for this!

If community icons are enabled, and post authors are not enabled, it'll show up like this:
| | |
|-|-|
|Screenshot_1742147308|Screenshot_1742147354|

If community icons are enabled, and post authors are also enabled, it'll show up like before:
| | |
|-|-|
|Screenshot_1742147485|Screenshot_1742147461|

Thoughts on this change?

Matth7878:
👍 looks nice. Thanks so much for having addressed it.

micahmo (approved):
LGTM, and I like the second change based on feedback!

Maybe we could also add an option to show user icons too? Can be separate.

hjiangsu:
> Maybe we could also add an option to show user icons too? Can be separate.

I had similar thoughts here! We can replace the community icons with user icons when viewing a community feed (where it wouldn't make sense to show the community icon).

micahmo:
Just FYI, it looks like a long community name can bump up against the right side (not sure if this is a new problem) and we lose the "subscribed" indicator.

| image |
| - |

Maybe we need to ellipsize it?

PR #1750, opened 2025-03-17
Refactor community avatar

hjiangsu:
## Pull Request Description

This PR refactors the community avatar by removing its dependence on `Community` and replacing it with primitive types (e.g., String, bool, etc.). This is part of a broader effort to migrate Thunder’s logic away from the Lemmy API models.

**Context**
The Lemmy API models have changed quite a bit from v0.19.x to the upcoming v1.0.0, which is a challenge for us since a lot of Thunder's logic is closely tied to the 0.19.x models. Adapting our existing models to support multiple API versions—0.19.x, 1.0.x, and beyond—would add unnecessary complexity, making it harder to keep track of which fields belong to which version.

To make Thunder more resilient to API changes, I want to decouple it from the Lemmy APIs and introduce our own internal models throughout the app. This will not only shield Thunder from larger API changes but also make it easier to extend support to other platforms, like PieFed.

To start this process, I plan to first replace our reliance on Lemmy API models with primitive types (this PR for example). Once that's done, we can begin building our internal models based on Thunder’s current needs.

@micahmo If you have any feedback or are interested in helping with this process, feel free to let me know!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -20,6 +20,7 @@ import 'package:thunder/account/models/draft.dart';
 import 'package:thunder/community/bloc/image_bloc.dart';
 import 'package:thunder/core/enums/media_type.dart';
 import 'package:thunder/core/models/media.dart';
+import 'package:thunder/core/models/models.dart';
 import 'package:thunder/post/widgets/post_action_bottom_sheet.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
 import 'package:thunder/core/auth/helpers/fetch_account.dart';
@@ -847,7 +848,10 @@ class _CommunitySelectorState extends State<CommunitySelector> {
             children: [
               Row(
                 children: [
-                  CommunityAvatar(community: widget.communityView?.community, radius: 16),
+                  CommunityAvatar(
+                    community: ThunderCommunity(widget.communityView?.community),
+                    radius: 16,
+                  ),
                   const SizedBox(width: 12),
                   widget.communityId != null
                       ? Column(
--- a/lib/community/widgets/community_drawer.dart
+++ b/lib/community/widgets/community_drawer.dart
@@ -12,7 +12,7 @@ import 'package:thunder/account/bloc/account_bloc.dart';
 import 'package:thunder/account/utils/profiles.dart';
 import 'package:thunder/community/bloc/anonymous_subscriptions_bloc.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
-import 'package:thunder/core/models/thunder_community.dart';
+import 'package:thunder/core/models/models.dart';
 import 'package:thunder/feed/feed.dart';
 import 'package:thunder/shared/avatars/community_avatar.dart';
 import 'package:thunder/shared/avatars/user_avatar.dart';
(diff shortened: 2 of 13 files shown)


micahmo (approved):
This LGTM and I like the general idea of decoupling from the Lemmy models.

I do see a lot of duplicated code that might be easy to forget (`postView.community.title.isNotEmpty ? postView.community.title : postView.community.name`). It makes me wonder if we should our Thunder model first, and give it some calculated fields based on the underlying Lemmy model. Then when the Lemmy model changes, we only have to change one place. Kinda like the MVVM pattern.

hjiangsu:
> It makes me wonder if we should our Thunder model first, and give it some calculated fields based on the underlying Lemmy model. Then when the Lemmy model changes, we only have to change one place. Kinda like the MVVM pattern.

I think that's also an approach we can take, although I'm a bit fuzzy on the details! Do you have an example to help visualize how we would approach it? (I'll be honest, my experience with MVVM pattern is very limited 😅)

micahmo:
It wouldn't be exactly MVVM, that's just how I'm approaching the thought process because it's a pattern I've used before. But something like this (in pseudocode).

The **M**odel is the Lemmy model from the API. The **V**iew is a Thunder widget. The **V**iew**M**odel is a wrapper class.

```
class ThunderCommunity {
  constructor ThunderCommunity(LemmyCommunity lemmyCommunity)

  string name => lemmyCommunity.title.isNotEmpty ? lemmyCommunity.title : lemmyCommunity.name
}
```

Then whenever we construct a view (widget) that depends on a community, we can pass a wrapper

```
CommunityAvatar(community: ThunderCommunity(lemmyCommunity))
...
text: thunderCommunity.name
```

Then if the Lemmy community model ever happens to change, we only have to change it one place (the wrapper), and all the views/widgets which depend on the name will continue to access the calculated property.

IMO, this balances the two concerns which are: (a) minimizing changes when the Lemmy API makes a breaking change and (b) reducing duplicate code and minimizing bugs when we forget (in this example) that name is always calculated based on title, then name.

hjiangsu:
Thanks for the explanation, I see where you're coming from! I can try to adapt the current PR to use this type of pattern to see how it works in practice.

micahmo:
It's totally up to you! I can just imagine that this new approach is actually harder to maintain, because (unlike breaking API changes, which we can detect with the compiler), it will be harder to remember various patterns like `lemmyCommunity.title.isNotEmpty ? lemmyCommunity.title : lemmyCommunity.name` going forward. The convenience of tying the Lemmy model to the widget was that the calculation only had to be done in one place, so I'd love to keep that benefit!

hjiangsu:
I think your suggestion works, so I'm open to trying it! A benefit to this is that we also build up the internal Thunder models while decoupling Thunder from the Lemmy models. At the end of the day, my view is to hopefully do something like this:

APIs (e.g., Lemmy, PieFed, etc.) -> Repository (e.g., CommunityRepository) -> Model (e.g., ThunderCommunity, which is consumed by widgets and other internal logic)

The repository would ultimately be in charge of converting the API models into our internal models to be consumed by blocs/widgets.

micahmo:
Sounds good! And yes, I see your point. In the first phase, the Thunder model is responsibly for doing the conversion, but that could be done via an adapter/repository type of pattern.

hjiangsu:
> In the first phase, the Thunder model is responsibly for doing the conversion, but that could be done via an adapter/repository type of pattern.

Exactly! Another external reason for this de-coupling of logic (aside from the upcoming Lemmy API changes) is to make it easier for other APIs to adapt to Thunder. For example, PieFed has recently started efforts to make an API that is compatible with Thunder - see https://codeberg.org/rimu/pyfedi/issues/13.

PS - the PR is ready for review again if you'd like to take a second look

micahmo (approved):

PR #1753, opened 2025-03-17
Refactor user header

hjiangsu:
## Pull Request Description

This PR refactors `UserHeader` to use the `ThunderUser` model. A few fields were added into `ThunderUser` to accommodate for this change, including:
- `banner` - the user's banner picture URL
- `totalPosts` - the number of total posts created by the user
- `totalComments` - the number of total comments created by the user

Since the 0.19.x Lemmy API splits up a user's counts (e.g., total posts/comments), I've added some additional fields to the constructor of `ThunderUser` to accept these values rather than creating a new model to hold these counts.

```dart
ThunderUser(this.user, {this.totalPosts, this.totalComments});
```

This change also somewhat matches up with the recent changes to the Lemmy API where aggregate values are combined into their respective parent model. See https://github.com/LemmyNet/lemmy-js-client/pull/492

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/feed/view/feed_page.dart
+++ b/lib/feed/view/feed_page.dart
@@ -16,6 +16,7 @@ import 'package:thunder/community/widgets/community_header.dart';
 import 'package:thunder/community/widgets/community_sidebar.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
 import 'package:thunder/core/enums/local_settings.dart';
+import 'package:thunder/core/models/models.dart';
 import 'package:thunder/core/models/post_view_media.dart';
 import 'package:thunder/core/singletons/lemmy_client.dart';
 import 'package:thunder/feed/bloc/feed_bloc.dart';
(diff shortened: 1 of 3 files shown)


lib/feed/view/feed_page.dart
```diff
@@ -447,7 +448,11 @@ class _FeedViewState extends State<FeedView> {
                                 child: Column(
                                   children: [
                                     UserHeader(
-                                      getPersonDetailsResponse: state.fullPersonView!,
+                                      user: ThunderUser(
+                                        state.fullPersonView!.personView.person,
+                                        totalPosts: state.fullPersonView!.personView.counts.postCount,
```
micahmo, line 453: You could also have the `ThunderUser` take the whole `personView`.

micahmo (approved):
LGTM! I'm liking this approach.

lib/feed/view/feed_page.dart
hjiangsu: This might be a bit more difficult since we don't always get back the whole `postView` in some cases

micahmo: Sure, it could be optional! `totalPosts` is nullable anyway, so it would be null if we didn't have the `fullPersonView`.

hjiangsu: Ahh sorry, I misunderstood! I thought you meant replace the current `person` with `personView` 😅.

We can definitely pass in the `personView` as an optional parameter in this case. I'll likely do this if I encounter another field that comes from `personView`.

PR #1754, opened 2025-03-17
Refactor user sidebar to use internal models and add localizations

hjiangsu:
## Pull Request Description

This PR refactors the user sidebar to use `ThunderUser` and `ThunderCommunity`. Additionally, I've:
- Added localizations to some of the existing strings (e.g., posts, contributions, etc.)
- Applied the feedback from https://github.com/thunder-app/thunder/pull/1753#discussion_r1999327406 and added a parameter `userView` (to match up with the user naming convention) to `ThunderUser`.
- Added missing comments to the widget parameters
- Adjusted the logic so that an instance admin cannot be blocked

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/core/models/thunder_community.dart
+++ b/lib/core/models/thunder_community.dart
@@ -1,13 +1,29 @@
 import 'package:lemmy_api_client/v3.dart';
 
 class ThunderCommunity {
-  Community? community;
+  /// The Lemmy API model for the community.
+  Community community;
 
   ThunderCommunity(this.community);
 
-  String get name => (community?.title.isNotEmpty == true ? community?.title : community?.name) ?? '';
+  /// The ID of the community.
+  int get id => community.id;
 
-  bool get locked => community?.postingRestrictedToMods ?? false;
+  /// The name of the community. If the community has a title, it is used. Otherwise, the name is used.
+  String get name => community.title.isNotEmpty == true ? community.title : community.name;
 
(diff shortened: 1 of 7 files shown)


lib/core/models/thunder_community.dart
```diff
@@ -1,13 +1,29 @@
...
+  /// The ID of the community.
+  int get id => community.id;
 
-  bool get locked => community?.postingRestrictedToMods ?? false;
+  /// The name of the community. If the community has a title, it is used. Otherwise, the name is used.
+  String get name => community.title.isNotEmpty == true ? community.title : community.name;
 
-  String? get icon => community?.icon;
+  /// The name of the community.
+  String get communityName => community.name;
```
hjiangsu, line 16: I'm not sure if there's a better name for this field (as `name` was already taken). For users, it's username, but there's no equivalent for communities.

micahmo: Hmm, maybe the other one should be called `titleOrName` to make it clear how the logic works? Then this one can just be `name`? Not sure.

micahmo (approved):
LGTM!

lib/core/models/thunder_community.dart
hjiangsu: Hmm, maybe I'll keep it as is for now and change it in the future if this becomes an issue!

PR #1756, opened 2025-03-17
Enable full-height images by default

hjiangsu:
## Pull Request Description

This PR changes the default setting for "Show Full Height Images" from `false` to `true`. Following the release of v0.19.9 on `lemmy.world`, the majority of users are now at a version that supports image dimensions from the API. This means that our (semi-expensive) workaround for fetching image dimensions no longer applies to a majority of users!

From a UX perspective, making images full-height by default in card view makes more sense than constraining their height. However, users will still have the option to limit the height if they prefer. Additionally, I _believe_ that Thunder is the only app that sets a standard height for images in card view by default.

Note: this change will only affect users who have not previously changed the setting (e.g., no value set in the user's shared preferences).

> Looking at https://fedidb.org/software/lemmy, there are still a few larger instances that are running v0.19.5 and below (image dimensions was introduced in 0.19.6). I don't expect Thunder to have a general release any time soon, so this does give additional buffer time for the remaining instances to update their versions!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/settings/pages/post_appearance_settings_page.dart
+++ b/lib/settings/pages/post_appearance_settings_page.dart
@@ -70,7 +70,7 @@ class _PostAppearanceSettingsPageState extends State<PostAppearanceSettingsPage>
   bool showCommunityIcons = false;
 
   /// When enabled, images and media will display full height. By default, they are displayed with a max height
-  bool showFullHeightImages = false;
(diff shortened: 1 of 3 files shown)


micahmo (approved):
Makes sense!

Issue #1725, opened 2025-03-05
Crosspost should copy body text and link to the original post for all post types, not just text-only


DavidHiraki:
### Is Your Feature Request Related to a Problem?

When crossposting a text-only post, Thunder automatically populates the body field with a link to the original post and quotes any existing body text. This is good.

When crossposting a URL or image post, the body field is empty. No link to the original post is added to the body field, nor is any body text quoted. This is not good.

Issue raised here: https://sh.itjust.works/post/24540591

### Feature Description

When crossposting a URL or image post, Thunder should automatically populate the body field with a link to the original post and quote any existing body text.

### Additional Context

Discussed here: https://lemmy.world/comment/12719015

There was at least one argument against this feature (wanting to change the contents of the cross-post to be relevant to the community to cross-post to), so it might be worth making this a user-configurable setting, though I think it should be enabled by default.

PR #1759, opened 2025-03-17
Add prompt to inject original post body for non-text cross posts

hjiangsu:
## Pull Request Description

This PR adds the ability to inject the original post body for non-text cross posts, as described in https://github.com/thunder-app/thunder/issues/1725 and the relevant discussions. This works by displaying a snackbar whenever the create post page detects a cross post. I'm not sure if this is the best way to do it, so some feedback on this would be appreciated @DavidHiraki!

> The snackbar will only show up if there's an existing link (e.g., link, image, etc.), and if the original post contained text. For text-only crossposts, the post body will automatically be added.

See the video below for a quick demo.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1725

## Screenshots / Recordings

https://github.com/user-attachments/assets/0b28f677-8e30-44f1-b436-6ef7fba6e406

## Checklist

### DIFF
--- a/lib/community/pages/create_post_page.dart
+++ b/lib/community/pages/create_post_page.dart
@@ -72,6 +72,9 @@ class CreatePostPage extends StatefulWidget {
   /// [postView] is passed in when editing an existing post
   final PostView? postView;
 
+  /// Whether or not this post is a cross post
+  final bool isCrossPost;
+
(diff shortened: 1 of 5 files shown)


micahmo (approved):
I think this is a very reasonable solution!

PR #1760, opened 2025-03-17
Fix visual glitch navigating to posts via link

micahmo:
## Pull Request Description

This PR fixes a visual glitch that appears when navigating to a post via a link.

## Issue Being Fixed

Issue Number: N/A

### Before

(It happens every time, but the screen recording only catches it sometimes.)

https://github.com/user-attachments/assets/f0bc00d0-3cae-4b7e-8437-cce139d9b7b8

### After

https://github.com/user-attachments/assets/6090d75e-08bc-42d7-93b5-8196d740c0a4

## Checklist


hjiangsu (approved):
LGTM!

PR #1768, opened 2025-03-20
Fix an issue where image previews would not be dimmed when read

micahmo:
## Pull Request Description

This PR fixes an issue where the image preview would not get dimmed when the post is marked as read. This seems to have been lost during a refactor, and I copied the logic from a similarly named file, `lib\shared\image_preview.dart`, which makes me concerned that some other logic might have been lost as well.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/shared/image/image_preview.dart
+++ b/lib/shared/image/image_preview.dart
@@ -106,6 +106,8 @@ class _ImagePreviewState extends State<ImagePreview> with SingleTickerProviderSt
       height: widget.height,
       width: widget.width,
       fit: widget.fit,
+      color: widget.viewed == true ? const Color.fromRGBO(255, 255, 255, 0.55) : null,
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM!

hjiangsu:
> which makes me concerned that some other logic might have been lost as well.

I believe it should be good if we set the changes here, as this widget is responsible for showing all images related to `MediaView` (e.g, images, links, video thumbnails)!

The other file `lib\shared\image_preview.dart` is used on some other parts of the app like advanced share, and markdown images in comments (although, I would like to merge them in the future to reduce code duplication)

Edit: Just checked the refactor PR (https://github.com/thunder-app/thunder/pull/1715/files) and it seems like those were the only two missing parameters!

micahmo:
> it seems like those were the only two missing parameters!

Perfect!

Issue #1770, opened 2025-03-20
Selected comments sort option in "Sort Options" menu and topbar does not match


nimxaa:
### Bug Description

The selected comments sort option in "Sort Options" menu and topbar does not match.

<img src="https://github.com/user-attachments/assets/f65853b9-da6d-4963-8988-f7250df654bc" width="40%">

<img src="https://github.com/user-attachments/assets/5cd376b9-46c3-4276-8fce-cfb96767cc4b" width="40%">

### Expected Behaviour

It should show same in both.

### Steps to Reproduce

1. Open post
2. Try changing comment sorting
3. See the issue

### App Version

0.6.1

### Device

OnePlus 7 Pro

### OS

Android 11

micahmo:
May I ask what Lemmy instance you are using?

nimxaa:
lemm.ee

PR #1771, opened 2025-03-21
Fix displayed comment sort mode

micahmo:
## Pull Request Description

This PR fixes an issue where the displayed comment sort type in the post page app bar was incorrect.

There was some weird mapping between `CommentSortPicker.getCommentSortTypeItems` and `allSortTypeItems`, and to be honest I don't know why that was needed since `getCommentSortTypeItems` returns a `ListPickerItem<CommentSortType>` which is exactly what we need here. If you can think of any potential problems here, please let me know!

## Issue Being Fixed

Issue Number: #1770

## Checklist


hjiangsu (approved):
LGTM!

Issue #1786, opened 2025-03-27
Please restore multi-line (pre 0.7.0-1) comment depth indicators


everdred:
### Is Your Feature Request Related to a Problem?

I was disappointed to see the visual style of Thunder's comment depth indicators change with the latest prerelease, as this was one of only a couple of Android Lemmy clients that have what I consider to be usable indicators. (Eternity being the other.)

The new-style depth indicators require the user to remember which color a comment at the parent level was and maintain an extraordinary amount of focus while scrolling, or memorize the colors of the rainbow in reverse. ("This was a green comment, I need to scroll upwards to find the next, uh... red, orange, yellow, green... yellow comment.")

![Image](https://github.com/user-attachments/assets/68ca6d05-b240-4683-a7b2-a37488a4b276)

(It also takes an extraordinary attention to detail to notice that user "Graph..." is two, not one, levels deeper than "Cowbee..." and colors alone don't help nearly as much as colors+lines would.)

### Feature Description

The old style (which was easiest to follow, IMO, in rainbow mode vs. monochrome) allows the user to simply visually follow a line, while scrolling, to locate the comment that is the parent or a sibling of the comment they just read, in order to place it in context. In a complex discussion this may be outside of the viewport, and may require quite a bit of scrolling to locate. 

### Additional Context

I'm not sure what the advantage is of the more popular style of depth indicators in modern software ("minimalism" I suspect), but it would be appreciated if the old style could at least remain an option in Thunder.

everdred:
What's more, I just discovered that the rainbow-themed colors I had been referring to are specific to the "accent color" theme I have selected ("Mango," btw), and aren't universal Thunder colors. In other color themes the depth colors don't seem to be as predictable.

hjiangsu:
> I was disappointed to see the visual style of Thunder's comment depth indicators change with the latest prerelease

Hey! Thanks for the report - I believe this is actually a bug that was introduced in the nightly build, as there were no intended changes to the comment indicators.

This change was likely due to our transition from the previous post page (pre-0.7.0) to the current post page.

> What's more, I just discovered that the rainbow-themed colors I had been referring to are specific to the "accent color" theme I have selected ("Mango," btw), and aren't universal Thunder colors. In other color themes the depth colors don't seem to be as predictable.

[...]

PR #1788, opened 2025-03-28
Improve comment depth indicators

hjiangsu:
## Pull Request Description

> Review with whitespace off

This PR improves on our current comment depth indicators and fixes an issue with the thin-style comment indicator mentioned in https://github.com/thunder-app/thunder/issues/1786.

The following changes were made:
- Implemented a custom decoration widget (`CommentDepthIndicatorDecoration`) which renders the proper comment indicator based on the indicator style and colours.
- Simplified logic within the comment-related widgets to use the new comment indicator decoration widget.

> Note: This is the first time I've implemented a custom Decoration class, so there may be bugs associated with this. Let me know if you encounter any issues!

**Additional Context**
The reason why we lost the previous thin-styled comment depth indicators was due to changes in the way we handle the rendering of comments in the new post page. Previously, we were rendering comments in a somewhat recursive manner (parent comment renders the child comments, and the child comments render their children, and so on).

When we rendered comments in that recursive fashion, we were able to apply a border to each of the comment's `Container`, and likewise, render the depth indicators. However, we are now rendering each comment one at a time rather than recursively. This new method provides some benefits:
- Improved performance for large comment chains
  - Imagine a comment chain consisting of hundreds of comments - in the previous method, if the parent comment was in the viewport, then all it's child comments would likely need to be rendered even if they were not in the viewport simply because the parent was in view
- Ability to navigate between child comments - since we no longer render comments recursively, it made it much easier to implement next/previous comment navigation
- Simplification of the comment rendering logic - removing the recursive nature of the comment rendering makes it much easier to understand and potentially optimize

However, due to this change, I wasn't able to apply our existing logic to handle the depth indicators. I did attempt some alternative methods of rendering the depth indicators but none of them accomplished what I needed (e.g., some implementations didn't fully fill the height of the comment, other implementations weren't performant because it required additional layout calculations to determine the height, etc.)

To solve this, I created a new `Decoration` which renders the comment indicators. This implementation seems to work fairly well (from my testing so far), and allows us to add more customization in the future! For example, we could have thick style indicators that show all of the comment levels (e.g., how the thin style is implemented) and vice versa

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1786

## Screenshots / Recordings

@everdred I've attached some screenshots of the implementation - let me know your thoughts on this as it might not exactly match up with the pre 0.7.0-1 versions.

The video was too large for GitHub, so here's a link to it: https://files.catbox.moe/uh2nij.mp4

| Thin | Thick |
|-|-|
|Screenshot_1743174277| Screenshot_1743174256|
|Screenshot_1743174323|Screenshot_1743174351|

## Checklist

### DIFF
--- a/lib/comment/widgets/comment_card.dart
+++ b/lib/comment/widgets/comment_card.dart
@@ -6,17 +6,16 @@ import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:lemmy_api_client/v3.dart';
 
 import 'package:thunder/comment/enums/comment_action.dart';
+import 'package:thunder/comment/widgets/comment_depth_indicator.dart';
 import 'package:thunder/utils/navigation.dart';
 import 'package:thunder/comment/widgets/comment_action_bottom_sheet.dart';
 import 'package:thunder/core/auth/bloc/auth_bloc.dart';
-import 'package:thunder/core/enums/nested_comment_indicator.dart';
 import 'package:thunder/core/enums/swipe_action.dart';
 import 'package:thunder/post/bloc/post_bloc.dart';
 import 'package:thunder/post/utils/comment_actions.dart';
 import 'package:thunder/shared/comment_content.dart';
 import 'package:thunder/shared/text/scalable_text.dart';
 import 'package:thunder/thunder/bloc/thunder_bloc.dart';
-import 'package:thunder/utils/colors.dart';
 
 class CommentCard extends StatefulWidget {
   /// The [CommentView] containing the comment information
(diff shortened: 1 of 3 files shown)


hjiangsu:
@micahmo This should be ready for review if you'd like to take a look at it

micahmo (approved):
Looks great! Nice implementation with decorations!

The only question I had (didn't see demonstrated) was how the decorations react to the post being swiped one way or the other? I don't think it matters too much, I was just curious.

hjiangsu:
> The only question I had (didn't see demonstrated) was how the decorations react to the post being swiped one way or the other?

Good question! Since the decoration is now part of the comment itself, swiping actions will move the comment depth indicators along with the comment. To resolve this, I just hide the comment indicators when swiping on the comment.

https://github.com/user-attachments/assets/34214d30-b3ac-426d-9d27-a751d52e3386

micahmo:
That looks fine to me! Although I also don't think it would be the end of the world if the indicators moved with the comment. That's how Relay for Reddit and Sync for Lemmy both work (just checked). Up to you.

hjiangsu:
Ahh I didn't know that! I do think removing the indicators feels a bit cleaner (and also helps focus on the content of the comment) so maybe we can roll with that for now

hjiangsu:
@everdred I'm going to merge this in first, but do let me know if you have any feedback on the UI/UX

everdred:
@hjiangsu Thanks so much for fixing this! I'm having trouble finding visual differences in your screenshots vs. some old ones I have, so I think it's good.

micahmo:
Sorry to jump back in here, but I did notice one visual difference compared to before. In thick mode, the first child comment is now completely aligned to the left of the screen. Previously there was one line-thickness worth of space indenting the first child (and thus all the rest of the children). It's not a big deal either way, but I think indenting each level by one would restore the previous appearance.

PR #1789, opened 2025-03-28
Fix a problem with editing not updating the post content, and related bugs

micahmo:
## Pull Request Description

This PR primarily fixes an issue where, after editing a post, the post page would show the original, un-edited post.

This was caused from the implementation of `didUpdateWidget` that would update the local copy of the `postViewMedia` from the original version of the `widget.postViewMedia` very frequently.

I also discovered that `widget.postViewMedia` was used in other cases, like cross-posting, causing the unedited version of the post to be used in those cases as well. By switching everything to `postViewMedia`, the other issues are fixed as well.

One more case was related to replying. This was using the `widget.initialPostViewMedia`. I found some cases where we checked the state first, like `state.postView ?? widget.initialPostViewMedia`, and I changed the rest of the cases to be like that, but the state wasn't getting the update from the edit. So I added a new event to the `PostBloc`, like we already have in the `FeedBloc` to update the post.

Finally, per @hjiangsu's suggestion, I also retested a use case of modifying the post and ensuring that the update is reflected in the feed. Sure enough, it did not work, but it's unclear if removing `didUpdateWidget` was the cause. Regardless, I fixed it my ensuring that the callback, `onPostUpdated`, happened more often for any change.

Note: There is still one use case that's broken: opening a post, then navigating back to the feed, then updating the post in the feed (e.g., upvote), then navigating back into the post. However, I'm almost positive this was broken by #1520 because that (intentionally) pulls in a "stale" post to preserve the state. We will have to think about how to fix that use case for modifications like upvoting. That can be a separate fix.

## Issue Being Fixed

Issue Number: https://lemmy.world/post/25585479

## Screenshots / Recordings

https://github.com/user-attachments/assets/fe9570bc-27b6-4ede-baee-84ec977518f1

## Checklist

### DIFF
--- a/lib/post/bloc/post_bloc.dart
+++ b/lib/post/bloc/post_bloc.dart
@@ -85,6 +85,10 @@ class PostBloc extends Bloc<PostEvent, PostState> {
     on<UpdateCollapsedComment>(
       _onUpdateCollapsedComment,
     );
+    on<PostUpdatedEvent>(
+      _onPostUpdated,
+      transformer: throttleDroppable(Duration.zero),
+    );
   }
 
   /// Fetches the post, along with the initial set of comments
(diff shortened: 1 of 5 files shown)


hjiangsu (approved):
LGTM! Thanks for tackling on this issue and for the additional context into this 😊

> Note: There is still one use case that's broken: opening a post, then navigating back to the feed, then updating the post in the feed (e.g., upvote), then navigating back into the post. 

Agreed, I think that was likely introduced in https://github.com/thunder-app/thunder/pull/1520 due to the stale state

PR #1795, opened 2025-03-28
Fix the lack of a spinner on startup when slow

micahmo:
## Pull Request Description

This PR fixes an issue where there is no progress indicator on the initial feed page when the user first starts the app if the user is on a slow connection. I feel like we've done something like this before, so if you can think of any reason why this is a bad idea, please let me know!

Also I tweaked the height of the appbar in this initial state so that the initial progress indicator I added here and the progress indicator shown subsequently on the feed page line up.

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/user-attachments/assets/921a5033-dfb7-4c0c-bb50-b09d3ae75ccc

### After

https://github.com/user-attachments/assets/9dbf7e3c-258e-4c9c-aed2-e870d49bb39e

## Checklist

### DIFF
--- a/lib/thunder/pages/thunder_page.dart
+++ b/lib/thunder/pages/thunder_page.dart
@@ -518,9 +518,9 @@ class _ThunderState extends State<Thunder> {
                         case AuthStatus.initial:
                           context.read<AuthBloc>().add(CheckAuth());
                           return Scaffold(
-                            appBar: AppBar(),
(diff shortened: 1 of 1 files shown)


hjiangsu (approved):
LGTM - I don't see any issues with adding this in, since it gives the user more information about the status of the app!

PR #1796, opened 2025-03-28
Improve post loading

micahmo:
## Pull Request Description

Following onto #1795, I was doing some network latency testing, and I noticed when I was on a slow connection, I would navigate to a post, it would spin for a while, and then I would get just the post with no comments. I thought perhaps the comments were still loading in the background without a progress indicator, so my original plan was just to tweak the UI to add a spinner.

https://github.com/user-attachments/assets/61bfc882-97e4-439f-a03f-c4c4c2f00e6e

However, in debugging, I found that the request was actually timing out! And the post I was seeing was populated using info from the feed. (Now, I do have a question about why we don't show that info initially, and just fetch the comments, but that's a separate problem.)

Therefore, I made two changes.

First, if the request really times out, we should show that to the user. I implemented a similar strategy as we have on the feed, indicating the problem and giving the option to retry. We previously didn't have any logic in the post page to handle a failure.

https://github.com/user-attachments/assets/13313612-a00e-4361-865d-25eb896fbaad

Second, I increased the timeout so that, even on a slow connection, the post and comments will eventually load! Because, to be honest, if I'm on a slow connection, sometimes I'm just willing to wait for it to load rather than prematurely/artificially timing out.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -16,6 +16,7 @@ import 'package:thunder/core/enums/fab_action.dart';
 import 'package:thunder/core/models/comment_view_tree.dart';
 import 'package:thunder/core/singletons/lemmy_client.dart';
 import 'package:thunder/shared/comment_sort_picker.dart';
+import 'package:thunder/shared/error_message.dart';
 import 'package:thunder/shared/gesture_fab.dart';
 import 'package:thunder/shared/input_dialogs.dart';
 import 'package:thunder/shared/snackbar.dart';
@@ -264,6 +265,10 @@ class _PostPageState extends State<PostPage> {
             }
             setState(() {});
           }
+
+          if (state.status == PostStatus.failure) {
+            showSnackbar(state.errorMessage ?? l10n.missingErrorMessage);
+          }
         },
         buildWhen: (previous, current) {
           return !current.didScrollPositionChange;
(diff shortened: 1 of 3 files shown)


hjiangsu:
> Now, I do have a question about why we don't show that info initially, and just fetch the comments, but that's a separate problem

This was actually added when I first implemented the new post page, but it was lost along the way with one of the follow up PRs.

lib/post/pages/post_page.dart
```diff
@@ -506,6 +511,33 @@ class _PostPageState extends State<PostPage> {
...
+                        SliverFillRemaining(
+                          hasScrollBody: false,
+                          child: Center(
+                            child: StatefulBuilder(
+                              builder: (context, setState) => ErrorMessage(
+                                title: l10n.unableToLoadPost,
+                                message: l10n.internetOrInstanceIssues,
+                                actions: [
+                                  (
+                                    text: AppLocalizations.of(context)!.retry,
```
hjiangsu: Use `l10n` here!

micahmo:
> This was actually added when I first implemented the new post page, but it was lost along the way with one of the follow up PRs.

Gotcha! Any plans to bring it back?

lib/post/pages/post_page.dart
micahmo: Good catch, done!

hjiangsu:
> Gotcha! Any plans to bring it back?

I don't see why not!

hjiangsu (approved):

Issue #1798, opened 2025-03-28
Post body instance settings not working as described


aedyr:
### Bug Description

Under Appearance > Posts> Post Body Settings, the switches for "Show User Instance" and "Show Community Instance" have unexpected behaviors.

Show User Instance- No change visible in post body regardless of setting
Show Community Instance- When turned on, BOTH user instance and community instance are shown in the post body details

### Expected Behaviour

These settings should toggle instance visibility in the post body independently for User and Community.

### Steps to Reproduce

1. Go to Settings > Appearance > Posts > Post Body Settings
2. Toggle "Show Community Instance" to ON, and "Show User Instance" to OFF.
3. Open any post, observe that instance is shown for both the user/author and the community.

### App Version

0.6.1

### Device

Pixel 8a

### OS

Android 15

PR #1799, opened 2025-03-28
Fix instance display in post

micahmo:
## Pull Request Description

This PR fixes an issue with the instance display setting for post bodies. The first problem is that the user instance display was pulling the setting from the community instance display. The second problem (not really a problem, but to keep it in parity with the user one) is to allow the setting to be passed in rather than pulling directly from the state.

## Issue Being Fixed

Issue Number: #1798

## Checklist


hjiangsu (approved):
LGTM!

PR #1802, opened 2025-03-29
Fix lag when favouriting a community

hjiangsu:
## Pull Request Description

> Hide whitespace when reviewing

This PR fixes an issue where favouriting a community would take a while to occur (upwards of a few seconds). After investigation, it looks like we're refreshing the entire account information when we favourite a community, rather than just refreshing the favourited communities.

This PR also adds snackbar messages when an error occurs, and fixes an issue where tapping on the "Add/Remove favourite" from the community feed popup menu would not refresh properly.

## Issue Being Fixed

Issue Number: https://lemmy.world/post/27418233/16093814

## Checklist

### DIFF
--- a/lib/feed/utils/community.dart
+++ b/lib/feed/utils/community.dart
@@ -56,7 +56,7 @@ Future<Map<String, dynamic>> fetchCommunityInformation({int? id, String? name}) 
 Future<void> toggleFavoriteCommunity(BuildContext context, ThunderCommunity community, bool isFavorite) async {
   if (isFavorite) {
     await Favorite.deleteFavorite(communityId: community.id);
-    if (context.mounted) context.read<AccountBloc>().add(const RefreshAccountInformation());
+    if (context.mounted) context.read<AccountBloc>().add(const GetFavoritedCommunities());
     return;
   }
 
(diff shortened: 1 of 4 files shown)


lib/feed/utils/community.dart
```diff
@@ -54,22 +55,26 @@ Future<Map<String, dynamic>> fetchCommunityInformation({int? id, String? name})
...
+    Favorite favorite = Favorite(
+      id: '',
+      communityId: community.id,
+      accountId: account!.id,
+    );
+
+    await Favorite.insertFavorite(favorite);
+    if (context.mounted) context.read<AccountBloc>().add(const GetFavoritedCommunities());
+  } catch (e) {
+    showSnackbar(e.toString());
```
micahmo: Can you do `showSnackbar(getExceptionErrorMessage(e))` just in case it's an exception we know and have a nice string for? If not, it'll still show the raw exception, so we shouldn't lose any information.

micahmo (approved):
LGTM

lib/feed/utils/community.dart
hjiangsu: Done!

Issue #1776, opened 2025-03-24
Add "Show User Avatar If Set" option


Coelacanthiform:
### Feature Description

Shows user Avatar in front of username for users who have set a custom profile picture, else hides the user avatar. Example:

![Image](https://github.com/user-attachments/assets/7ba828f5-5899-43e9-88ca-2e97b751d20a)

PR #1806, opened 2025-03-29
Show user avatars only when they have an associated image

hjiangsu:
## Pull Request Description

This PR changes the logic for showing user avatars in the post page (comments). When user avatars are enabled, only user avatars that have images will be displayed within comments. Otherwise, they will be hidden.

I'm not sure whether this change should be its own separate option, or if it's good to use this as the default. If you have any opinions on this, let me know!

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1776

## Screenshots / Recordings

|Before|After|
|-|-|
|Screenshot_1743283908|Screenshot_1743283862|

## Checklist

### DIFF
--- a/lib/shared/chips/user_chip.dart
+++ b/lib/shared/chips/user_chip.dart
@@ -79,7 +79,7 @@ class UserChip extends StatelessWidget {
               child: Row(
                 mainAxisSize: MainAxisSize.min,
                 children: [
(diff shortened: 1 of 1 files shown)


micahmo (approved):
LGTM and I'm totally good with having it on by default (but we'll see if anyone objects).

Issue #1818, opened 2025-04-04
Can't see who message is from


Fmstrat:
### Is Your Feature Request Related to a Problem?

I get a message from `tom`. I can't tell if it's `tom@here.com` or `tom@there.com`.

### Feature Description

Always include instance names in messages.

PR #1823, opened 2025-04-10
Show full username in private messages

micahmo:
## Pull Request Description

This PR improves the private message widget to use our standard `UserFullNameWidget` so that we can see instance names, as well as respecting other settings related to color, weight, etc.

Notes:
- We know that our private message page could use an overhaul, but this was a nice easy change to knock out.
- I added wrapping in case the names are long. On a device with a normal DPI (i.e., not my emulator) or on an instance with a shorter name, everything should still be on one line as before.
- Please review without whitespace.

## Issue Being Fixed

Issue Number: #1818

## Screenshots / Recordings

| Before | After |
|--------|--------|
| image | image | 

## Checklist


hjiangsu (approved):
LGTM!

PR #1834, opened 2025-04-30
Fix color of block community button

micahmo:
## Pull Request Description

This is exactly the same as #1693 but for communities instead of users.

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu (approved):
LGTM!

Issue #1833, opened 2025-04-29
Comments are being shown in reply to the wrong comments.


Glazed:
### Bug Description

I was reading some comments and one reply was entirely out of line and misquoting the parent comment to boot. So I looked at the website and it turns out Thunder is threading the comments incorrectly. A refresh didn't help, either.

I'm on Lemmy.world:

https://lemmy.world/post/28880815

The original instance's post:

https://sh.itjust.works/post/36982892

Thunder screenshot:

![Image](https://github.com/user-attachments/assets/0f9006ee-4d71-4d8f-9f59-b20fdb4b0711)

Lemmy.world screenshot:

![Image](https://github.com/user-attachments/assets/38f12248-3db3-4c9e-9bc9-ca29ddb2a57b)

And I can't get those "2 more replies" to load in Thunder.

I even cleared the app's cache and tried again after adding my own comment so I could find the post again, and now it's even worse. The comment in question appears to be in reply to my comment now.

![Image](https://github.com/user-attachments/assets/b1822a5e-0b26-4fb4-882d-6ebe1671a7e5)

### Expected Behaviour

Comments should be shown in reply to their real parent.

### Steps to Reproduce

View the post comments in Thunder.

### App Version

0.7.0-1

### Device

Pixel 7 Pro

### OS

Android 15

hjiangsu:
Thanks for the report!

That's odd, I tried it myself and it seems to be correct on my end (verified with the web UI using anonymous lemmy.world account)

![Screenshot_20250429-153818.png](https://github.com/user-attachments/assets/919e6694-8617-4f5c-976e-a68341558c67)

hjiangsu:
Just to help me reproduce the issue:
- What's the current sorting applied to the comments?
- If you access the post from an anonymous account or another account, does the issue still occur?
- How are you accessing the post itself? Was it directly through the feed (e.g., tapped the post from the feed page), through searching, or some other method?

Glazed:
I figured it out. I have the root commenter blocked: sunzu2@thebrainbin.org

I realized this when I logged into the website in my browser and the entire comment thread wasn't visible at all. But Thunder is still showing me the replies, and since the root commenter is blocked, it tacks the reply onto the most recent root comment. The "build the wall" comment was the most recent root one until mine.

This is also why I can't load the replies. I'm trying to load replies that the Lemmy server refuses to show me because I blocked the author.

Try blocking that user and you might be able to reproduce the issue.

PR #1835, opened 2025-04-30
Fix issue where blocked user comments are attached to wrong comment tree

hjiangsu:
## Pull Request Description

This PR fixes an issue where comments are sometimes attached to the wrong parent comment. This primarily happens when a user is viewing a post containing comments for a blocked user.

Additional details:
Previously, we had logic to attach a comment to the root node if a parent comment was not found. This was to account for top-level comments that don't have a parent node.

```dart
if (parent == null) {
  return root.addReply(commentNode);
}
```

This logic doesn't work as expected in situations where there are blocked users. It turns out that blocking a user prevents the API from returning back their comments. However, any replies to the blocked user are returned back from the API. Due to this, the replies no longer have a parent node to attach to and is therefore attached to the root node.

To match parity with the web UI, I've added another condition to which requires the `parentId` to be `0` in order for us to attach the comment to the root node. Otherwise, the comment is disregarded if there is no parent node found.

I've verified this behaviour by blocking a user and comparing the results from Thunder to the one shown in the web UI and vice versa.

## Issue Being Fixed

Issue Number: https://github.com/thunder-app/thunder/issues/1833

## Checklist

### DIFF
--- a/lib/comment/models/comment_node.dart
+++ b/lib/comment/models/comment_node.dart
@@ -46,11 +46,11 @@ class CommentNode {
   static void insertCommentNode(CommentNode root, String parentId, CommentNode commentNode) {
     CommentNode? parent = findCommentNode(root, parentId);
 
-    if (parent == null) {
+    if (parent == null && parentId == "0") {
       return root.addReply(commentNode);
     }
 
-    parent.addReply(commentNode);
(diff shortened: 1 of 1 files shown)


micahmo (commented):
Nice!

PR #1842, opened 2025-05-04
Chore: Improve UserPreferences singleton usage

hjiangsu:
## Pull Request Description

This PR improves our usage of `UserPreferences` singleton by adding methods to get, set, and remove settings based on a given `LocalSetting`. I've also gone ahead and looked through the codebase to improve our consistency when it comes to using this singleton!

Note: Since there are some internal preferences that are not defined in `LocalSettings`, I've decided to keep those as is, but we could clean things up there by adding them as `LocalSettings` (with some way to indicate that they are internal fields to exclude them from search, etc.)

Usage:

To fetch the preferences instance (useful for internal settings not exposed by LocalSettings): `final prefs = UserPreferences.instance.preferences;`

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/account/bloc/profile_bloc.dart
+++ b/lib/account/bloc/profile_bloc.dart
@@ -147,7 +147,7 @@ class ProfileBloc extends Bloc<ProfileEvent, ProfileState> {
       if (account == null) return emit(state.copyWith(status: ProfileStatus.failure));
 
       // Set this account as the active account
-      final prefs = (await UserPreferences.instance).sharedPreferences;
+      final prefs = UserPreferences.instance.preferences;
       prefs.setString('active_profile_id', account.id);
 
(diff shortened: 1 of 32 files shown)


micahmo:
> some way to indicate that they are internal fields to exclude them from search

FYI, we do have a `searchable` flag on `LocalSettings`.

https://github.com/thunder-app/thunder/blob/5c463ecc7a50c5489926b6cb8e2c0940aa9fd351/lib/core/enums/local_settings.dart#L462-L463

hjiangsu:
> FYI, we do have a searchable flag on LocalSettings.

Thanks for reminding me about this! In any case, I'll likely make these changes in follow up PRs since this one is fairly large already 😅

PR #1850, opened 2025-05-21
Upgrade Flutter version to 3.32.0

hjiangsu:
## Pull Request Description

This PR upgrades Thunder to use the latest stable Flutter version (3.32). A few other changes were performed here to allow for the upgrade to run smoothly:
- Bumped the `compileSdkVersion` and `targetSdkVersion` to SDK 35 (Android 15)
- Re-enabled Impeller by default as there have been improvements to the engine. If additional visual glitches are observed, we can disable it again if needed
- Ran `flutter pub upgrade` to upgrade our dependencies (does not include major versions as that should be dealt with in a separate PR)
- Replaced `package:flutter_gen/gen_l10n` with `package:thunder/l10n` as synthetic packages are being deprecated
  - The `app_localization` files are now generated under `localizations` directory. See the [deprecation notice](https://docs.flutter.dev/release/breaking-changes/flutter-generate-i10n-source) for more details 
- The `l10n_esperanto` source was changed to `https://github.com/veloce/l10n_esperanto` which contains fixes for the `intl` package (required for Flutter 3.32). This change can be reverted once the main package updates it's `intl` version.
- Updated workflows to use the new Flutter version

## Issue Being Fixed

Issue Number: N/A

## Checklist

- [ ] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [ ] Did you use localized strings (and added appropriate descriptions) where applicable?
- [ ] Did you add `semanticLabel`s where applicable for accessibility?

PR #1852, opened 2025-05-27
Reduce initial post loading API calls

hjiangsu:
## Pull Request Description

This PR fixes an issue where loading the initial feed triggers more `GetPost` API calls than necessary. The problem was caused by the following check running too early/frequently (before the initial API calls had completed)

`bool isScrollable = _scrollController.position.maxScrollExtent > _scrollController.position.viewportDimension;`

A delay was added to allow the initial API calls finish before performing the above check. Additionally, this PR also fixes a small regression from https://github.com/thunder-app/thunder/pull/1850, where the images within comments were not properly constrained. 

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
Nice!

PR #1861, opened 2025-06-03
Fix the position of the refresh indicator

micahmo:
## Pull Request Description

This PR fixes a longstanding issue I've had where the position of the refresh indicator was not in the right location on my physical device. The problem was that the offset was hard-coded, rather than being calculated as the height of the status bar plus the app bar.

Note that in making this change, I refactored the app bar height, everywhere it was used, into a constant, which is why it looks like there are a lot of changes.

@hjiangsu Can you verify that it still looks right on iOS?

## Issue Being Fixed

Issue Number: N/A

## Checklist


hjiangsu (approved):
LGTM! Thanks for consolidating the app bar height logic!

PR #1865, opened 2025-06-04
Improve notification user handling

micahmo:
## Pull Request Description

This PR makes a slight tweak to how notifications across multiple accounts are handled, specifically the use case of opening a notification directed at a user which is not the currently selected user in Thunder.

The previous behavior is:
- Temporarily switch the context of the app to the new user
- Perform any actions while looking at the notification as the new user
- Restore the previous context when navigating out of the notifications page

This workflow has the potential for confusion regarding which user you are when you look at the notification, as well as when you navigate back (it changes without any indication), and it has potential for bugs as well.

I propose a new workflow, which is to *explicitly* change the selected user to the one that received the notification, and tell the user we are making this change via a snackbar. Then do *not* automatically switch back to another user. That can be done manually if desired.

This change is inspired by how some popular social media apps work, and I think it makes the whole interaction much more clear and straightforward.

## Issue Being Fixed

Issue Number: N/A

## Screenshots / Recordings

https://github.com/user-attachments/assets/1315d9e1-c0c6-4708-a063-1fdbc6f3f8ac

## Checklist

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -2571,6 +2571,10 @@
   "@suggestedTitle": {
     "description": "Suggested title for the post."
   },
+  "switchedAccount": "Switched to {username}",
+  "@switchedAccount": {
+    "description": "Message shown to the user when Thunder automatically switches accounts"
+  },
   "system": "System",
   "@system": {
     "description": "Describes using the system settings for theme"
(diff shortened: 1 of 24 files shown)


hjiangsu (approved):
LGTM! I agree with the new flow 😊

PR #1878, opened 2025-06-16
Fix community naming

micahmo:
## Pull Request Description

This PR fixes an issue where community display names were being shown in the post body, despite having display names turned off in settings.

The problem originated from one of the refactorings where a confusing property, `name`, returns either the `title` (display name) or the `name`. You would have to explicitly use `communityName` to *not* get the `title`.

The concern was raised at the time...
https://github.com/thunder-app/thunder/pull/1754#discussion_r1999475977
...and I think this bug is a perfect example of the confusion in usage.

If it's ok, I've implemented the change that I proposed at the time, where the property that returns the title or name is explicitly called `titleOrName` and the `name` is just `name`. I think it's much more clear. Plus, there is very little usage of `titleOrName`, so the fact that it has kind of a clunky name is not really big deal. Let me know if that's ok or if there is a better suggestion.

While implementing the fix, I found several other places where I believe we were using the wrong value. Due to the renaming, the places that were _wrong_ before are _right_ now, so they don't appear in the PR, but I'll list them below.

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/community/widgets/community_drawer.dart#L485

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/modlog/widgets/modlog_item_context_card.dart#L126

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/modlog/widgets/modlog_item_context_card.dart#L282

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/modlog/widgets/modlog_item_context_card.dart#L394

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/modlog/widgets/modlog_item_context_card.dart#L400

(this is the one that prompted the PR)
https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/post/widgets/post_body/post_body_title.dart#L212

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/post/widgets/post_bottom_sheet/general_post_action_bottom_sheet.dart#L105

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/post/widgets/post_bottom_sheet/post_action_bottom_sheet.dart#L94

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/shared/cross_posts.dart#L76

https://github.com/thunder-app/thunder/blob/78a5d9c22de57bba2d66a21cde93f0127169a355/lib/shared/cross_posts.dart#L122

## Issue Being Fixed

Issue Number: N/A

### Before

https://github.com/user-attachments/assets/0fe18513-b75e-480b-8781-c6823a03fc4f

### After

https://github.com/user-attachments/assets/cfb188f5-e325-4702-98ba-e2351b0df4ff

## Checklist

### DIFF
--- a/lib/comment/view/create_comment_page.dart
+++ b/lib/comment/view/create_comment_page.dart
@@ -443,7 +443,7 @@ class _CreateCommentPageState extends State<CreateCommentPage> {
                                         _bodyTextController.text = _bodyTextController.text.replaceRange(
                                           _bodyTextController.selection.end,
                                           _bodyTextController.selection.end,
-                                          '!${community.communityName}@${fetchInstanceNameFromUrl(community.url)}',
+                                          '!${community.name}@${fetchInstanceNameFromUrl(community.url)}',
                                         );
                                       });
                                     },
--- a/lib/community/bloc/community_bloc.dart
+++ b/lib/community/bloc/community_bloc.dart
@@ -58,7 +58,7 @@ class CommunityBloc extends Bloc<CommunityEvent, CommunityState> {
           emit(state.copyWith(
             status: CommunityStatus.success,
             community: community,
-            message: response.blocked ? l10n.successfullyBlockedCommunity(community.communityName) : l10n.successfullyUnblockedCommunity(community.communityName),
+            message: response.blocked ? l10n.successfullyBlockedCommunity(community.name) : l10n.successfullyUnblockedCommunity(community.name),
           ));
         } catch (e) {
           return emit(state.copyWith(status: CommunityStatus.failure));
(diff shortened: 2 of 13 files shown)


hjiangsu (approved):
LGTM and good catch!

> If it's ok, I've implemented the change that I proposed at the time, where the property that returns the title or name is explicitly called titleOrName and the name is just name.

I think that's reasonable - I did want to keep the `name` field across `ThunderUser` and `ThunderCommunity` consistent (e.g., in that they represent either the display/community name or the username/name as a fallback) but I do agree that it can lead to some confusion!

micahmo:
> I did want to keep the `name` field across `ThunderUser` and `ThunderCommunity` consistent (e.g., in that they represent either the display/community name or the username/name as a fallback)

We could always rename `ThunderUser.name` to `nameOrTitle` or something similar for consistency.

Issue #1893, opened 2025-06-25
"View all comments" context dissapears when refreshing comments from pulling down refresh function


danwardvs:
### Bug Description

Pulling down from the top to refresh comments doesn't show all the messages after viewing a comment from the inbox. Pressing the refresh button does.

### Expected Behaviour

Pulling down from the top would show all comments just like the refresh button does.

### Steps to Reproduce

Open a comment on a submission from the inbox

![Image](https://github.com/user-attachments/assets/2b8b8ace-6260-4a1b-9200-550663a68eb1)

Press "View all comments" on the submission comment view

![Image](https://github.com/user-attachments/assets/f5887214-c9ef-46f4-a06f-3b8495b3e1d1)'

This shows all the comments as expected.

![Image](https://github.com/user-attachments/assets/68e9cd46-a726-42a3-ae74-5689f71031ac)

Then when refreshing by pulling down from top, it enters the bug state where it shows only the comment from the inbox but also doesn't show a "View all comments" button. Refreshing by pressing the button at the top of the screen. Can swap between these two states by using the two buttons.

![Image](https://github.com/user-attachments/assets/ab2262f0-c1ed-4290-b9be-2d45e959dc28)

### App Version

v0.7.1

### Device

Moto G Power 5G 2024

### OS

Android 15

hjiangsu:
Just an update - this should now be fixed with v0.7.3!

PR #1894, opened 2025-06-26
Fix inconsistent comment refresh logic when navigating from inbox

hjiangsu:
## Pull Request Description

This PR fixes an issue where the refresh logic on the post page was inconsistent when navigating from the inbox. With this fix:
- All refresh actions should now account for whether the post is being viewed in a comment context (e.g., when "View all comments" is displayed).
- If "View all comments" is available, then any refresh will re-fetch the post with the specific comment.
- Once "View all comments" is tapped, then any refresh will fetch the post with all the comments as usual.

## Issue Being Fixed

Issue Number: #1893

## Screenshots / Recordings

https://github.com/user-attachments/assets/8b58c0da-c12c-4173-8c60-d5cd46a129db

## Checklist

### DIFF
--- a/lib/post/pages/post_page.dart
+++ b/lib/post/pages/post_page.dart
@@ -327,7 +327,12 @@ class _PostPageState extends State<PostPage> {
           return RefreshIndicator(
             onRefresh: () async {
               HapticFeedback.mediumImpact();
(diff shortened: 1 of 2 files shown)


micahmo (approved):
Nice fix

PR #1895, opened 2025-06-27
Fix comment metadata disappearing for comments with no children

micahmo:
## Pull Request Description

This PR fixes a small regression from #1890 where metadata from comments with no children would get hidden when tapped, even though there was no child comment count to take its place.

## Issue Being Fixed

Issue Number: N/A

#### Before

https://github.com/user-attachments/assets/95211983-8687-4fd2-9fe0-2a05057b76cc

#### After

https://github.com/user-attachments/assets/5af15718-6716-443e-a46b-93061b9c3146

## Checklist


hjiangsu (approved):
LGTM!

PR #1918, opened 2025-08-05
Fix unified push notification logic

hjiangsu:
## Pull Request Description

This PR fixes a issue with unified push notifications not being registered properly. This was due to a regression when upgrading the UnifiedPush package. Additionally, error messages are now exposed via the snackbar when testing notifications.

## Issue Being Fixed

Issue Number: N/A

## Checklist


englut:
Thanks for addressing this so quickly @hjiangsu :pray:

PR #1977, opened 2025-11-12
Fix settings search elevation

micahmo:
## Pull Request Description

Hey, I'm popping back in! 😊 I noticed that the search widget on the settings page has an elevation now. I'm not sure if the change was intentional, but it no longer matches the search page. It looks like the change was lost when the structure was reorganized (https://github.com/thunder-app/thunder/pull/1924/files#diff-70c4cf2c8269388c5f6753ba0c9b01e972187c04800f701e7cd9dd0494e67b31L69).

## Issue Being Fixed

Issue Number: N/A

### Before

| image |
| - |

### After

| image |
| - |

## Checklist


ggichure (approved):

hjiangsu:
Thanks for the fix!

PR #1989, opened 2025-12-03
Upgrade to Flutter 3.38

hjiangsu:
## Pull Request Description

This PR upgrades Thunder to work with Flutter 3.38. I've also updated the relevant GitHub workflows to use the new version!

Additionally:
- I've enabled the usage of Swift Package Manager to fix an issue related to background_fetch. For background_fetch, I've also removed the setup steps from Android as that is no longer required.
- While attempting to upgrade package dependencies, I ran into some issues where some packages were locked at older versions. This was primarily due to `xayn_readability` using older dependencies. Since I don't quite have the time to maintain that package, and that package is not being actively maintained, I've decided to remove the reader mode for Android in the meantime.

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo:
Hey, I recently discovered something called [Flutter Version Management](https://fvm.app/) (FVM). It really helps with working on multiple Flutter projects. Instead of having a global `flutter` with a fixed version for all projects, you use `fvm` to prefix all commands, and then it ensures that the right version is used for each project. So for this latest upgrade that you did, I just had to run `fvm use 3.38.3`, and then `fvm flutter run` or any other command like `fvm flutter pub get` and everything works great.

If you're interested in supporting FVM, it just adds these changes to the repo:
 - A `.fvmrc` file which just specifies the Flutter version
 - Ignoring the FVM cache (`.fvm/`) in the `.gitignore`
 - And since we use VS Code, it can also modify `.vscode/settings.json` so VS Code knows how to invoke the right version of Flutter

And it doesn't prevent anyone from continuing to just use `flutter` commands directly. But the configuration is available for anyone who happens to use FVM.

Here's another Flutter app that uses it: 
https://github.com/pizi-0/flutter-scrcpygui/blob/main/.fvmrc
https://github.com/pizi-0/flutter-scrcpygui?tab=readme-ov-file#build-steps

Just a thought!!

PR #1990, opened 2025-12-08
Improve background notification checks

hjiangsu:
## Pull Request Description

This PR improves the logic for handling background local notification checks. Each time a background check runs, it now triggers a silent notification that is automatically dismissed once the check completes. This allows us to cross-verify that background checks are being triggered properly.

Additionally, this PR adds support for user mentions and private message notifications. I realized we only had checks in place for comment replies, so I added the missing logic for these notification types as well!

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/lib/src/features/notification/presentation/utils/android_notification.dart
+++ b/lib/src/features/notification/presentation/utils/android_notification.dart
@@ -1,21 +1,23 @@
-// Package imports
 import 'dart:convert';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 
-// Project imports
 import 'package:thunder/src/features/account/account.dart';
 import 'package:thunder/src/core/enums/full_name.dart';
 import 'package:thunder/src/core/enums/local_settings.dart';
 import 'package:thunder/src/core/singletons/preferences.dart';
 import 'package:thunder/src/features/notification/notification.dart';
 
-const String _repliesChannelId = 'inbox_replies';
-const String _repliesChannelName = 'Inbox Replies';
-const String _mentionsChannelId = 'inbox_mentions';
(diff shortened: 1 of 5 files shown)


lib/src/features/notification/presentation/utils/android_notification.dart
```diff
@@ -1,21 +1,23 @@
...
 import 'package:thunder/src/features/notification/notification.dart';
 
-const String _repliesChannelId = 'inbox_replies';
-const String _repliesChannelName = 'Inbox Replies';
-const String _mentionsChannelId = 'inbox_mentions';
-const String _mentionsChannelName = 'Inbox Mentions';
-const String _messagesChannelId = 'inbox_messages';
-const String _messagesChannelName = 'Inbox Messages';
+const String _repliesChannelId = 'replies';
+const String _repliesChannelName = 'Replies';
```
micahmo, line 13: The only problem with changing these channel names now is that it will create duplicate channels for existing users. On Android, at least, these is what the user sees when they go into system notification settings (and they can turn of/off each channel individually).

micahmo (commented):
> I realized we only had checks in place for comment replies, so I added the missing logic for these notification types as well!

Just checking, does tapping these notifications navigate to the right place in the app? I think that's why we hadn't added the other types before.

lib/src/features/notification/presentation/utils/android_notification.dart
hjiangsu: Ahh okay, I'll revert the changes here - thanks for the heads up!

hjiangsu:
> Just checking, does tapping these notifications navigate to the right place in the app?

Let me check! I'll likely do a follow up PR to add these if they're missing

PR #1994, opened 2025-12-17
Additional local notification improvements

hjiangsu:
## Pull Request Description

This PR improves the logic for local notifications and applies the following changes:
- Fixed an issue where tapping on a notification would not work properly.
  - Previously, tapping on a notification would briefly show the notification page, and then immediately dismiss it. Additionally, tapping on a notification meant for a different profile would not switch over to the profile properly.
  - This regression was likely introduced when I refactored the profile switching logic. Tapping on a notification should now behave properly.
- Added notification pages for comment mentions and private messages
- Reverts the notification channel ids back to their previous values

## Issue Being Fixed

Issue Number: N/A

## Checklist


micahmo (approved):
Awesome!

PR #2012, opened 2026-01-01
Fix 16KB page size requirement

hjiangsu:
## Pull Request Description

This PR fixes an issue where some libraries were not 16KB page size aligned. This issue comes from a transitive dependency on `unified_push` with version 6.2.0. `webcrypto.dart` package. For now, I've pinned the unified_push version to 6.1.0 which doesn't have the same issue.

## Issue Being Fixed

Issue Number: N/A

## Checklist

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1463,46 +1463,6 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "0.28.0"
-  screen_retriever:
-    dependency: transitive
-    description:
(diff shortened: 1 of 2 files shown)


TotallyAvailable:
@hjiangsu ***Please*** remember to set the GITHUB_TOKEN permissions for **all** workflows to the ***absolute minimum*** necessary (including 'Secret Source: None' unless required). 

Workflows default to using **write-all** tokens unless actively changed.

PR #2061, opened 2026-03-03
Remove archive.today alternative source

hjiangsu:
## Pull Request Description

This PR removes archive.today as an alternative source after the recent controversy surrounding it.

## Issue Being Fixed

Issue Number: https://lemmy.world/post/43495098

## Checklist

### DIFF
--- a/lib/src/app/shell/navigation/link_navigation_utils.dart
+++ b/lib/src/app/shell/navigation/link_navigation_utils.dart
@@ -341,7 +341,6 @@ List<({String sourceName, String link})> generateAlternateSources(String link) {
 }
 
 List<({String sourceName, MessageFormat template})> _alternateSources = [
-  (sourceName: 'Archive Today', template: MessageFormat('https://archive.today/{link}')),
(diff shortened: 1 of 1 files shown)


micahmo:
Man, for some reason archive.today always felt like a ticking time bomb, but I couldn't put a finger on it. Maybe it was too good to be true! 😆

Per the Ars Technica article on this topic...

> The guidance says editors can remove Archive.today links when the original source is still online and has identical content; replace the archive link so it points to a different archive site, like the Internet Archive, Ghostarchive, or Megalodon; or “change the original source to something that doesn’t need an archive (e.g., a source that was printed on paper), or for which a link to an archive is only a matter of [convenience](https://en.wikipedia.org/wiki/Wikipedia:Convenience_link).”

So maybe we could add Ghostarchive and Megalodon?

It looks like the format for Ghostarchive is `https://ghostarchive.org/search?term=<encoded-url>`

Megalodon appears to be Japanese, so I'm not sure exactly how to navigate the site, but it could still be useful.

Issue #2046, opened 2026-01-27
Add support for lemmyverse links


RandomDennisLee:
### Is Your Feature Request Related to a Problem?

Lemmyverse.link is an utility that allows users to provide instance-agnostic links to posts / comments. The URL format it uses is pretty straightforward, lemmyverse.link/instance/post. It would be nice if thunder could read those links and open the correct post instead of loading them in a browser. 

### Feature Description

.

PR #2117, opened 2026-07-15
Handler for lemmyverse.link shared links.

RandomDennisLee:
## Pull Request Description

Added check for links in the format of "lemmyverse.link/known_instance/post/1234". Compared against known_instance.dart to look for lemmy / piefed shared links. Domain is generic, will match any domain. 

On match, will rewrite the link from "lemmyverse.link/known_instance/post/1234" to "known_instance/post/1234" and pass it on to the existing link parsers.

## Issue Being Fixed

Issue Number: [Add support for lemmyverse links](https://github.com/thunder-app/thunder/issues/2046)

## Checklist

- [N/A] If a new package was added, did you ensure it uses an appropriate license and is actively maintained?
- [N/A] Did you use localized strings (and added appropriate descriptions) where applicable?
- [N/A] Did you add `semanticLabel`s where applicable for accessibility?

### DIFF
--- a/lib/src/foundation/utils/threadiverse_link_parser_utils.dart
+++ b/lib/src/foundation/utils/threadiverse_link_parser_utils.dart
@@ -32,10 +32,8 @@ final RegExp _lemmyShortUserUrl = RegExp(r'^@?(https?:\/\/)?(.*)/u/([^@\n]*)$');
 /// Groups: 2=username, 3=instance
 final RegExp _lemmyUserMention = RegExp(r'^@?(https?:\/\/)?((?:(?!\/u\/u).)*)@(.*)$');
 
-/// Matches instance.tld/post/123
+/// Matches instance.tld/post/123?foo=bar
 /// Groups: 2=instance, 3=postId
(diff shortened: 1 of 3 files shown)


lib/src/foundation/utils/threadiverse_link_parser_utils.dart
```diff
@@ -34,7 +34,9 @@ final RegExp _lemmyUserMention = RegExp(r'^@?(https?:\/\/)?((?:(?!\/u\/u).)*)@(.
 
 /// Matches instance.tld/post/123
 /// Groups: 2=instance, 3=postId
-final RegExp _lemmyPostUrl = RegExp(r'^(https?:\/\/)(.*)/post/([0-9]+)$');
+// final RegExp _lemmyPostUrl = RegExp(r'^(https?:\/\/)(.*)/post/([0-9]+)$');
```
hjiangsu: No need to keep this, we can just remove it if we're replacing the regex!

hjiangsu (changes requested):
Looks good - just one small change!

